PR #1469, opened 2023-03-29
Spike: E2E browser tests using WDIO

jimCresswell:
## Changes
- Removes Playwright related deps, code, config, and docs.
- Add [WDIO](https://webdriver.io/docs/gettingstarted) related deps, code, config, and docs.
- Add an example passing and failing test, the failing test is skipped by default.
- Implement code to close the Confirmic overlay
- Wire it up to Browserstack so it runs against PR deploys.
- Add an initial user journey

## Usage
To run the browser test locally, start the dev server `npm run dev` then in a separate process (terminal) run `npm run wdio`, Chrome should start and the example test should run.


e2e_tests/browser/pageobjects/teacher_key-stages_subjects.page.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+class TeacherKeyStagesSubjectsPage extends Page {
+  /**
+   * define selectors using getter methods
+   */
+
+  public get subjectMaths() {
+    // This is a pretty fragile selector, but there is very little
+    // semantic page structure to use.
+    return $("=Maths");
```
JBR90, line 14: Could we select by data-test-id ?
$('[data-test-id="your-data-test-id"]');

JBR90 (approved):
Very cool! 
Had a allign my chrome driver version with installed version of chrome to get the tests to run, might be worth adding that to docs.
npm install chromedriver@114 --save-dev

jimCresswell:
> Very cool! Had a allign my chrome driver version with installed version of chrome to get the tests to run, might be worth adding that to docs. npm install chromedriver@114 --save-dev

Good idea. Generally up to date Chrome and Chromedriver should work together. This is something that will need to be kept up to date.

PR #1601, opened 2023-05-11
chore(code quality): don't parse env args to shell commands

jimCresswell:
Addressing the warnings from the GH code scanning

## Don't shell expand args from envs
https://github.com/oaknational/Oak-Web-Application/security/code-scanning/6

To test, push and pull config. Pull seems to work fine.

## Keep regexes efficient
https://github.com/oaknational/Oak-Web-Application/security/code-scanning/1

To test, the Zod slug schema in the API route for blog previews should work the same


src/pages/api/preview/[[...path]].ts
```diff
@@ -5,7 +5,9 @@ import serverConfig from "../../../config/server";
 import errorReporter from "../../../common-lib/error-reporter";
 import OakError from "../../../errors/OakError";
 
-export const slugStringSchema = z.string().regex(/^\w+(?:[-_]\w+)*$/);
+export const slugStringSchema = z
+  .string()
+  .regex(/^[a-zA-Z0-9]+(?:[-_][a-zA-Z0-9]+)*$/);
```
Comment (jimCresswell) [inline src/pages/api/preview/[[...path]].ts:10]: `\w` is equivalent to `[a-zA-Z0-9_]`, so the difference here is that the character group no longer includes `_`, which I think is fine for slugs. Technically the old version would have passed for a slug starting with `_`, but I am not sure that would have been correct

mantagen (approved):
Looks good to me. Might need to re-run CI as it looks like a flaky pa11y test is failing

PR #1602, opened 2023-05-11
docs(security): communicate GitHub private vulnerability reporting option

jimCresswell:
Update the security policy to indicate the reporting via GitHub is an option (once we turn it on)

### DIFF
--- a/public/.well-known/security.txt
+++ b/public/.well-known/security.txt
@@ -11,6 +11,8 @@ Preferred-Languages: en
 Encryption: https://www.thenational.academy/.well-known/pgp-key.txt
 Encryption: https://keys.openpgp.org/vks/v1/by-fingerprint/E925058764BF1E055A95CFA7971F062051B60B7D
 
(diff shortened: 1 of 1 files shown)


johnrobeds (approved):

public/.well-known/security.txt
```diff
@@ -11,6 +11,8 @@ Preferred-Languages: en
 Encryption: https://www.thenational.academy/.well-known/pgp-key.txt
 Encryption: https://keys.openpgp.org/vks/v1/by-fingerprint/E925058764BF1E055A95CFA7971F062051B60B7D
 
+Alternatively please follow the GitHub private vulnerability reporting steps https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability .
```
johnrobeds, line 14: I think this should be inserted as line 8:

`Contact: https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability`

And then the new signature generated across the new file

PR #1669, opened 2023-05-31
feat(infra): add lesson video bucket tf config to infra

shobbsd:
## Description

- PR to add terraform infra for the storage of lesson videos
- This infra creates 2 buckets with upload access granted Oak retool instances


andykwoods:
who are we expecting to review this?

infrastructure/web/variables.tf
```diff
@@ -0,0 +1,20 @@
+variable "project" {
+  type    = string
+  default = "oak-national-academy"
+}
+
+variable "region" {
+  default = "europe-west2"
```
tweakster, line 7: Could we add type string here please. Just for consistency.

infrastructure/web/lesson_video_buckets.tf
```diff
@@ -0,0 +1,23 @@
...
+  location      = var.region
+  force_destroy = var.environment != "production"
+
+  versioning {
+    enabled = false
+  }
+
+  cors {
+    origin          = [var.retool_origin]
+    method          = ["GET", "HEAD", "PUT", "POST", "DELETE"]
```
tweakster, line 13: Is there a reason for specifying the lst 3 methods?

If this bucket is only intended to be read over http then I think we should only specify `["GET", "HEAD"]`

infrastructure/web/lesson_video_buckets.tf
```diff
@@ -0,0 +1,23 @@
...
+    response_header = ["Content-Type", "x-goog-acl"]
```
tweakster, line 14: As I read it the response header is only relevant for PUT/POST requests so if this is read only I don't think we need this.

infrastructure/web/lesson_video_buckets.tf
```diff
@@ -0,0 +1,23 @@
...
+    max_age_seconds = 3600
+  }
+}
+
+resource "google_storage_bucket_iam_member" "lesson_videos_public" {
+  bucket = google_storage_bucket.lesson_videos.name
+  role   = "roles/storage.objectViewer"
+  member = "allUsers"
```
tweakster, line 22: Will there be a `thenational.academy` domain associated with this?

infrastructure/web/main.tf
```diff
@@ -0,0 +1,14 @@
+terraform {
+  required_providers {
+    google = {
+      source  = "hashicorp/google"
+      version = "4.50.0"
+    }
+  }
+}
+
+# Uses GOOGLE_CREDENTIALS environment variable
```
tweakster, line 10: I think it's my fault this comment is here. I know I put it on another file. Although I was wrong. This is not true so can we remove please?

infrastructure/web/variables.tf
```diff
@@ -0,0 +1,20 @@
...
+
+variable "retool_origin" {
+  type    = string
+  default = "https://creator.thenational.academy"
+}
+
+
+variable "environment" {
+  type    = string
+  default = "staging"
```
tweakster, line 19: I would keep all the defaults in step. It seems that the rest of the defaults are production defaults.

infrastructure/web/main.tf
```diff
@@ -0,0 +1,14 @@
+terraform {
+  required_providers {
+    google = {
+      source  = "hashicorp/google"
+      version = "4.50.0"
+    }
+  }
```
tweakster, line 7: When we start using this properly we'd want to store the state file in the Terraform Cloud, rather than locally.

Inside the terraform block we want a [cloud block](https://developer.hashicorp.com/terraform/cli/cloud/settings):

```terraform
  cloud {
    organization = "oak-national-academy"
    workspaces {
      name = "app-name-${var.environment}"
    }
```
Obvs changing `app-name` to something relevant to the app this is to be part of.

infrastructure/README
```diff
@@ -0,0 +1,34 @@
...
+This folder current manages the buckets responsible for holding uploaded lesson videos.
+
+The terraform project is broken out into multiple workspaces
+
+- production
+- staging
+- demo
+
+Each workspace should have an associated `<env>.tfvars` file. The `example_vars` can serve as a template.
+i.e. `staging.tfvars`
```
tweakster, line 12: tfvars files are excluded from the repo so you wouldn't use a tfvars file. When implemented in TF Cloud the vars are strored there.

infrastructure/README
```diff
@@ -0,0 +1,34 @@
...
+
+- [terraform](https://developer.hashicorp.com/terraform/downloads?product_intent=terraform)
+- [gcloud](https://cloud.google.com/sdk/docs/install)
+
+1. `terraform login` (you will need a terraform account to perform this step)
+2. follow [these instructions](https://github.com/oaknational/Cloud-Config/tree/feat/ISSUE-82-deployment-bucket/infrastructure/iam#google-cloud) on authenticating your terraform cli
+3. `terraform workspace list` - to view available workspaces
+4. `terraform workspace select <environment>` - to select your the correct environment
+5. `terraform plan -var-file=<environment>.tfvars` - to see the effects your changes will make
+6. `terraform apply -var-file=<environment>.tfvars` - to apply those changes
```
tweakster, line 34: I think it's really cool that you wrote all this. Which is why I feel bad as this is not how we'd do it once it's running in Terraform Cloud. Sorry.
I don't think TF Cloud supports the workspace command anymore and we won't be using tfvars files.

I am planning on doing a general doc on Terraform management (once I've solved all the problems around a standard process) for now I would just reference the doc in Data-Tools, I think it's good enough
https://github.com/oaknational/Data-Tools/tree/development/infrastructure#terraform-cloud

infrastructure/web/example_vars
```diff
@@ -0,0 +1,4 @@
+project       = "oak-national-academy"
+region        = "europe-west2"
+retool_origin = "https://creator-staging.thenational.academy"
+environment   = "staging"
```
tweakster, line 4: I don't think you need examples as the variables.tf file has defaults.

infrastructure/web/main.tf
tweakster: Note. I don't think you have access to Terraform Cloud atm, but I am trying to negotiate getting more people access to Terraform Cloud so hopefully that will change.

tweakster:
@shobbsd Sorry it's so many comments. You are a bit of a guinea pig on this process. I do hope to create and document something which makes this a smoother process in the future.

infrastructure/web/lesson_video_buckets.tf
shobbsd: Delete and one more are the two that arent necessary. 
Retool will be uploading the videos and displaying the videos but I need to investigate further to find out if it is POST or PUT it uses.

infrastructure/web/lesson_video_buckets.tf
shobbsd: [see comment](https://github.com/oaknational/Oak-Web-Application/pull/1669#discussion_r1230688769)

infrastructure/web/lesson_video_buckets.tf
shobbsd: I think for now it would be worth either removing this all together or adding a service account. (testing in staging first)

.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+infrastructure/ @oaknational/platform-group
```
shobbsd, line 1: the @oaknational/platform-group will need to have write access to this repo for this to work

johnrobeds:
Does this need further work @shobbsd @tweakster ?

.github/CODEOWNERS
johnrobeds: This has been granted

PR #1688, opened 2023-06-06
chore: remove error reporting when pageview fails

mantagen:
## Description

- do not report error when getPageViewProps doesn't find the internal page, as it is happening quite a bit with people landing at 404s (eg https://app.bugsnag.com/oak-national-academy/oak-web-application/errors/645bbd94de56b500096bb45e)


JBR90 (approved):
LGTM

PR #1715, opened 2023-06-13
[ ENG-49 ] spike curriculum api 2023 depending on url param

### LINKED ISSUE #1713: Allow OWA to consume data from both 2020 and 2023 hasura endpoints

mantagen:
## Description

- makes `[viewType]` param in beta path, which can take value:
  - "teachers"
  - "teachers-2023"
- write function `replaceViewType2023`, which:
  - if the user has manually navigated to the `/beta/teachers-2023` version of the site, will keep them in that version by replacing the `/teachers/` part of the link href with `/teachers-2023`
  - if the user is in `/beta/teachers`, will leave link hrefs unchanged, so the user stages in the beta teachers experience.
- adds `/curriculum-api-2023` with a `teachersHomePage.gql` query
- in `/beta/[viewType]/index.ts` (aka teachers homepage):
  - fetch data from curriculum-api or curriculum-api-2023 depending on context.params.viewType (see getStaticProps)

## Issue(s)

Fixes #1713 

Notion ticket: https://www.notion.so/oaknationalacademy/7a067876b8d94be0b4100d6c831d5a1a

## How to test

1. Go to any page in the beta.
2. Change "teachers" in url to "teachers-2023"
3. Site should still work, and clicking around should keep you in "teachers-2023"
4. If you go to teachers homepage (in teachers-2023), you should see different key stages (these are key stages fetched from the new curric api


src/common-lib/urls/betaStage.ts
```diff
@@ -0,0 +1,2 @@
+export const BETA_STAGE = ["beta", "gamma"] as const;
```
jimCresswell, line 1: Is the beta/gamma distinction still used in addition to teachers/teachers-2023?

src/__tests__/pages/beta/key-stages/[keyStageSlug]/subjects/index.test.tsx
```diff
@@ -4,7 +4,7 @@ import curriculumApi from "../../../../../../node-lib/curriculum-api/__mocks__";
 import SubjectListingPage, {
   getStaticPaths,
   getStaticProps,
-} from "../../../../../../pages/beta/teachers/key-stages/[keyStageSlug]/subjects";
+} from "../../../../../../pages/beta/[viewType]/key-stages/[keyStageSlug]/subjects";
```
Comment (jimCresswell) [inline src/__tests__/pages/beta/key-stages/[keyStageSlug]/subjects/index.test.tsx:7]: is this `viewType` rather than `teacherViewType` because you anticipate a pupil view using the same URL structure as the teacher view?

jimCresswell (approved):
Fantastic job! Two questions on naming

src/common-lib/urls/betaStage.ts
mantagen: whoops, left that in unintentionally. No more gamma

src/__tests__/pages/beta/key-stages/[keyStageSlug]/subjects/index.test.tsx
Comment (mantagen) [reply src/__tests__/pages/beta/key-stages/[keyStageSlug]/subjects/index.test.tsx:7]: More or less, yes. The parameter at that place in the path could either be "teachers" or "pupils". After September it may no longer be a nextjs url paramter ... but would still be a a parameter in the sense that the url structure is assumed to be the same for routes nested within. Whether on teachers/pupils view. E.g. unit listing ... are you viewing it as a teacher or a pupil.

jack-skerrett-oak (approved):

PR #1720, opened 2023-06-21
[ENG-144] feat: improve sharability of pages in preview mode

mcky:
## Description
Allow the user to copy the current URL in preview mode (aka with `/api/preview` prepended).

School support keep running into the issue where they share a link that looks like a preview one (has `?secret=xxx` appended) but the actual preview link should start with `/api/preview`). Hopefully this makes their lives a bit easier.

As per this comment in the code
```
   /**
     * To trigger preview mode the user has to navigate to `/api/preview/$thePath?secret=$previewSecret`
     * This link construction relies on the user currently being in preview mode and having the ?secret=
     * query param in the URL.
     *
     * The preview secret should only be accessible server side so we can't read it from the env in
     * the context of this component. In future if this turns out to be a problem, we can pass it securely
     * via cookie with `res.setPreviewData({ previewSecret  })` within the API endpoint, however passing
     * it to this component would require changes to every single CMS-controlled page's getStaticProps.
     */
```
This basically means if the user is clicking around between pages and the `?secret=` param is lost from the URL, the button will disappear.
This was a tradeoff between the confusion caused by school support copying non-preview URLs currently, and a decent chunk more dev work. To make it work on every page while in preview mode we would have to pass it from every page to the component.

n.b. I hate the white background but spent a while fighting `ButtonAsLink` trying to get the minimal variant to have a different foreground color, as the regular variant has a bit too much padding for the location IMO. If I'm being dense and missing something obvious I'd hapilly change it to a color that sticks out a bit more

## Issue(s)

Fixes ENG-144

## How to test

1. Go to {cloud link}
2. Click on _______
3. You should see _______

## Screenshots
CleanShot 2023-06-21 at 15 31 33@2x

## Checklist

- [ ] Design sign-off

### DIFF
--- a/src/components/Button/CopyLinkButton/CopyLinkButton.test.tsx
+++ b/src/components/Button/CopyLinkButton/CopyLinkButton.test.tsx
@@ -39,4 +39,36 @@ describe("Copy link button", () => {
     const clickedButton = getByLabelText("Copied to clipboard");
     expect(clickedButton).toBeInTheDocument();
   });
+
+  it("copies the current URL to the clipboard by default", async () => {
+    const { getByLabelText } = renderWithTheme(
+      <ToastProvider>
+        <CopyLinkButton />
+      </ToastProvider>
+    );
+
+    const user = userEvent.setup();
+
+    const button = getByLabelText("Copy to clipboard");
+    await user.click(button);
+
+    const clipboardText = await navigator.clipboard.readText();
+    expect(clipboardText).toBe("http://localhost/");
+  });
+
+  it("copies the provided URL to the clipboard", async () => {
+    const { getByLabelText } = renderWithTheme(
+      <ToastProvider>
+        <CopyLinkButton href="https://example.com" />
+      </ToastProvider>
+    );
+
+    const user = userEvent.setup();
+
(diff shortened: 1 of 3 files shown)


JBR90 (approved):
Looks good to me, re-minimal variant I don't think it's currently possible but @mantagen may be able to help.

jack-skerrett-oak (commented):
I'm a bit lost. In the PR description you say that if you navigate around and lose the secrets param then the copy button disappears, but see the below video.


https://github.com/oaknational/Oak-Web-Application/assets/102532470/820c3d0e-c362-423d-bb82-175ce19405da

mcky:
> I'm a bit lost. In the PR description you say that if you navigate around and lose the secrets param then the copy button disappears, but see the below video.
> Screen.Recording.2023-06-22.at.10.21.08.mov

The commit adding that was never pushed 🤦🤦🤦. I'll be more careful to wait for the deploy and do my own QA pass on there and not just locally 😅

jack-skerrett-oak (approved):
Removing the secret from the URL removes the share button from the preview component in the bottom corner.

PR #1724, opened 2023-06-22
feat: automatically add the deployment URL to PR descriptions

mcky:
## Description
Add a post deployment action that will replace the `{owa_deployment_url }` in PR descriptions with the actual URL of the deployment

To get the action working I had to fork it to get it to work outside of the `pull_request` event context (as ours is on `deployment_status`), although I plan to PR the changes back upstream, or if we want we could bring a minimal version of the same into the OWA codebase

I've also changed the template from `{cloud_link}` to `{owa_deployment_url }`

See the PR edit history below (a victim of it's own success, it replaced the one describing what it's supposed to do)
CleanShot 2023-06-23 at 11 48 18

## How to test
1. Branch off of this branch and create a PR including`{owa_deployment_url }` in the body (without the trailing space)
2. See that it's replaced with the deployment URL


jimCresswell (approved):

mcky:
@jimCresswell my only hesitation clicking merge on this is that if the step fails it would potentially derail the whole lot of CI jobs. If jobs are run in parallel and don't depend on eachother then it's fine, as long as it won't block merging

jimCresswell:
> @jimCresswell my only hesitation clicking merge on this is that if the step fails it would potentially derail the whole lot of CI jobs. If jobs are run in parallel and don't depend on eachother then it's fine, as long as it won't block merging

@mcky I'm pretty sure they run serially by default (we could look at changing that to improve speed), but no job depends on a previous job running unless that [relation is expressed in the config file](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds).

mantagen (approved):

mcky:
> @mcky I'm pretty sure they run serially by default (we could look at changing that to improve speed), but no job depends on a previous job running unless that [relation is expressed in the config file](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds).

@jimCresswell alright I'll merge, of course if it stops working while I'm away feel free to nuke it from orbit

PR #1735, opened 2023-07-03
fix: remove log statements from components

lioneljingles:
## Description

- Removed log statements from components that are loaded into the web app. 
- Did not change config, test, or storybook log statements.


jimCresswell (approved):
Thank you @lioneljingles

PR #1749, opened 2023-07-18
chore(dependabot): always increase version requirement to match new version

mantagen:
## Description

- update dependabot config so that we don't get package-lock version updates without updating the package.json version

## How to test

1. From now on, we should only get PRs from dependabot which update both package.json and package-lock.json


jimCresswell (approved):
🦾

PR #1769, opened 2023-07-31
[ENG-308] Feat/tighten server config

jimCresswell:
We discovered that setting a branch value for an env in Netlify will set _an_ env value for all environments (prod, dev, etc), if you don't specify a value it will set that value to empty string `""`.

This change to the config value management is so that if an env has `required: true` then setting the value to an empty string will cause it to fall back to the default value.

### DIFF
--- a/src/node-lib/getServerConfig.test.ts
+++ b/src/node-lib/getServerConfig.test.ts
@@ -19,6 +19,30 @@ describe("getServerConfig()", () => {
 
     expect(getServerConfig("disableIsr")).toEqual(false);
   });
+  it("Should return the default value if a required env is set to undefined", async () => {
+    process.env.SANITY_DATASET_TAG = undefined;
+    const { default: getServerConfig } = await import(
+      "../node-lib/getServerConfig"
+    );
+
+    expect(getServerConfig("sanityDatasetTag")).toEqual("default");
+  });
+  it("Should return the default value if a required env is set to an empty string", async () => {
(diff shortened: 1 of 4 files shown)


src/node-lib/getServerConfig.test.ts
```diff
@@ -19,6 +19,30 @@ describe("getServerConfig()", () => {
...
+    expect(getServerConfig("sanityDatasetTag")).toEqual("default");
+  });
+  it("Should return the default value if a required env is set to an empty string", async () => {
+    process.env.SANITY_DATASET_TAG = "";
+    const { default: getServerConfig } = await import(
+      "../node-lib/getServerConfig"
+    );
+
+    expect(getServerConfig("sanityDatasetTag")).toEqual("default");
+  });
```
mcky, line 37: nit-pick not for this specific PR: feels a tad funky having the real keys and values in here instead of some explicit testing values

_but_ it doesn't really matter and these lines are just an extension of how it already is

mcky (approved):

src/node-lib/getServerConfig.test.ts
jimCresswell: when we break this out into a reusable module we should take care of that

jimCresswell: marking resolved so I can merge, but it's a bit annoying, I'd rather this comment remained visible

PR #1770, opened 2023-07-31
feat: add logging and bugsnag error reporting to sanity-graphql client

mcky:
## Description
Add bugsnag / console logging to failed sanity-graphql requests

❓ Will catching and re-throwing result in confusion from 2 bugsnag bugs? Instead of reporting should I be creating an OakError in the hopes it will propogate with metadata to the page-level bug?

❓ Should the same be added to the curriculum API?

## Issue(s)

Arose via ENG-308 (Make OWA  config resilient to Netlify setting empty strings on unspecified envs)


jimCresswell (approved):
Thanks @mcky ! This looks good to me, but I'd like @mantagen to take a look as well

src/node-lib/sanity-graphql/index.ts
```diff
@@ -1,14 +1,17 @@
 import { GraphQLClient } from "graphql-request";
 
+import errorReporter from "../../common-lib/error-reporter";
 import getServerConfig from "../getServerConfig";
 
 import {
+  getSdk,
+  SdkFunctionWrapper,
```
mantagen, line 8: oh handy

mantagen (approved):
> ❓ Will catching and re-throwing result in confusion from 2 bugsnag bugs? Instead of reporting should I be creating an OakError in the hopes it will propogate with metadata to the page-level bug?

Still haven't formulated a consistent approach here, so what you've done is great

> ❓ Should the same be added to the curriculum API?

Good point, would make sense to add in one place per geenrated sdk

johnrobeds (approved):

PR #1772, opened 2023-08-01
Fix/lesq 186/path aliases b

kimon-satan:
## Description
I want to avoid long relative paths in OWA by using aliases such as ```@components```

## How to test

```npm run dev``` access a lesson page 
```npm run storybook``` acccess LessonList
```npm run test``` 
```npm run lint```

- [ ]  Necessary TS config updates
- [ ]  Relative paths replaced in sample files
- [ ]  Ensure that tests also run

## Checklist

- [ ] Design sign-off

### CODE REVIEW

File: tsconfig.json [hunk 1]
```diff
@@ -17,7 +17,11 @@
     "incremental": true,
     "noUncheckedIndexedAccess": true,
     "noImplicitAny": true,
-    "downlevelIteration": true
+    "downlevelIteration": true,
+    "baseUrl": ".",
```
mantagen, line 21: does it work without `baseUrl`?

mantagen (approved):
Nice one, much better!
Just wonder if it works without `baseUrl` in config?

tsconfig.json
kimon-satan: Even if it does, it's not supposed to [see here](https://nextjs.org/docs/app/building-your-application/configuring/absolute-imports-and-module-aliases) where it says "Each of the "paths" are relative to the baseUrl location"

PR #1784, opened 2023-08-09
Fix/lesq 203/default title logic

kimon-satan:
**Job Story:**

When *using buttonAsLink*, I want the title to render without risk of “undefined” being added as a suffix

**Detail:**

Untitled

I think the logic is *supposed* to be as follows (but sounds like the logic might not be reflected properly in the code).

- `defaultTitle` is the string passed to html title attr if `title` prop not passed
- it takes value `ariaLabel` if that prop is passed (if there's a need to pass a different `ariaLabel` than the visible label)
- otherwise it should take the visible `label` prop, and add a suffix of `labelSuffixA11y` if that is passed

Essentially, the advice of a11y experts is that ideally the screen reader label should be same as visible -- but in cases where there are visual cues which would add context to a sighted user, then it's best to add a suffix to add context for assistive tech. The label read allowed should *start with* the visible label. Apologies I can't find a reference for this right now.Looking at the code i think it's missing some brackets:

```
const defaultTitle =
    ariaLabel || (labelSuffixA11y ? `${label} ${labelSuffixA11y}` : "");
```

## Acceptance Criteria:

- [x ]  Fix logic so that it copes with ariaLabel , labelSuffixA11y or neither being defined
- [ x]  Add a unit test for this

## How to test

1. Go to https://deploy-preview-1784--oak-web-application.netlify.thenational.academy
2. Click on _______
3. You should see _______

### DIFF
--- a/src/components/Button/ButtonAsLink.test.tsx
+++ b/src/components/Button/ButtonAsLink.test.tsx
@@ -1,0 +1,63 @@
+import renderWithTheme from "../../__tests__/__helpers__/renderWithTheme";
+
+import ButtonAsLink from ".";
+
+describe("ButtonAsLink", () => {
+  it("renders a button as a link", () => {
+    const { getByRole } = renderWithTheme(
+      <ButtonAsLink
+        label="Click me"
+        variant="minimal"
+        icon="arrow-right"
+        $iconPosition="trailing"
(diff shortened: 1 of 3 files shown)


k-huggs (approved):
LGTM ✅

PR #1833, opened 2023-08-29
chore(CI): temp disable browser E2E tests in CI

jimCresswell:
Temporarily disable the E2E browser tests in CI for OWA. The current setup was only a POC, it needs more attention to be useful, and will remain flaky until we replace the current consent management solution.


mantagen (approved):
Ideally we remove the code (but also remove /adapt the e2e test suites themselves). But since this is a temporary fix until we re-work the e2e stuff, happy to comment for now. But could we ticket the next phase?

PR #1842, opened 2023-08-30
chore: update msw

jimCresswell:
Attempt to update msw, causes two tests to fail because of apparent changed behaviour in network error handling. After the upgrade the network error seems to because an Oak Error "hubspot/unknown". It's not immediately clear from the hubspotSubmitForm code that this is incorrect behaviour.


mantagen:
@jimCresswell fyi it seems msw changed their message for a networkError ... we now use the (likely more stable) 'name' property to check

mantagen (approved):

PR #1849, opened 2023-08-31
fix: exclude api test fixtures from sonar duplication checks

lioneljingles:
## Description

- Added `src/node-lib/curriculum-api-2023/fixtures/**`  and  `src/node-lib/curriculum-api/fixtures/**` to list of files / folders excluded from Sonar code duplication checks


jimCresswell (approved):
🎉

PR #1855, opened 2023-08-31
Fix/error reported

mantagen:
## Description

- adds `hasBeenReported` to `OakError`
- in `errorReporter` check/set `hasBeenReported` to avoid duplication notifications for the same error


mcky (approved):
Sweet!

PR #1873, opened 2023-09-05
ENG-467 chore: update portabletext/react

mantagen:
## Description

- updates @portabletext/react to latest
- wraps `<PortableText />` injecting default components (since [they've moved away from context pattern](https://github.com/portabletext/react-portabletext/blob/main/MIGRATING.md))
- New component is called `<PortableTextWithDefaults />`, merges components passed as props with a set of default components. There's a prop `withoutDefaultComponents` which does not merge the default components. This was necessary to avoid regression since previously in a few places PortableText was called outside of the context provider. Another option would have been to just keep PortableText without the context in these places. Happy to do that instead.

## Testing

- poratble text should be rendered the same as it was


src/components/PortableText/PortableText.tsx
```diff
@@ -181,12 +184,13 @@ export const basePortableTextComponents: PortableTextComponents = {
...
+export function PortableTextWithDefaults<
+  B extends TypedObject = PortableTextBlock,
+>(props: PortableTextProps<B>): React.JSX.Element {
   return (
-    <PortableTextComponentsProvider components={basePortableTextComponents}>
-      {props.children}
-    </PortableTextComponentsProvider>
+    <PortableText
+      {...props}
+      components={merge(basePortableTextComponents, props.components)}
```
mcky, line 193: Is this merging the components deep enough?

`_.merge` docs say:
> Array and plain object properties are merged recursively.

I thought there was `_.mergeDeep` but I can't find it, so maybe lodash has always just had that behaviour

mcky: Ah I see this PR came after we spoke about regressions, so guessing it's fine and this was the fix

mantagen: Yep i think that's right! Will come back to this PR

johnrobeds (approved):
LGTM

rightsaidjames (approved):

PR #1883, opened 2023-09-07
ENG-407 husky improvements

mantagen:
## Description

- prevents push unless your working tree is clean
- cleans up lint-staged commands
  - only run prettier, eslint, and stylelint on staged files, and commit any changes resulting from --fix(es)


.husky/pre-push
```diff
@@ -1,5 +1,6 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
+git diff-index --quiet HEAD -- || (echo "Error: You have uncommitted changes. Please commit or stash them before pushing." && exit 1)
```
mantagen, line 4: Only allow push with clean head (otherwise pre push checks lose meaning).

.lintstagedrc.js
```diff
@@ -1,10 +1,7 @@
-const path = require("path");
-
-const buildEslintCommand = (filenames) =>
-  `next lint --fix --base-dir . --file ${filenames
-    .map((f) => path.relative(process.cwd(), f))
-    .join(" --file ")}`;
-
 module.exports = {
-  "*.{js,jsx,ts,tsx}": [buildEslintCommand, "npm run lint:styles"],
+  "*.{js,jsx,ts,tsx}": [
```
mantagen, line 2: "format" then "lint" then "lint styles" on changed files

.husky/pre-merge-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npm run check-types
```
mantagen, line 4: Checks types on merge commit.

Does it make more sense to **not**  run checks on merge commit, and  leave it to pre-push and CI?
- it won't be often that merge commits cause checks to fail, so letting them fail pre-push/CI is probably okay
- if this pre-merge-commit fails it could leave working tree in a potentially fiddly state

kimon-satan: Yes it's probably okay now that we can't push unclean trees

kimon-satan (approved):

PR #1886, opened 2023-09-08
chore(jest config): increase test timeout to 15s

mantagen:
## Description
- bumps jest timeout to 15 s, because in CI too frequently a test will timeout, and so fail, despite nothing being broken.
- obvsiouly the "ideal" fix is to have more efficient tests ...  but it's not so simple with testing-library


johnrobeds (approved):
Just double check percy visuals (shouldn't be touched by this) but happy with the PR.

PR #1908, opened 2023-09-15
Feat/lesq 291/hamburger update [LESQ-291]

njblore:
## Description
Ended up being a few changes, key ones to review are:
- I added `early-realease-units` to `urls.ts` in order to be able to link to that page from the hamburger menu - I'm not sure if this has any ramifications or should have been done differently
- created new types for the new menu links rather than try and shoe horn into the existing ones
- created new components for the hamburger menu sections and links, imaginatively called `NewMenuLink` and `NewMenuSections` - better suggestions on a postcard
- replaces the `betaMenuSections` with updated links and replaced them in `AppHeader` - I believe this means it won't be visible on the old page and only the new ones so correct me if wrong
- updated some styles

For Monika:
- I couldn't find an existing button that matched the style in the design for `hover` and `focus` exactly, I used `ButtonWithIcon` in the variant `buttonStyledAsLink`, but could alternately use the`minimal` variant which[ looks like this ](https://storybook.thenational.academy/?path=/story/buttons-button--button-with-icon/path=/story/buttons-button--button-with-icon&args=label:Home;variant:minimal) but with the icon at the end.

## How to test

1. Go to https://deploy-preview-1908--oak-web-application.netlify.thenational.academy/teachers-2023
3. You should see the new design and all the links should go to the correct place

## Screenshots

Screenshot 2023-09-15 at 12 19 51
  
  
How it should now look:
Screenshot 2023-09-15 at 12 20 46

## ACs
- [ ]  Scrolls if it doesn’t fit
- [ ]  New links
    - [ ]  Hover
    - [ ]  Focus
- [ ]  “New” next to early release units
- [ ]  Opens in a new tab icon on some of the links
- [ ]  New “X” logo for Elon Musk


src/browser-lib/avo/Avo.ts
```diff
@@ -1141,6 +1141,7 @@ export const PageName = {
   'CURRICULUM_OVERVIEW': 'Curriculum Overview',
   'CURRICULUM_UNIT_SEQUENCE': 'Curriculum Unit Sequence',
   'CURRICULUM_DOWNLOADS': 'Curriculum Downloads',
+  'EARLY_RELEASE_UNITS': 'Early Release Units'
```
JBR90, line 1144: Did you add this or was it automatically added from avo?

src/components/NewMenuSections/types.ts
```diff
@@ -0,0 +1,21 @@
+import { ResolveOakHrefProps } from "@/common-lib/urls";
+
+type HrefLink = { href: string };
+
+export const linkIsHref = (input: unknown): input is HrefLink => {
+  return typeof (input as HrefLink).href === "string";
+};
+
+export type BetaMenuLink = {
```
JBR90, line 9: I wonder if we should change the naming away from beta, as it's not going to exist much longer

src/common-lib/urls/urls.ts
```diff
@@ -603,6 +610,12 @@ export const OAK_PAGES: {
     configType: "internal",
     pageType: "curriculum-downloads",
   }),
+  "early-release-units": createOakPageConfig({
+    pathPattern: "/beta/:viewType/early-release-units",
```
JBR90, line 614: We should move this into teachers and out of beta

src/components/AppHeader/AppHeader.tsx
```diff
@@ -46,7 +47,7 @@ const AppHeader: FC<HeaderProps> = () => {
           onClick={openMenu}
         />
         <Menu menuButtonRef={menuButtonRef}>
-          <MenuLinks menuSections={betaMenuSections} />
+          <NewMenuSections menuSections={betaMenuSections} />
```
JBR90, line 50: Does NewMenuSection replace MenuLinks ? If it's not used anywhere else maybe this should just be MenuSections

JBR90 (changes requested):
Code looks great, few comments/questions on naming

src/components/AppHeader/AppHeader.tsx
njblore: It does here but I think `MenuLinks` is still used in the hamburger menu for the non-beta pages

src/components/NewMenuSections/types.ts
njblore: yeah I agree it could be better named! any suggestions?

src/browser-lib/avo/Avo.ts
njblore: I added this as in `url.ts` it required a value for `AnalyticsPageName` value - I have no idea how these values are used though or if there is a knock on effect in the analytics to be mindful of?

Monika-Koziol (changes requested):
it's looking great. Just a couple of small questions/comments

- [x] The links on the mobile version all had hover state rather than default. Could that be removed?

- [x] I've noticed that once navigating to blog pages and Oak section the hamburger changed to the old style. I presume it's only for now but wasn't sure how the change will happen once we go live
- [x] No link to curriculum page yet. Will that be implemented as a separate ticket?

Screenshot 2023-09-15 at 14 37 56

njblore:
@Monika-Koziol thanks, the curriculum link should be working now I just missed it accidentally! 
  
My understanding is there will need to be a ticket to swap the old hamburger for this new one when we launch - I don't think we can use the new one on the old pages yet as there are links we don't want to make public, but correct me if this is wrong!
  
For the button underline on mobile, I can make that change but it will also affect the filters on the search page as it's using the same button style, is that ok? 
Screenshot 2023-09-15 at 15 39 58

Monika-Koziol:
@njblore thanks for the explanation. Yeah it is all good to remove the line even on the search results. Swapping the menu once we are ready to launch makes total sense. Thanks for adding external icon for help.

Monika-Koziol (approved):

jack-skerrett-oak (changes requested):
Heyo 
- I noticed that the hover state and focus state seem to be the same on your PR. Not the end of the world but worth fixing if it is easy (figma: https://www.figma.com/file/YcWQMMhHPVVmc47cHHEEAl/Oak-Design-Kit?type=design&node-id=342-7129&mode=dev)  
Screenshot 2023-09-18 at 09 21 13
- [x] It would be good to add aria-labels or something similar to the links that open in a new tab to warn screen reader users in advance https://www.w3.org/TR/WCAG20-TECHS/G200.html

njblore:
thanks @jack-skerrett-oak I've made those changes - for the aria-label I've added `(opens in a new tab)` to external links, do you think that will suffice?

src/components/NewMenuSections/types.ts
njblore: renamed to `BurgerMenuLink` hopefully that's a bit more descriptive

jack-skerrett-oak (approved):

src/browser-lib/avo/Avo.ts
JBR90: Avo.ts is generated automatically so we shouldn't edit it directly. I suggest removing this for now and writing a ticket to add  EARLY_RELEASE_UNITS to AnalyticsPageName in Avo. Maybe check in Verity but I don't think it's a priority to add tracking to the early release page now.  Happy to pair and go through this in more detail if it is. Here are the docs.

https://www.notion.so/oaknationalacademy/Avo-5ff21cbd33994c84907c83c6af5a25ee

njblore: @JBR90 Aha I see thanks! I think since [Ben's PR](https://github.com/oaknational/Oak-Web-Application/pull/1909/files#diff-89a5d358f9c3d0fd56861e3060f31d592b8ad78a406f5b393f8eb861f5c47c62) this change is already in main now so is no longer showing in the latest diff - presumably no new ticket will be required?

JBR90 (approved):

PR #1941, opened 2023-09-27
fix: maths ks4 curriculum map download - LESQ-332

JBR90:
## Description

- Fixes ks4, maths, legacy curriculum map download (zip on programme page)
- The zip file is downloaded and the user remains on the page
- refactored the error message so it works for both buttons.

## Issue(s)

Fixes #

## How to test

1. Go to https://deploy-preview-1941--oak-web-application.netlify.thenational.academy/teachers/key-stages/ks4/subjects/maths-l/programmes
(should download a zip)
check other downloads are working

## Checklist

- [ ] Design sign-off


jack-skerrett-oak (approved):
Seems to work much better now :)

src/components/CurriculumDownloadButtons/CurriculumDownloadButton.tsx
```diff
@@ -71,9 +71,9 @@ const CurriculumDownloadButton: FC<CurriculumDownloadProps> = ({
   };
 
   return (
-    <Flex>
-      {lessonPage && keyStageSlug === "ks4" && subjectSlug === "maths" ? (
-        <Flex $flexDirection={"column"}>
+    <>
+      <Flex>
+        {!tier && keyStageSlug === "ks4" && subjectSlug === "maths" ? (
```
njblore, line 76: I know this isn't new but it would be helpful to pull this kind of long conditional out into a named variable for better readability - something like `isKs4MathsWithoutTier`? Even better if it could explain why we're doing this (as I'm not really sure hah)

JBR90: Yeah good point, have added to a variable and left a comment.

If we are on a Maths programme page, we want to download a zip of foundation, core, higher curriculum map .pdfs. This is pretty brittle atm but will be removed shortly after the launch as curriculums maps are updated.

njblore (approved):

PR #1943, opened 2023-09-27
ENG-481 build(image data queries): update to filter assets with full data

mantagen:
## Description

- filters assets by published + existence of necessary properties such as url/slug
- this should further prevent issues we saw before of incomplete assets causing issues in builds

## How to test

1. Go to https://deploy-preview-1943--oak-web-application.netlify.thenational.academy
2. Click around
3. You should see no change (and percy shouldn't either!)


jimCresswell (approved):
Looks good... having to make that change several times in the consuming app makes me feel like maybe there is a back-end change we could make longer-term, but I have no idea what yet

k-huggs (approved):
LGTM, no assets appearing broken on the site ✅

PR #1957, opened 2023-10-02
feat: point to new curriculum units mv

lioneljingles:
## Description
- Update API endpoint for curriculum units data to point to new MV with lessons updated filtering

Note: This PR does not include showing the lessons in the UI.


rightsaidjames:
Still lots of duplicate lessons in Y10/11, although I can see that some of the duplicates (e.g. DNA and the genome) have gone away.

Screenshot 2023-10-02 at 11 25 17

I realise that this is not something you can fix in this PR, though.

k-huggs (approved):
LGTM ✅

PR #1960, opened 2023-10-02
feat(apis): generate mux tokens

jimCresswell:
create an API route that takes a Mux ID and asset type and returns a token for a signed Mux URL

See https://www.npmjs.com/package/@mux/mux-node

## To Do

- [ ] Add the Mux signing key and signing secret to the server config

### DIFF
--- a/src/node-lib/getServerConfig.ts
+++ b/src/node-lib/getServerConfig.ts
@@ -143,6 +143,20 @@ const envVars = satisfies<Record<string, EnvVar>>()({
     availableInBrowser: false,
     default: null,
   },
+  muxSigningKey: {
+    value: process.env.MUX_SIGNING_KEY,
+    envName: "MUX_SIGNING_KEY",
+    required: true,
+    availableInBrowser: false,
+    default: null,
+  },
+  muxSigningSecret: {
+    value: process.env.MUX_SIGNING_SECRET,
+    envName: "MUX_SIGNING_SECRET",
+    required: true,
+    availableInBrowser: false,
+    default: null,
+  },
 });
 
 for (const [, envVarConfig] of Object.entries(envVars)) {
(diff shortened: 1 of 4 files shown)


src/node-lib/getServerConfig.ts
```diff
@@ -143,6 +143,20 @@ const envVars = satisfies<Record<string, EnvVar>>()({
...
+  muxSigningKey: {
+    value: process.env.MUX_SIGNING_KEY,
+    envName: "MUX_SIGNING_KEY",
+    required: true,
+    availableInBrowser: false,
+    default: null,
+  },
+  muxSigningSecret: {
+    value: process.env.MUX_SIGNING_SECRET,
+    envName: "MUX_SIGNING_SECRET",
```
njblore, line 155: just thinking about how we handle legacy vs new videos - do we need two sets of keys for the different MUX envs? or continue to use the current api for old videos and this one for new? 🤔

jimCresswell: Good shout, I've added a query param switch `twenty_twenty` to switch between credential sets. Feel free to give it a better name

jimCresswell:
@njblore , we're going to get this in as-is to enable testing and refinement. The 2020 Mux config still needs to be added to the secret manager

src/pages/api/video/signed_url/index.ts
```diff
@@ -0,0 +1,67 @@
...
+    response
+      .status(400)
+      .json("Must provide a single id and a single type query parameter");
+    return;
+  }
+
+  const use2020 = twenty_twenty === "true";
+  const baseOptions = use2020 ? auth2020 : auth2023;
+
+  const token = Mux.JWT.signPlaybackId(id, {
```
jimCresswell, line 58: needs some error handling

mantagen (approved):

PR #1970, opened 2023-10-04
[ENG-471] more data in auth errors for CMS preview API calls

jimCresswell:
We're getting hundreds of invalid auth request a day, possibly Detectify, hence extra data to help with analysis.


mcky (approved):

mantagen (approved):

src/pages/api/preview/[[...path]].ts
```diff
@@ -27,6 +27,10 @@ const preview: NextApiHandler = async (req, res) => {
     if (req.query.secret !== getServerConfig("sanityPreviewSecret")) {
       throw new OakError({
         code: "preview/invalid-token",
+        meta: {
```
Comment (mantagen) [inline src/pages/api/preview/[[...path]].ts:30]: TBH we should be adding this stuff on all the server side errors!

PR #1973, opened 2023-10-04
ENG-474 fix: handle downloads responses with zod and send error data to bugsnag

mantagen:
## Description

- amend download api helpers
  - use zod to parse responses for better type safety (and for clarity on response schema)
    - **note**: all points at the edge of the app (i/o, network requests etc) should use zod, for type parsing
  - append context (meta data) to error when reporting to bugsnag (for easier debugging)
- reset api error message onSubmit 

## How to test

1. Go to https://deploy-preview-1973--oak-web-application.netlify.thenational.academy/teachers/lessons/transverse-waves/downloads?preselected=all
2. disable network in browser
3. attempt download
4. should see error
5. enable network
6. attempt download
7. error should be cleared and download should work


JBR90 (approved):
Yep this seems much more secure, works as expected

PR #1978, opened 2023-10-05
Feat/lesq 363/signed video ids

njblore:
## Description

*In draft until the data has been updated*

- Removes the temporary measure to use public IDs for new video content
- Integrates the new api endpoint to generate tokens for videos
- Integrates the updated MV to retrieve signed video IDs for mux

### DIFF
--- a/src/components/CMSVideo/CMSVideo.tsx
+++ b/src/components/CMSVideo/CMSVideo.tsx
@@ -20,6 +20,7 @@ const CMSVideo: FC<CMSVideoProps> = ({ video, ...rest }) => {
       thumbnailTime={video.video.asset.thumbTime}
       playbackId={video.video.asset.playbackId}
       title={video.title}
+      isLegacy={true}
       {...rest}
     />
   );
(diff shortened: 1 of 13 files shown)


kimon-satan (approved):

jack-skerrett-oak (approved):
I'm not sure what you mean about legacy secrets :D But...

- Webinar videos work
- Videos on the home pages work
- Old lesson videos work
- New lesson videos work (and are coming from our custom domain)

So hopefully everything is OK...!

njblore:
@jack-skerrett-oak oh yeah sorry that wasn't super clear was it 😅.  
What I was getting at is we are now fetching lessons using two different sets of credentials depending on whether they are legacy (using the OG `Production` mux environment) or they are new (using the new Mux environment `CAT Ingest Production`)   
So I've made the assumption that webinars/blogs etc will still live in the OG Mux environment and use those credentials, but if there is a plan to upload them to the new mux environment that will become a problem! and we need to know about it

PR #2006, opened 2023-10-11
Feat/lesq 354/card hover updates [LESQ-354]

njblore:
## Description

- Updated background colour on hover for: `ListItemCard`, `ListItemIndexDesktop, `ListItemIconDesktop` and `ProgrammeListItem`
- Removed transform on hover from `ProgrammeListItem`, `ListItemCard` and `ListItemIndexDesktop` and removed drop shadow change
- Added an override for early release units as they use custom coloring
- Added tests for the colour changes

## How to test

1. Go to https://deploy-preview-2006--oak-web-application.netlify.thenational.academy/teachers/programmes/english-secondary-ks3/units
2. Hover over the unit
3. You should see the new hover state
4. Click through to lessons and hover

## Screenshots

Screenshot 2023-10-11 at 09 16 08

How it should now look:

Screenshot 2023-10-11 at 09 15 45

## ACs
- [ ]  Update hover state so the lesson/unit title gets underlined and add colour transition (a bit like in [[here](https://www.w3schools.com/howto/tryit.asp?filename=tryhow_css_buttons_fade1)](https://www.w3schools.com/howto/tryit.asp?filename=tryhow_css_buttons_fade1)) to the coloured box (blue one on the unit listing and pink on the lesson listing)
- [ ]  On hover make pink/blue background darker, change white background to a very light grey, see Figma for values, underline main heading only
- [ ]  Apply to unit listing card
- [ ]  Apply to search results card
- [ ]  Apply to lesson listing card
- [ ]  Apply to card on programmes page (the one with tiers)


Monika-Koziol (changes requested):
@njblore looking good. One little change on the early-release page the coloured backgrounds (pink and blue) don't change. Is it possible to update those as well so it's consistent?

src/styles/theme/default.theme.ts
```diff
@@ -137,6 +138,7 @@ const theme: OakTheme = {
     pink: "black",
     pink50: "black",
     pink30: "black",
+    pink60: "black",
```
kimon-satan, line 141: isn't this rather strange ?

kimon-satan (approved):
Maybe we should create a ticket to work out what is going on in default.theme.ts ?

njblore:
@Monika-Koziol yes of course! I left them as their current colour is very similar (possibly the same?) as the new hover - what should the new values for their hover colours be?

src/styles/theme/default.theme.ts
njblore: from what I gather this is used to decide whether an icon will have a white or black fill against whatever colour background, but could def use some clarity!

Monika-Koziol:
@njblore that's a good point. The limited tonnes can be limiting. The pink 60 should work as it's a new darker shade anyway you can create lavender-60 with #8FA6E4.

jack-skerrett-oak (approved):
I'm happy once Monika is happy :)

PR #2012, opened 2023-10-12
feat: add homeschool option in school picker LESQ-374

JBR90:
## Acceptance Criteria:

- [ ]  Add an option into the list of schools for "Homeschool"
- [ ]  Remove radio button from the list below

## Issue(s)

Fixes #

## How to test

1. Go to https://deploy-preview-2012--oak-web-application.netlify.thenational.academy/teachers/programmes/combined-science-secondary-ks4-higher-ocr/units/measuring-waves/lessons/transverse-waves/downloads?preselected=all
Or any download page, radio buttons have gone, school picker has homeschool option

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist

- [ ] Design sign-off


njblore (approved):
Looks good! just a minor syntax quibble but not blocking

jack-skerrett-oak (commented):
I am only commenting as I won't be around to see these comments resolved, you'll need to point this at James!

- The data being sent to hubspot is different when you select homeschool. Might need to chat with Simon Hacking about how disruptive this would be. On this PR hubspot receives schoolname = "Homeschool" vs "homechool" currently. On this PR we also send through a URN of "home", I don't know how destructive sending through a string URN will be.

Behaviour looks good other than that (and the above might turn out to be a non-issue anyway, but I think it would be best to either make it behave the same, or if we can't, at least have that chat with Simon)

Monika-Koziol (changes requested):
@JBR90 one small change please. The quotation marks around homeschool are not right. Can you can you try those: ‘homeschool’. On mac to type them you need to  do this: 

- [ ] left one `option + right square bracket` 
- [ ] right one `option + shift + right square bracket`

They should look like this: 
Screenshot 2023-10-13 at 13 54 26

JBR90:
> initialValue === "notListed"

JBR90:
> I am only commenting as I won't be around to see these comments resolved, you'll need to point this at James!
> 
> * The data being sent to hubspot is different when you select homeschool. Might need to chat with Simon Hacking about how disruptive this would be. On this PR hubspot receives schoolname = "Homeschool" vs "homechool" currently. On this PR we also send through a URN of "home", I don't know how destructive sending through a string URN will be.
> 
> Behaviour looks good other than that (and the above might turn out to be a non-issue anyway, but I think it would be best to either make it behave the same, or if we can't, at least have that chat with Simon)

Ok think i've sorted this now, should match what we had before in local storage / hubspot / posthog
@rightsaidjames could you take a look when you get a sec?

Monika-Koziol (approved):
Looking good.

rightsaidjames (changes requested):
Two things:
* Can I confirm that the `homeschool` option isn't meant to have an accompanying `urn` number sent to Hubspot?
* When editing my details after selecting the `Homeschool` option and downloading some files, the school name is displayed as `homeschool` (lower case). I've confirmed with @Monika-Koziol that the first letter should still be capitalised.

rightsaidjames (approved):
Looking good:
Screenshot 2023-10-17 at 12 44 44

And to confirm, the current behaviour is that the `homeschool` name does not have an associated `urn`, unlike the other 'actual' schools that can be selected from the list.

src/components/SchoolPicker/SchoolPicker.tsx
```diff
@@ -46,17 +46,19 @@ const SchoolPicker: FC<SchoolPickerProps> = (props) => {
...
+          ? item.urn
+          : `${item.urn}-${item.name}`;
+        const comboItem =
+          item.urn === HOMESCHOOL_URN
+            ? item.name
+            : formatSchoolName(
+                `${item.name}, ${item.la}, ${item.postcode}`,
+                props.schoolPickerInputValue,
+              );
+        return <Item key={comboItemKey}>{comboItem}</Item>;
```
benprotheroe, line 60: i think if youre going to have a function rendered inside the item tag you need to include the "testValue" prop

PR #2021, opened 2023-10-13
ENG-489 feat: handle preview errors

mantagen:
## Description

- I'd wanted this work to result in the user (whoever is editing content on CMS) to be shown nicely displayed errors telling them what's wrong
- However, this wasn't possible, so this simply stops such errors being reported to BugSnag
- @mcky it turns out that the errors we were getting in BugSnag were valid errors (i.e. schema in OWA did't match config in Sanity). So maybe it's best to not merge this and instead to actually treat BugSnag errors from PreviewMode seriously?

### How to test

- locally you can change a schema and check a preview version and see that it doesn't attempt to report


src/node-lib/getPageProps.ts
```diff
@@ -62,14 +63,30 @@ const getPageProps = async <PageProps>({
...
+    if (context.preview && error instanceof ZodError) {
+      oakError = new OakError({
+        code: "preview/zod-error",
+        originalError: error,
+        meta: {
+          errors: error.errors,
+        },
+      });
+    }
+
```
mcky, line 80: @kimon-satan 
I wonder if instead of trying to reshape the errors here to avoid throwing, if we should add a flag to all error metadata that says if it's in preview mode.

We could then filter them and the alerts in the bugsnag UI so they don't create noise but we still have the error data there for debugging

It also feels like this reshaping into an OakError might be better in the actual CMS layer than `getPageProps`

kimon-satan:
Superceded by 2050

PR #2031, opened 2023-10-17
Feat/lesq 395/remember homepage tab [LESQ-395]

njblore:
## Description

- update url with hash when selecting tabs on the homepage in order to store the state for browser navigation
- let me know if there is a more idiomatic nextjs way to do this! 

## How to test

1. Go to https://deploy-preview-2031--oak-web-application.netlify.thenational.academy
2. Click on curriculum tab
3. follow a link to another page (eg early release units)
4. click back on the browser
5. You should see the curriculum tab is active

## ACs

- [x]  The app ‘remembers’ which homepage tab the user came from when the user returns to the homepage via the back button.
- [x]  No extra browser history entries are added as a result of this change.

### DIFF
--- a/CHANGE_LOG.md
+++ b/CHANGE_LOG.md
@@ -1,3 +1,18 @@
+# [1.282.0](https://github.com/oaknational/Oak-Web-Application/compare/v1.281.0...v1.282.0) (2023-10-13)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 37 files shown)


rightsaidjames (approved):
Looks good - all AC met (along with Sarah's use case). Tested in Chrome (desktop) and iOS Safari.

JBR90 (approved):

PR #2037, opened 2023-10-19
Fix/seo

jimCresswell:
Allow the now home page and early-release-units page to be indexed


mcky (approved):
Approving for the sake of getting it out asap

PR #2050, opened 2023-10-26
Feat/catch and log preview errors

kimon-satan:
## Description

Zod Errors occuring whilst in Sanity's preview mode were causing clutter in bugsnag. This change catches such errors as the data is parsed and gives them a code which allows the errorReporter to filter and avoid reporting them to bug snag.

- new preview/zod-error error code
- addition to errorReporter to prevent reporting errors flagged as shouldNotify = false
- catch error and wrap in src/node-lib/cms/sanity-client/parseResults.ts

## How to test

locally you can change a schema and check a preview version and see that it doesn't attempt to report


src/common-lib/error-reporter/errorReporter.ts
```diff
@@ -208,6 +208,11 @@ const errorReporter = (
         return;
       }
 
+      if (maybeError instanceof OakError && !maybeError.config.shouldNotify) {
+        logger.log("Error should not be reported, aborting reportError()");
+        return;
```
mcky, line 213: I'd _err_ towards logging out the actual error here too?
```suggestion
        logger.log("Error should not be reported, aborting reportError()");
        logger.error(maybeError);
        return;
```

src/node-lib/cms/sanity-client/parseResults.ts
```diff
@@ -89,34 +91,54 @@ export const parseResults = <S extends ZodSchema, D>(
...
-    if (isZodArraySchema(schema)) {
-      /**
-       * Take the provided schema and create a version of it that
-       * will silently filter out any items that are invalid
-       */
-      const invalidRejectingSchema = createInvalidRejectingSchema(schema);
-      const parsedItems = invalidRejectingSchema.parse(data);
+    try {
+      if (isZodArraySchema(schema)) {
+        console.log("duh");
```
mcky, line 96:
```suggestion
```

mcky (approved):

PR #2051, opened 2023-10-26
Feat/add ai page to homepage

tomwisecodes:
- Adds a new tab to the home page tabs for ai.
- Adds a new hero for that tab linking out to the labs.thenational.academy 
- Adds labs as a possible route for oak pages


mcky (approved):
Approved code-wise **PENDING** a decision about the analytics from the team

src/pages/index.tsx
```diff
@@ -47,7 +48,7 @@ export type HomePageProps = {
   posts: SerializedPost[];
 };
 
-export type HomePageTab = "teachers" | "curriculum" | "pupils";
+export type HomePageTab = "teachers" | "curriculum" | "pupils" | "ai";
 const isHomePageTab = (u: unknown): u is HomePageTab =>
```
kpopper, line 52: Does `isHomePageTab` need to be set if the tab is `ai`?

rightsaidjames: I think this is needed to take the user back to the AI tab when they press the back button from another page.

tomwisecodes: Good flag, I think this persists the same state when the page reloads. Will update

rightsaidjames (approved):
Looks good. I'm assuming the illustration to the right of the heading and copy text (like in other tabs) is still to be added in?

kpopper (approved):

PR #2058, opened 2023-10-30
Feat/lesq 411/loading button [LESQ-411]

njblore:
## Description

- Add a new button `LoadingButton` using new button design from figma
- Add a new button border component `DoubleButtonBorders` which adds an additional double outline on focus
- Replace button on the download with the new LoadingButton
- Disable button on downloads page when there are form errors or an incomplete form

## How to test

1. Go to https://deploy-preview-2058--oak-web-application.netlify.thenational.academy/teachers/programmes/combined-science-secondary-ks4-foundation-ocr/units/measuring-waves/lessons/transverse-waves/downloads?preselected=all
2. Click on Download
3. You should see the button update with a loading state and revert on success
5. Also check the component  on storybook and toggle between loading/enter different text - it's in the Button section and called `Loading`

## Screenshots

Screenshot 2023-10-30 at 13 14 34

How it should now look:
Screenshot 2023-10-30 at 13 15 16
Screenshot 2023-10-30 at 15 13 09
Screenshot 2023-10-30 at 15 17 35

## ACs
- [ ]  Use disabled button style (in grey) with an animated spinner (animation like the one in the second row, in the middle here: https://loading.io/css/)
- [ ]  Accessibility: if user has motion reduction turned on, ensure the animation doesn’t work
- [ ]  Button copy to say ‘`Downloading…`’
- [ ]  Button has a `disabled` attribute on it, so it can’t be clicked while in the loading state, and this is communicated as such to screen readers and other assistive tech
- [ ]  Button is re-enabled after a successful (or failed) download, with its original button text


JBR90 (approved):
Looks good to me

Monika-Koziol (changes requested):
Looking great and we have a spinner! 
- [ ] Did we agree this morning to have a disabled state when the errors need correcting? If I got this wrong please ignore this 
- [ ]  It also needs a focus state for accessibility, especially when tabbing.

Monika-Koziol (approved):
Nice one!

jack-skerrett-oak (changes requested):
- When I nav to a downloads page with no errors present, I cannot download anything until I interact with an element on the page
- I am going to be slightly annoying and say that in an ideal world, the disabled button should still be focusable, and tell the user why it is dimmed. I think with the context around the button we don't need to worry about the second bit, but making it focusable would be useful (my reasoning for this is that someone using a screen reader or keyboard nav in general might not realise there is a download button and get lost on the page. If they navigate and it is disabled, it gives them information that there is something blocking them from downloading)

njblore:
@jack-skerrett-oak for the disabled button focus state should that be for when the button is disabled for any reason (ie. when loading) or only when it's disabled due to incomplete form?

jack-skerrett-oak (approved):

PR #2088, opened 2023-11-09
Feat/lesq 442/generate share links [LESQ-442]

njblore:
## Description

- Adds helper methods and data types for generating links to be used in sharing components
- Adds new icons for sharing

## How to test
There is no front end to test this with currently, only the unit tests

## ACs

- [ ]  unit tests in place to prove link generation works for each route


JBR90 (approved):
Looks good, we can test it more thoroughly once it's hooked up

PR #2116, opened 2023-11-29
feat: add legacy curriculum downloads

lioneljingles:
## Description

Music year: 1956

- Create new module for Curriculum Downloads
- Create page for previous (or legacy) curriculum document downloads
- Create fixture to store previous curriculum download data

## How to test

1. Go to https://deploy-preview-2116--oak-web-application.netlify.thenational.academy/teachers/curriculum/previous-downloads
2. You should that the `EYFS` tab is selected by default.
3. Click on the tab `KS1`
4. You should see the tab updated to show relevant downloads.
5. Click on "Music"
6. Fill in the "Your details" section.
7. Click "Download PDF".
8. The Music curriculum document PDF should download.

## Screenshots

How it should now look:
Screenshot 2024-02-02 at 11 07 48
Screenshot 2024-02-02 at 11 07 58
Screenshot 2024-02-02 at 11 08 07

## Checklist

- [ ] Design sign-off


HelenHarlow (changes requested):
Heya, looks great! A few things from me:

- Can we make the download tiles a little closer together? 12px both horizontally & vertically between them 
- Could the spacing between the T&C’s copy and the download button be 16px?
- For me, the header section breaks on mobile because the tabs exceed the width of the screen. In design, I had looked at the tabs becoming a dropdown on smaller screens. Would this be possible?
- I noticed that when I select one of the cards, the form immediately generates errors. From what I can tell, I think this is happening because we’re using the same pattern as other downloads pages? I think ideally, the button wouldn’t be disabled, and errors would appear only if the user clicks ‘download’ without entering mandatory fields, but not sure what you think about that?

rightsaidjames:
Overall, it looks great!

@lioneljingles Found a juicy edge case bug:

1. Select a subject available in multiple key stages
2. Switch to a different key stage
3. Note that the subject is pre-selected
4. Click the download button

**Expected:** The correct document (currently-selected key stage and subject) is downloaded

**Actual:** The incorrect document (wrong key stage, right subject) is downloaded.

For example, select KS1, then History, then switch to KS2 and click the Download button: `key-stage-1-history.pdf` will be downloaded.

My suggested fix for this would be to deselect the chosen resource when switching between phases. This would also fix the edge case of switching from one phase to another where the previously-chosen subject is not available in the new phase.

https://github.com/oaknational/Oak-Web-Application/assets/3018395/bede2395-f6ee-4cdf-a39d-304dec5f1d70

HelenHarlow (approved):

lioneljingles:
> Overall, it looks great!
> 
> @lioneljingles Found a juicy edge case bug:
> 
>     1. Select a subject available in multiple key stages
> 
>     2. Switch to a different key stage
> 
>     3. Note that the subject is pre-selected
> 
>     4. Click the download button
> 
> 
> **Expected:** The correct document (currently-selected key stage and subject) is downloaded
> 
> **Actual:** The incorrect document (wrong key stage, right subject) is downloaded.
> 
> For example, select KS1, then History, then switch to KS2 and click the Download button: `key-stage-1-history.pdf` will be downloaded.
> 
> My suggested fix for this would be to deselect the chosen resource when switching between phases. This would also fix the edge case of switching from one phase to another where the previously-chosen subject is not available in the new phase.
> legacy.downloads.bug.mp4

Good find. Should be resolved now.

rightsaidjames (approved):
Looks good. I have confirmed that Avo tracking events and Hubspot requests are sent at the point of Download.

Screenshot 2024-02-08 at 09 27 58
Screenshot 2024-02-08 at 09 28 24
Screenshot 2024-02-08 at 09 29 48

PR #2131, opened 2023-12-13
I feat/iwant to show duplicate lessons and units in drop down LESQ-462

JBR90:
## Description

If a lesson or unit has duplicate results (same unit, different paths), then don’t display all those duplicates, instead hide them within a drop down on a card that contains the information of ALL The units/lessons 

https://www.notion.so/oaknationalacademy/I-want-owa-to-be-able-to-handle-duplicate-lessons-and-units-within-a-drop-down-b9fdf12bd68d4d59868d7da9779350fb

Music year: 2011

- Adds pathway to legacy and new API
- Adds dropdown to search hit card when there is more than one pathway in a search hit  - _add &flag=enable_2023_search_api to see results from new search api_
- updates tests

## How to test

1. Go to https://deploy-preview-2131--oak-web-application.netlify.thenational.academy
Search should work as before, you should see drop down for search hits with more than one pathway. Waves is a good one to test.

## Checklist

- [ ] Design sign-off


jack-skerrett-oak (approved):
I'm not going to block the PR because of it, but the component does seem to be dead in storybook for me https://deploy-preview-2131--owa-storybook.netlify.thenational.academy/?path=/story/foundations-searchdropdown--search-dropdown

k-huggs:
When tabbing through, for some reason the `MiniDropDownComponent` has lost it's highlighting


https://github.com/oaknational/Oak-Web-Application/assets/91190841/1729ff32-cf0c-45b2-98e2-5bcaef7d825a

src/components/SearchComponents/Search.page.test.tsx
```diff
@@ -297,4 +368,61 @@ describe("Search.page.tsx", () => {
 
     expect(setSearchStartTime).toHaveBeenCalledTimes(1);
   });
+  test("searchResultClicked is called when a search hit is clicked", async () => {
+    const { getByText } = render(<Search {...props} {...resultsProps} />);
+    const description = getByText("lesson title");
+    const user = userEvent.setup();
+    await act(async () => {
```
k-huggs, line 375: I don't think you need to wrap this in an act, you should be able to just await the user.click

src/components/SearchComponents/Search.page.test.tsx
```diff
@@ -297,4 +368,61 @@ describe("Search.page.tsx", () => {
...
+      unitSlug: "topic-slug",
+    });
+  });
+  test("searchResultClicked is called when a pathway hit is clicked", async () => {
+    const { getByText } = render(
+      <Search {...props} {...resultsPropsPathWays} />,
+    );
+    const dropdown = getByText("Select exam board");
+    const user = userEvent.setup();
+    await act(async () => {
```
k-huggs, line 402: same comment as above: 
https://flyingsquirrel.medium.com/when-should-i-use-act-in-react-testing-library-d7dd22a3340e

src/components/SearchDropdown/SearchDropdown.stories.tsx
```diff
@@ -22,8 +22,10 @@ type Story = StoryObj<typeof Component>;
...
   args: {
-    dropdownTitle: "Select exam board",
-    dropdownContent: unitListData,
+    label: "Select exam board",
+    ...searchResultsData,
   },
   render: (args) => <Component {...args} />,
 };
+
+// SearchDropdown.storyName = "MyComponent";
```
k-huggs, line 31: 🔥

src/components/SearchComponents/Search.page.test.tsx
k-huggs: Also might be worth double checking that this is still passing with a .only, for some reason it failed when I did

k-huggs (commented):
Some really minor changes

src/components/SearchComponents/Search.page.test.tsx
JBR90: I added this to suppress the waring in the test 
```
 Warning: An update to ForwardRef(LinkComponent) inside a test was not wrapped in act(...).
    
    When testing, code that causes React state updates should be wrapped into act(...):
    
    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */
```

Adding the act removes the warning

k-huggs (approved):

rightsaidjames:
In terms of functionality, accessibility, usability etc. this looks good - I'll leave design compliance and consistency with other parts of the teacher experience to teacher squad members to rule on though.

One edge case I've found (based on the data currently available to us):
https://deploy-preview-2131--oak-web-application.netlify.thenational.academy/teachers/search?term=surds&flag=enable_2023_search_api

Maths doesn't have exam boards, but individual units will vary by tier. The 'Select exam board' dropdown is displayed for the Surds unit and the first lesson (which is the only lesson present in the Foundation tier), but the dropdown does not have any options:
Screenshot 2023-12-20 at 16 27 25

Suggest an alternate code path for 'choose tier' that doesn't present exam board choices **but** does present a tier choice.

rightsaidjames (approved):
Edge case fix (and all pre-existing changes) looks good to me, so assuming the tests are fixed then 👍 from me.

Monika-Koziol (changes requested):
Some changes from me:
- [x] can the whole component be clickable? At the moment only the chevron is.  
- [x] on expanded state the chevron should change colour to the same shade as the text
- [ ] The focus state seems a bit too strong I imagined it just for the keyboard tabbing. Maybe use the standard double line instead in this case?

rightsaidjames:
You could investigate `:focus-visible` to see if that will only show the existing focus style to keyboard and other AT users, but that does leave you at the mercy of browser vendors and doesn't always work as intended.

Monika-Koziol:
@rightsaidjames I was looking at BBC https://www.bbc.co.uk/, and that's what I had in mind. I'm curious to know how they've done it, though.

JBR90:
> You could investigate `:focus-visible` to see if that will only show the existing focus style to keyboard and other AT users, but that does leave you at the mercy of browser vendors and doesn't always work as intended.
> 
> Edit: here's the MDN docs: https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible

Great suggestion James, we were using react aria to handle the focus and they have a keyboard only focus ring hook. @Monika-Koziol stuck with the original focus but just for keyboard

rightsaidjames:
This looks correct to me. This is the appearance when interacting with the dropdown/accordion with a mouse:
Screenshot 2024-01-03 at 14 36 00

And here's the original behaviour, but now only displayed when you tab to the dropdown/accordion with a keyboard:
Screenshot 2024-01-03 at 14 36 09
Screenshot 2024-01-03 at 14 36 17

The behaviour is consistent on Chrome, macOS Safari and Firefox, so on that basis I think we can say that the react-aria hook is working as intended.

Monika-Koziol (approved):
Beautiful 🙌

jack-skerrett-oak (approved):

PR #2134, opened 2023-12-13
Fix/required form attribute LESQ-165

k-huggs:
## Description

Music year: 2005

- Updates newsletter form

## Issue(s)

Fixes #LESQ-165

## How to test

1. Go to https://deploy-preview-2134--oak-web-application.netlify.thenational.academy/contact-us

## Screenshots

Screenshot 2023-12-13 at 16 45 04

How it should now look:

Screenshot 2023-12-13 at 16 44 33

## Checklist

- [ ] Design sign-off

### DIFF
--- a/CHANGE_LOG.md
+++ b/CHANGE_LOG.md
@@ -1,3 +1,7 @@
+## [1.311.5](https://github.com/oaknational/Oak-Web-Application/compare/v1.311.4...v1.311.5) (2023-12-15)
+
+## [1.311.4](https://github.com/oaknational/Oak-Web-Application/compare/v1.311.3...v1.311.4) (2023-12-14)
+
 ## [1.311.3](https://github.com/oaknational/Oak-Web-Application/compare/v1.311.2...v1.311.3) (2023-12-12)
 
 
--- a/CHANGE_LOG.md
+++ b/CHANGE_LOG.md
@@ -1,3 +1,10 @@
+## [1.311.3](https://github.com/oaknational/Oak-Web-Application/compare/v1.311.2...v1.311.3) (2023-12-12)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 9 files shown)


JBR90 (changes requested):
I think 'required' is a different font - I think we did something similar with the school picker

JBR90 (approved):
👍

jack-skerrett-oak (changes requested):
Though they are labelled by the correct name, it doesn't seem like the checkboxes have the required attribute set

jack-skerrett-oak (changes requested):
The email input doesn't have the text "(required)" in the label anymore!

rightsaidjames:
The label fields for the newsletter signup fields are `aria-hidden="true"` and I have no idea why this might be the case, as it means that, from a screen reader perspective, these fields have no labels. Functionally and visually it all looks correct, and I can see that Jack's comments are addressed by the latest changes.

Screenshot 2023-12-18 at 10 29 05

Screenshot 2023-12-18 at 10 26 53

rightsaidjames:
Update: it seems that the aria-hidden attribute was originally added for a good reason:
https://github.com/oaknational/Oak-Web-Application/blame/fc4c27edd7ea569d3a0b23807a150ee002b61c52/src/components/Input/Input.tsx#L149
https://github.com/oaknational/Oak-Web-Application/pull/880#pullrequestreview-1140337057

So on that basis Kyle has reverted the change and I can discuss this with @jack-skerrett-oak in the new year. My initial suspicion is that the redundant `aria-labelledby` attribute on the label was causing VoiceOver to read the label twice, so the correct fix would have been to remove this redundant attribute rather than adding `aria-hidden` to the label.

rightsaidjames (approved):

PR #2144, opened 2024-01-03
Fix/lesq 160/hamburger nav a11y [LESQ-160]

njblore:
## Description

Music year: 2000

- Wrapped the hamburger menu links section in a `nav` element
- Added `aria-expanded` true/false to the menu container when open/closed respectively
- The menu button is already marked as aria-expanded=false when the menu is closed:
- 
- @jack-skerrett-oak please verify the `nav` is at the correct level as I had some trouble figuring out where it should go!

## How to test

1. Go to https://deploy-preview-2144--oak-web-application.netlify.thenational.academy
3. You should see the element is a `nav`


src/components/Menu/Menu.tsx
```diff
@@ -133,7 +141,7 @@ const Menu: FC<MenuProps> = ({ children, menuButtonRef }) => {
                   aria-expanded={open}
                 />
               </Box>
-              <Flex
+              <NavMenuList
```
JBR90, line 144: Did you try  role="navigation" - If it works could save on creating the extra component ?
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role

JBR90 (approved):
Looks good, approved with a question

src/components/Menu/Menu.tsx
njblore: I did consider that but I think it's better / more a11y friendly to use semantic HTML where possible, and it's easy enough to do here!

jack-skerrett-oak (changes requested):
So there is some oddities happening with the button and aria-expanded states. You can see them change depending on which button I am currently viewing the state of. In your screen shot, you can see the title is "Close Menu" and the value is "aria-expanded=false", which is odd, because when it is closed the menu title is "Menu". You can see this happening in my video. When inspecting the hamburger button on the homepage, it doesn't have an aria-expanded value

https://github.com/oaknational/Oak-Web-Application/assets/102532470/10525c3f-25f0-4813-801c-7dd86f4d7194

njblore:
Thanks @jack-skerrett-oak I've made those changes - for the menu button there are actually 2 separate buttons so hopefully that isn't an issue?

jack-skerrett-oak (approved):

PR #2145, opened 2024-01-03
chore: update avo events LESQ-529

k-huggs:
## Description

Music year: {owa_music_year}

- Updates Avo search events on homepage and searchpage

## Acceptance Criteria:

- [x]  Search Attempted tracking event
- [x]  Search Journey Initiated
- [x]  Search Result Opened
- [x]  Search Results Displayed

## Issue(s)

Fixes #LESQ-529

## How to test

1. Go to https://deploy-preview-2145--oak-web-application.netlify.thenational.academy
2. Click on \_\_\_\_\_\_

## Checklist

- [ ] Design sign-off


JBR90 (commented):
Looks good - only this is the 'only' in the tests and 
 `Analytics Use Case` = null from home page search - is this expected ? 



<br class="Apple-interchange-newline">

k-huggs:
> Looks good - only this is the 'only' in the tests and `Analytics Use Case` = null from home page search - is this expected ?

Fixed the .only - I think this may be because the URL is /#teachers this was pre-exisiting behaviour should I add it to the ticket?

JBR90 (approved):

jack-skerrett-oak (changes requested):
- "Search journey initiated" doesn't seem to be (always) sent when the search box is selected. 
- "Search journey initiated" has a depth of "refine", I can't confirm if that is right or wrong but I feel it should be "interest"?
- I don't know how Ewan intends for "context" to work, but currently if you search from the homepage, the resulting "Search Results Displayed" event has a context of "search" which kind of makes sense because that is where the search results are displayed, but we then lose that those search results were displayed as a result of a search from the homepage... possibly one for Ewan.
- If you click on a unit result, then in the "Search result opened" event, it populates the "Lesson name" as whatever the unit name is. I imagine this should be null along with the "lesson slug"
- If you click a lesson result, then in the "Search result opened" event, it populates the "Unit name" with whatever the lesson name is. Kind of the inverse of the above, but we have a unit name that we should be using. "Unit slug" is fine!

Only key stage filters returned in the "Search Filter Option Selected", I assume that will be fixed on the exam board filter branch, as well as the introduction of the "Search Refined" event.

"Search result expanded" not included on this branch.

src/components/HomePageTabs/OldBetaHomeTab/OldBetaHomeTab.tsx
```diff
@@ -30,6 +30,8 @@ const OldBetaHomeTab: FC<OldBetaHomeTabProps> = (props) => {
       pageName,
       searchFilterOptionSelected: [],
       searchSource: "homepage search suggestion",
+      context: "homepage",
+      depth: "attempt",
     });
```
njblore, line 35: is this component in use anymore? I don't see it being used anywhere

k-huggs:
> * earch journey initiated" has a depth of "refine", I can't confirm if that is right or wrong but I feel it should be "interest"?

- Removed the depth field as we're going to do more work on defining it
- Search journey is initiated after entering some text, not just on selection of the search box
- Unit and Lesson slug bugs should be fixed
- Filters will be added on the examboard PR
- Search results expanded will be on Joe's PR

jack-skerrett-oak (changes requested):
Still looks like unit name is not reported when opening a lesson result
Screenshot 2024-01-04 at 16 22 46

jack-skerrett-oak (approved):
Kyle has said we do not have access to the unit name at this point, so happy to dismiss my previous review comment

PR #2155, opened 2024-01-05
Fix/hubspot errors 

njblore:
## Description

- creates a hook to submit to hubspot separate from performing downloads

## Fixes
There seems to be a bug with hubspot submissions failing and causing the downloads page to show an error message even though the download succeeded. The error with hubspot is still occurring but now it is not blocking the downloads journey, and should be clearer in bugsnag.

## How to test
I can reproduce this error consistently on firefox (mac) but not on chrome

To reproduce the error:
1. go to https://thenational.academy/teachers/programmes/citizenship-secondary-ks3-l/units/citizenship-whats-it-all-about-e038/lessons/what-is-citizenship-c4t38d/downloads?preselected=all
2. click Download
3. see the error 

To test the fix:
1. Go to https://deploy-preview-2155--oak-web-application.netlify.thenational.academy/teachers/programmes/citizenship-secondary-ks3-l/units/citizenship-whats-it-all-about-e038/lessons/what-is-citizenship-c4t38d/downloads?preselected=all
4. Click on Download
5. You should see the download succeed and page update to confirmation state


jack-skerrett-oak (approved):
Looks good. Downloads still work on Chrome and you can see the form submission in hubspot. 

In firefox, the files are downloaded and no error is shown to the user, but the form isnt submitted to hubspot (as expected).

kimon-satan (approved):
don't fully understand what the issue was here but I was able to reproduce and see that you've resolved it with the new hook

PR #2170, opened 2024-01-09
chore: update support your team text

k-huggs:
## Description

Music year: 2009

- Removes heading tags from bubble message on support team page

## Issue(s)

Fixes #LESQ-151

## How to test

1. Go to https://deploy-preview-2170--oak-web-application.netlify.thenational.academy/support-your-team
2. Tags of headings changed


src/styles/utils/typography.ts
```diff
@@ -106,6 +106,7 @@ export type TypographyProps = FontProps & {
   $wordWrap?: ResponsiveValues<"normal" | "break-word" | "initial" | "inherit">;
   $textOverflow?: ResponsiveValues<"clip" | "ellipsis">;
   $overflowWrap?: ResponsiveValues<"normal" | "break-word" | "anywhere">;
+  $fontWeight?: ResponsiveValues<FontWeight>;
```
njblore, line 109: is this not already within the `FontProps`?

JBR90: I think font weight is tied into the 'font' types. 'body-1' etc - I think this was to keep it restricted to the predefined font types. Could a font type be passed into OutlineHeading ?

njblore: Ah I see, makes sense! Yeah I agree then we should pass in a font type - if there is no predefined font type that fits we either need to make one or adjust the design to use a predefined one (I'd prefer the latter)

k-huggs: I did this because when I passed a font into the outline heading the text below would assume the position underneath the heading
Screenshot 2024-01-10 at 09 59 03

Will try an alternative solution

njblore (approved):

jack-skerrett-oak (approved):

PR #2180, opened 2024-01-11
chore(deps): msw 2 update

jimCresswell:
Update MSW to major version 2, breaks tests.

PR #2183, opened 2024-01-12
SPIKE/RFC: move from jest to vitest

mantagen:
# Vitest spike/request for comments

Reason for my wanting to spike was through frustration at having to polyfill to fix tests broken by an [MSW update](https://github.com/oaknational/Oak-Web-Application/pull/2180), only for [that polyfill](https://mswjs.io/docs/faq/#requestresponsetextencoder-is-not-defined-jest) to break a load of other tests.

Initial thoughts/notes:
- it's faster (I think roughly 40% on our codebase) <- this was local, though in CI I'm not sure it's faster 🤔 
- far less polyfilling and hacks required (`vi.hoisted()` is a nice solution to the jest mock hoisting closure hacks)
- i've gone for the more explicit 'import everything' approach, rather than enabling globals (like jest has)

## Description

Music year: 1962

- changes test framework from jest -> vitest
- exports coverage
- disable 'globals' so that all imports (describe, expect, etc) are imported from "vitest"
  - happy to change this subsequently if people aren't happy with it
- updates npm scripts
- only application code change is that the pagination util now sets the href value to be {rotuer.asPath} (i.e. the current path) for disabled prev/next buttons (previously value was set to "", but actually what was rendering was "/[blogSlug]/some-page" etc.

### Still to do

- would be good to remove some of the logs, that can be a separate PR
  - my preference would be to mock out console on a per test basis when we expect there to be logs
  - a lot of them might be coming form error-reporter, so maybe could mock that out in vitest.setup.ts
  - we can also run vitest with `--silent`, though this feels like brushing it under the carpet perhaps?
  - one of the warnings in the tests is about dynamic creation of a styled-component, which should be dealt with in the application code
- I've had to `.skip` all the SEO props tests -- ideally (for me) we'd accept this regression until next week when I can submit another PR which fixes it; reason being that this PR is very very very susceptible to conflicts. On the other hand, as it makes so many changes, it might be better to have it thoroughly tested etc.

## Checklist

- [ ] Design sign-off


kimon-satan:
I think this is well worth it. Vitest is faster and seems to be better maintained than Jest. I think it handles concurrency slightly differently so I think this would be another thing to look out for - specifically where beforeAll etc has not been used for the setup and tear down

mantagen:
Closing as it didn't get the traction required

PR #2197, opened 2024-01-17
[ENG-576] Remove MSW from OWA

jimCresswell:
We use Mock Service Worker in the Hubspot tests, but the upgrade to v2 is causing problems, this PR changes the approach to mock `fetch` calls with Jest mocks instead.

Validation: the Hubspot tests still make sense and still pass.


src/browser-lib/hubspot/forms/hubspotSubmitForm.test.ts
```diff
@@ -130,13 +154,24 @@ describe("hubspotSubmitForm", () => {
...
+      global.fetch = fakeFetch.asFetch;
+
       try {
         await hubspotSubmitForm({ hubspotFormId, payload });
       } catch (error) {
         //
       }
+
+      // Get the mock response from the fallback form.
+      const fakeFetchResponse = fakeFetch.getResult(1) as ResponsePromise;
```
jimCresswell, line 172: @mantagen maybe these values should be taken from the response matcher config, rather than the return value of the mock fetch function, what do you think?

mantagen (approved):
LGTM, nice one!

PR #2198, opened 2024-01-17
test(wrapped in act warning): set testing-library/dom override

mantagen:
## Description

Music year: 2003

- sets override of "testing-library/dom" , because "user-events" was running a lower version, which caused the warning
  - see https://github.com/testing-library/user-event/discussions/906 for more details
- removes a bunch of act() calls which are now unnecessary (and lead to warnings themselves!)

## How to test

1. run `npm t`
2. should see far fewer "wrapped in act()" warnings


njblore (approved):

jimCresswell (approved):
🚀

PR #2205, opened 2024-01-22
fix: hubspot failure on downloads page [LESQ-563]

njblore:
## Description

Music year: 2019

- Send hubspot submission before downloads on the download page
- I don't fully understand what the issue was ... 
- I've tested on Chrome, Firefox and Safari (mac) and all seems to be working while on Production I can reproduce the error 100% in Firefox on the downloads page

## Issue(s)

Fixes # 
https://www.notion.so/oaknationalacademy/Hubspot-form-is-sometimes-failing-to-send-647a3bc82c5040db9886deb6ca24f1fa?pvs=4 

## How to test

1. Got to https://www.thenational.academy/teachers/programmes/chemistry-secondary-ks4-l/units/atomic-structure-and-the-periodic-table-u68nqqc/lessons/atoms-elements-and-compounds-h3hasv/downloads?preselected=all on Firefox and download all, see the hubspot failure in the network tab and error in the console
2. Go to https://deploy-preview-2205--oak-web-application.netlify.thenational.academy/teachers/programmes/chemistry-secondary-ks4-l/units/atomic-structure-and-the-periodic-table-u68nqqc/lessons/atoms-elements-and-compounds-h3hasv/downloads?preselected=all and repeat the steps to download
4. You should see hubspot success 

## Screenshots

How it should now look:


jack-skerrett-oak (approved):
Downloads work across browsers and the event is seen in hubspot. The only outstanding question is whether we are happy for the form submission to happen before the download and other tracking (and I think we are)

mantagen (commented):
Nice one @njblore !

My only comment is whether we should switch the order so that the download happens first ... otherwise if there's an issue with Hubspot then it will affect our users. cc @jack-skerrett-oak 

E.g. this current issue would have been much worse if it meant users couldn't download content.

I wonder if the code should be more like:

```ts
await downloadStuff()

try {
  await submitToHubspot()
} catch () {
  // bugsnag already notified i believe
}

setSuccess(true)

```

That way the download happens regardless ... then we send to hubspot, but we await it so that nothing unmounts etc until browser had a chance to fetch properly?

njblore:
@mantagen yeah I have tried in a few ways to submit to hubspot after downloading succeeds, from further investigation it looks like the actual download event onClick is blocking the hubspot post request, wrapping the hubspotSubmit call in setTimout allows it through based on rough testing, but there must be a better solution...

Sidenote the hubspotSubmit method does have error handling, so it won't block the download if it fails first

PR #2210, opened 2024-01-24
fix: point dowloads gql to correct mv

kimon-satan:
point src/node-lib/curriculum-api-2023/queries/lessonDownloads/lessonDownloads.gql and src/node-lib/curriculum-api-2023/queries/lessonDownloadsCanonical/lessonDownloadsCanonical.gql to mv_downloads_3_0_2 which corrects and issue with duplicate lessons

### testing

Head to any lesson on the teacher journey and download some items


mantagen (approved):
LGTM

PR #2207, opened 2024-01-22
PUPIL-249 feat(oak-layout): add oak layout to quiz

benprotheroe:
Add quiz layout to quiz view

## How to test

1. Go to https://deploy-preview-2207--oak-web-application.netlify.thenational.academy/pupils/programmes/english-secondary-ks4-aqa/units/macbeth-and-masculinity-the-struggle-for-power/lessons/duncan-as-a-father-figure
2. Click on quizzes
3. You should see quiz layout

## Screenshots

How it used to look (delete if n/a):
{screenshots}

How it should now look:
{screenshots}

## Checklist

- [x]  Use oakLessonLayout
- [x]  MCQs display ‘select x answers’
- [x]  Question count
- [x]  Back button and header
- [x]  Feedback in bottom nav
- [x]  Different colour backgrounds based on starter or exit quiz
- [x]  mobile, tablet browser views
- [x]  final button is ‘continue lesson’ instead of ‘next question’

PR #2211, opened 2024-01-24
PUPIL-334: Serve quiz images directly from cloudinary

carlmw:
## Description

* Points oak-components at its new home under the oaknational npm scope
* Updates images under the new quiz UI to use `OakCloudinaryImage` which serves up resized/responsive images directly from Cloudinary

https://www.notion.so/oaknationalacademy/Configure-OWA-for-new-OakImage-arrangement-d32790e622db476385963421a867b4b7?pvs=4

## How to test

1. Go to https://deploy-preview-2211--oak-web-application.netlify.thenational.academy/pupils/programmes/maths-secondary-ks3/units/graphical-representations-of-data/lessons/constructing-bar-charts-by-utilising-technology
2. Take the quizzes
3. You should see that the images are snappy to load


kimon-satan (approved):

rightsaidjames (approved):
Tested in Chrome, Safari and Firefox, all looking great. I'm unable to test on iOS or Android because this branch does not have the responsive layout that is being introduced in PUPIL-249 / #2207 so some form fields are pushed off screen when the question stem is too long.

With Cache disabled and the connection throttled to 'Regular 3G' in Firefox, I'm still getting excellent loading speeds and no obvious delay beyond a brief flash.
Screenshot 2024-01-25 at 09 12 11

@carlmw Are the placeholder images still to be implemented in OWA?

carlmw:
> @carlmw Are the placeholder images still to be implemented in OWA?

😭 yes they should be appearing, but after some diving through old versions of the Next.js docs i've found that i've used a feature to implement them that is only available in Next 14 (which the oak-components Storybook is running) but OWA is running Next 13 which doesn't support it.

PR #2212, opened 2024-01-24
PUPIL-332 replace Grid and GridArea with OakGrid and OakGridArea

weronika-szalas:
## Description

An attempt to replace `Grid` and `GridArea` with `OakGrid` and `OakGridArea` across the `Oak-Web-Application` app.

### Remaining files/components:
- `PostSingleLayout` - can't replace one occurrence of `GridArea` because it takes `HEADER_HEIGHT` variable to calculate the top position and the props of this variable are not inline with allowed values in `oak-components`
- `LI` from `SharedComponents` uses `GridAreaProps` but this component will be eventually replaced by `OakLI` from `oak-components`
- `DownloadCardGroup` used `Grid` with properties that are not available in `OakGrid` such as `$gridTemplateAreas` and `$width` set to `max-content`
- `LessonOverview.view` and `develop-your-curriculum` - use `GridArea` as `OakGridArea` doesn't properly apply `display` props (ie. `$display={["none", "block"]}` is overwritten by `display: flex`)

### Comments

- Added deprecated extension and comment to `Grid` and `GridArea`
- Related to this, we still need to port over `GridList`, `GridOrderedList` and `GridAreaListItem` to `oak-components`

## Issue(s)

Fixes #PUPIL-332

## How to test

1. Go to https://deploy-preview-2212--oak-web-application.netlify.thenational.academy
2. Check through the pages
3. There should be no visual differences with live app


rightsaidjames (approved):
I've sense checked all of the page templates in Chrome, Firefox, macOS Safari and iOS Safari to ensure that all of the page components fit together without weird spacing, stack appropriately on mobile/tablet and generally look at they should. I haven't done a pixel-perfect comparison, given that we have Percy to do this for us at scale.

Assuming that @Monika-Koziol, @HelenHarlow and @messmike are happy, I'm happy!

JBR90 (approved):

PR #2218, opened 2024-01-30
fix: use correct field names in highlight query [LESQ-528]

njblore:
## Description

Music year: 2017

- update fields for highlight in the 2023 query to use correct casing
- this will need to be replicated in the search api when we switch to constructing the query there

## How to test

1. Go to https://deploy-preview-2218--oak-web-application.netlify.thenational.academy
2. Search for eg 'macbeth'
3. You should see the new results contain the keyword highlighted in bold

## Screenshots

How it should now look:

## Checklist
- [ ]  The matched search term should be in bold for any returned result


jack-skerrett-oak (approved):

JBR90 (changes requested):
Looks good, but we will need to move this change to the search api or we will see a regression when we move to versioned endpoints and construct the elastic query in the search api rather than OWA.

njblore:
closing as this will be done in the search api

PR #2223, opened 2024-02-01
Feat/connect to versioned search api - LESQ-581

JBR90:
## Description

Music year: 2013

- Removes construct elastic query (We now construct in the search api)
- Adds version to endpoint in browser config

## How to test

1. Go to https://deploy-preview-2223--oak-web-application.netlify.thenational.academy
search should work the same as before

## Acceptance Criteria:

- [x]  Release search api version to staging
- [x]  Delete construct query from OWA and just send query object
- [x]  Update staging config with search endpoint with version number (2)
- [x]  Test search is working
- [x]  Release search api version to production
- [x]  Update production config with search endpoint with version number (2)
- [x]  Test search is working


src/browser-lib/getBrowserConfig.ts
```diff
@@ -147,7 +147,7 @@ const envVars = satisfies<Record<string, EnvVar>>()({
     default: null,
   },
   searchApiUrl2023: {
-    value: process.env.NEXT_PUBLIC_SEARCH_API_URL_2023,
+    value: `${process.env.NEXT_PUBLIC_SEARCH_API_URL_2023}/v2`,
```
njblore, line 150: wondering if this should be a part of the env var instead of added in the code?

JBR90: Yeah i originally had the change in config, this was @mantagen suggestion. Although he suggested I could reach out for other opinions. Happy to move back to config if that's better.

njblore: we discussed this and think it makes sense to handle in the code, so any version cahnges can be properly tested in a PR

njblore (approved):

PR #2226, opened 2024-02-05
fix: responsive width

kimon-satan:
## Description

This update ensures that the quiz text input remains at a constant size before and after the validation icon appears.

comments from designer:

“there is some shift in the size of the text input when submitting the question on desktop, and when the icon appears when in a valid or invalid state. The text input width shouldn't increase in width ideally across states. I think the width should be increased to something like 500px as a default for desktop. The text input should span the full width for tablet and mobile.” 

### Acceptance criteria

- [x]  Text input is same width in feedback and input modes
- [x] Works on mobile and desktop states

## How to test

1. Go to https://deploy-preview-2226--oak-web-application.netlify.thenational.academy/pupils/programmes/english-primary-ks2/units/crazy-about-cats-reading/lessons/analysing-use-of-language-in-crazy-about-cats?section=starter-quiz

## Screenshots

image

image

How it should now look:
image

image


jimCresswell (approved):
🚀

Monika-Koziol (commented):
@kimon-satan, I see you implemented it like Mike asked. Should it be less wide, though? From the UX perspective, the size of the input field suggests how much to type in. This looks like the answer should be a longer word or even a sentence.

PR #2227, opened 2024-02-05
Feat/lesq 601/subject unit listing cohort [LESQ-601]

njblore:
## Description

Music year: 2010

- update mvs for `unit_listing` and `subject_listing`
- add filter to the queries to limit to non-legacy results

## How to test

I tested that there is no contamination by publishing legacy programmes locally. 

1. Go to https://deploy-preview-2227--oak-web-application.netlify.thenational.academy
2. Go to the subject listing page, then the unit listing page for legacy and non legacy
3. You should see no difference

## Checklist

- [ ]  Owa should look and behave the same as it does now with regards to the new tag on content
- [ ]  update GQL query to filter out legacy data


JBR90 (approved):

jack-skerrett-oak (approved):
New tag looks fine. No change in the content displayed on the subject listing and unit listing pages.

PR #2231, opened 2024-02-06
[CUR-237] Curriculum visualiser mobile optimisation

sonalisin:
## Description

Music year: {owa_music_year}

- List of changes
Designed according to Figma here for mobile: https://www.figma.com/file/3cz9fb0ObFDLv1c212OtJv/Final-designs---Curriculum-squad?node-id=811%3A357218&mode=dev

To do:
[] Look into fixing app header
[] Styling / spacing finishing touches
[] Adding offset to scrolling to year group to account for the fixed header

Desktop should remain the same

## How to test

1. Go to {owa_deployment_url}
2. Click on any subject phase combination in mobile
3. You should see the visual changes according to Figma

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist

- [ ] Design sign-off


HelenHarlow (commented):
Looks so good! A few things from me:

- For the YG buttons, could they be lighter (grey20 background with black text), and just the one in focus be the primary black? It’s currently not possible to tell which YG the user is currently on

- For me, when I click one of the YG buttons it shoots me down the page and then back up again. Is it possible to take the user directly to the corresponding YG?

- Is it possible to make it so the spacing above and below the YG buttons (when it is sticky) is 8px? On a desktop browser it’s fine because there is a scroll bar, but when I looked on my phone there's no spacing below

Happy to discuss any of this in more detail

rightsaidjames:
This is looking much better now, and more stable. Previously it was crashing and re-rendering on my iPhone whenever I switched between year groups or scrolled up the page.

A few small niggles:
- There is a small white line on Safari at the right hand side of each year group button, I think this bug has been fixed for other buttons:
image

- The subject phase picker modals need to have a higher z-index than the mobile-only components:
image

One for @HelenHarlow - the 'Done' button in the thread modal is overlapped by the Gleap button, can we move it? 
image

src/components/CurriculumComponents/UnitsTab/UnitsTab.test.tsx
```diff
@@ -24,6 +24,13 @@ jest.mock("@/context/Analytics/useAnalytics", () => ({
 describe("components/pages/CurriculumInfo/tabs/UnitsTab", () => {
   beforeEach(() => {
     jest.clearAllMocks();
+    const mockIntersectionObserver = jest.fn();
```
njblore, line 27: I'd move this out of the before each so it's not getting redeclared for each test

src/components/CurriculumComponents/UnitsTabMobile/UnitsTabMobile.tsx
```diff
@@ -0,0 +1,256 @@
...
+import Button from "@/components/SharedComponents/Button/Button";
+import ButtonGroup from "@/components/SharedComponents/ButtonGroup";
+import Radio from "@/components/SharedComponents/RadioButtons/Radio";
+import RadioGroup from "@/components/SharedComponents/RadioButtons/RadioGroup";
+
+// Types and interfaces
+
+type UnitsTabMobileProps = {
+  selectedThread: Thread | null;
+  handleSelectThread: (thread: string) => void;
```
njblore, line 16: I'd rename this param `threadName` / `threadSlug` or whatever it's really expecting to avoid confusion

njblore (approved):
looks good to me! just left a few small comments

rightsaidjames:
Some extra feedback for when you're back:

* Updating the active 'chip' when you scroll down the page works well, feels nice and smooth! :tada: 
    * However for some reason it doesn't work when you get to Year 10 on Secondary Science:
    * ... unless you select a different child subject, and then the chip works correctly for both year 10 and year 11
    * Despite this, you can still click on the buttons to scroll to those years, they just don't become active when the child subject for year 10 is Combined Science.
    * There's a similar problem with Primary English (due to the domains?)
* On both Android and iOS I can't scroll the list of threads inside the modal, I can only see the top ones. This is particularly problematic for Secondary Science, Secondary Music and Primary Maths, which have a lot of threads to choose from.
*  The mobile experience otherwise works well on iOS and Android, and the desktop experience is the same as it was before.

**Screenshots**:
Screenshot 2024-03-01 at 11 42 11
Screenshot 2024-03-01 at 11 43 29
Screenshot 2024-03-01 at 11 43 38
IMG_1068
Screenshot 2024-03-01 at 12 01 25
Screenshot 2024-03-01 at 12 15 21

HelenHarlow:
> One for @HelenHarlow - the 'Done' button in the thread modal is overlapped by the Gleap button, can we move it? image

Sorry, I just saw this! Yes if we can move it to the left that would be ideal. Thanks

src/components/CurriculumComponents/UnitsTab/UnitsTab.tsx
```diff
@@ -531,215 +515,24 @@ const UnitsTab: FC<UnitsTabProps> = ({ data, examboardSlug }) => {
...
+            yearData={yearData}
+            handleSelectDomain={handleSelectDomain}
+            handleSelectSubject={handleSelectSubject}
+            handleSelectTier={handleSelectTier}
+            duplicateUnitSlugs={duplicateUnitSlugs}
+            mobileHeaderScrollOffset={mobileHeaderScrollOffset}
+            setUnitData={setUnitData}
+            selectedThread={selectedThread}
+            setVisibleMobileYearRefID={setVisibleMobileYearRefID}
+          />
```
BTitterington, line 532: I think this was a good move. Maybe a separate PR as you mentioned in slack :). I can see that a few of my comments were really reviewing the structure of this and you are just moving it (as it was). We can create a separate PR for the smaller components I mentioned earlier `Year` & `Unit`. Up to you if you want to move modal in this PR also.

src/components/CurriculumComponents/UnitsTabMobile/UnitsTabMobile.tsx
```diff
@@ -0,0 +1,256 @@
...
+  highlightedUnitCount,
+  trackSelectYear,
+  yearOptions,
+  visibleMobileYearRefID,
+}) => {
+  // Initialize constants
+  const [mobileThreadModalOpen, setMobileThreadModalOpen] =
+    useState<boolean>(false);
+  const [mobileYearSelection, setMobileYearSelection] = useState<string | null>(
+    "7",
```
BTitterington, line 43: Shouldn't need the "7", if it does then it probably should be fixed in a different location or with a different non hardcoded value.

src/components/CurriculumComponents/UnitsTabMobile/UnitsTabMobile.tsx
```diff
@@ -0,0 +1,256 @@
...
+                      }
+                      isCurrent={yearOption === mobileYearSelection}
+                      key={yearOption}
+                      label={`Year ${yearOption}`}
+                      onClick={() => {
+                        setMobileYearSelection(yearOption);
+                        // Scroll into view used also in Lesson Overview - prevents rerender
+                        document
+                          .getElementById(`year-${yearOption}`)
+                          ?.scrollIntoView({ behavior: "smooth" });
```
BTitterington, line 241: I have to admit the "click to select year" with scroll makes this ticket tricky. Really you shouldn't need to call setMobileYearSelection and leave the state update to the intersection observer, but that means the intersection observer needs a rootMargin to deal with the top spacing and also the years will cycle through all selected options as it navigates (with a delay), which won't look great. 

I think we should chat about this. A single reference could be created and passed into both the CurriculumVisualiser & UnitsTabMobile. CurriculumVisualiser could populate the ref with the item elements and then all of the intersection observer logic could live in the UnitsTabMobile component (which is probably better anyway). It would be easier to deal with this part then, although I still would need to think on it more.

BTitterington (changes requested):
Sorry for the volume of comments here. It's a complex ticket with a lot of code to review.

src/components/CurriculumComponents/CurriculumVisualiser/CurriculumVisualiser.tsx
```diff
@@ -0,0 +1,405 @@
...
+                      setUnitVariantID={setUnitVariantID}
+                      unitData={unitData}
+                      displayModal={displayModal}
+                      setUnitOptionsAvailable={setUnitOptionsAvailable}
+                      unitOptionsAvailable={unitOptionsAvailable}
+                      isHighlighted={
+                        unitData ? isHighlightedUnit(unitData) : false
+                      }
+                    />
+                  </UnitsTabSidebar>
```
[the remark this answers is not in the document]
sonalisin: I agree - the data being passed through is a little weird and doesn't seem to show consistent behaviour if I move it. Will create a ticket to investigate.

src/components/CurriculumComponents/UnitsTab/UnitsTab.tsx
```diff
@@ -100,17 +77,15 @@ const UnitsTab: FC<UnitsTabProps> = ({ data, examboardSlug }) => {
...
   const [yearSelection, setYearSelection] = useState<YearSelection>({});
   const [selectedThread, setSelectedThread] = useState<Thread | null>(null);
   const [selectedYear, setSelectedYear] = useState<string | null>(null);
+  const [mobileHeaderScrollOffset, setMobileHeaderScrollOffset] =
+    useState<number>(0);
+  const [visibleMobileYearRefID, setVisibleMobileYearRefID] = useState<
+    string | null
+  >(null);
 
   // Put data formatting code in useEffect to avoid unnecessary re-renders
```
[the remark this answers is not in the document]
sonalisin: Creating a ticket!

src/components/CurriculumComponents/CurriculumVisualiser/CurriculumVisualiser.tsx
```diff
@@ -0,0 +1,405 @@
...
+    <OakGridArea $colSpan={[12, 9]} data-testid="curriculum-visualiser">
+      {yearData &&
+        Object.keys(yearData)
+          .filter((year) => !selectedYear || selectedYear === year)
+          .map((year, index) => {
+            const { units, childSubjects, domains, tiers } = yearData[
+              year
+            ] as YearData[string];
+            return (
+              <Box
```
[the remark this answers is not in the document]
sonalisin: Creating a ticket for refactor

HelenHarlow (approved):
Looks good!

BTitterington (approved):
Looks good to me. Also quickly tested behaviour on desktop & android phone and worked well.

rightsaidjames (approved):
Looks great. Tested thoroughly on iOS and Android, and I can confirm that it behaves smoothly and that all of the previous bugs are gone. Amazing work!

PR #2233, opened 2024-02-07
PUPIL-420 replace Flex occurrences in OWA with OakFlex from oak-components

weronika-szalas:
## Description

An attempt to replace `Flex` occurrences in OWA with `OakFlex` from `oak-components`.

I have replaced all straightforward occurrences of `Flex`. There are still many occurrences left in the project that need a closer look/individually checking components as they contain properties that don't exist in `OakFlex`. Those need either achieving the same result with different properties or adding relevant props to `OakFlex`. To be decided on per component basis.

One thing I've noticed that `OakFlex` accepts `$display` property but always overwrites them to `display: flex` anyway so using `<OakFlex $display=["none", "flex"]/>` will always result in `display: flex. This needs to be fixed in `oak-components` before occurrences of `Flex` in OWA that use this prop can be replaced.

I have also added `@depracated - use OakFlex from oak-components instead` comment in `Flex` and changed the name to `Flex.deprecated` to encourage using `OakFlex` going forward.

I have added above findings to this document:
https://www.notion.so/oaknationalacademy/Replacing-SharedComponents-in-OWA-app-with-oak-components-de39c55b33e947bbbd30888547f5b10e

## Issue(s)

Fixes #PUPIL-420

## How to test

1. Go to https://deploy-preview-2233--oak-web-application.netlify.thenational.academy
2. The PR affects most of the changes. Check at lest the main journeys of the app.
3. There should be no visual changes.


kimon-satan (approved):
I made a ticket for OakFlex and added it to the backlog

PR #2240, opened 2024-02-09
fix: schema for search highlight - LESQ-528

JBR90:
## Description

- Typo in the search highlight schema was removing the lesson description highlight field.

- List of changes

## Issue(s)

Fixes #

## How to test

1. Go to https://deploy-preview-2240--oak-web-application.netlify.thenational.academy
Search for object - you should the it highlighted in the search results

There are still some issues - `objects` does not match, but `object` matches 'object' and 'objects', we should look into tweaking the analyser for highlighting when we work more on search.


njblore (approved):

jack-skerrett-oak (approved):
We have highlighting again!

PR #2242, opened 2024-02-12
feat: switch to curriculum_units_0_1

lioneljingles:
## Description

Music year: 2018

- List of changes
1. Update API request for curriculum visualiser to curriculum_units_0_1


kimon-satan (changes requested):
I think your forgot to run
```npm run gql-codegen:curriculum-2023```
without that I think the old mv continues to be used

kimon-satan (approved):

PR #2247, opened 2024-02-14
Feat/lesq 637/legacy remove duplicate description [LESQ-637]

njblore:
## Description

Music year: 1976

- remove the pupil lesson outcome from the lesson header when it is included in key learning points
- only applies to EYFS currently, but will apply to legacy when migrated

## How to test
1. Go to https://deploy-preview-2247--oak-web-application.netlify.thenational.academy/teachers/programmes/citizenship-secondary-ks4-l/units/how-well-does-the-media-hold-those-in-power-to-account-63c4/lessons/what-is-the-role-of-the-media-in-a-democracy-cmw32c
3. You should see no pupil lesosn outcome in the header (same as prod)
4. Go to https://deploy-preview-2247--oak-web-application.netlify.thenational.academy/teachers/programmes/combined-science-secondary-ks4-higher-aqa/units/measuring-waves/lessons/transverse-waves
5. You should see the pupil lesson outcome in the header (same as prod)

## Screenshots

How it should now look:

### DIFF
--- a/src/components/TeacherViews/LessonOverview/LessonOverview.view.tsx
+++ b/src/components/TeacherViews/LessonOverview/LessonOverview.view.tsx
@@ -38,12 +38,24 @@ import LessonOverviewAnchorLinks from "@/components/TeacherComponents/LessonOver
 import { MathJaxProvider } from "@/browser-lib/mathjax/MathJaxProvider";
 import { GridArea } from "@/components/SharedComponents/Grid.deprecated/GridArea.deprecated.stories";
 import { LEGACY_COHORT, NEW_COHORT } from "@/config/cohort";
+import { keyLearningPoint } from "@/node-lib/curriculum-api-2023/shared.schema";
 
 export type LessonOverviewProps = {
   lesson:
     | LessonOverviewCanonical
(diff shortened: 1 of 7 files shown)


src/components/TeacherViews/LessonOverview/LessonOverview.view.tsx
```diff
@@ -46,6 +47,17 @@ export type LessonOverviewProps = {
     | SpecialistLessonOverview;
 };
 
+// helper function to dedupe key learning points from the header in legacy lessons
+export const isPupilLessonOutcomeInKeyLearningPoints = (
```
lioneljingles, line 51: If this helper were changed to return `pupilLessonOutcome` if it is found, or `undefined` if it is not, it would make the view a bit simpler

lioneljingles (approved):

jack-skerrett-oak (approved):
Can't see any changes in current legacy/new content. It is worth nothing that looking more closely, some current legacy data does have different fields for pupil lesson outcome and key learning points, so it will be interesting to see what that looks like with the legacy data (not important for now!)

PR #2248, opened 2024-02-14
Fix/lesq 628/lesson share [LESQ-628]

njblore:
## Description

Music year: 2014

- make shareable resource metadata nullable as it only exists when the resource exists 

## Issue(s)

Fixes # 
share page not loading for EYFS lessons

can't be tested until it is deployed to the migration instance

### DIFF
--- a/src/components/TeacherComponents/LessonShareCardGroup/LessonShareCardGroup.test.tsx
+++ b/src/components/TeacherComponents/LessonShareCardGroup/LessonShareCardGroup.test.tsx
@@ -1,0 +1,104 @@
+import { screen } from "@testing-library/react";
+import { useForm } from "react-hook-form";
+import userEvent from "@testing-library/user-event";
+
+import { ResourceFormProps } from "../types/downloadAndShare.types";
+
+import LessonShareCardGroup from "./LessonShareCardGroup";
+
+import { LessonShareData } from "@/node-lib/curriculum-api";
+import renderWithTheme from "@/__tests__/__helpers__/renderWithTheme";
+
+const ComponentWrapper = (props: {
+  shareableResources: LessonShareData["shareableResources"];
+  shareLink: string;
+}) => {
+  const { control, trigger } = useForm<ResourceFormProps>();
+
+  return (
+    <LessonShareCardGroup
+      {...props}
+      control={control}
+      triggerForm={trigger}
+      shareableResources={props.shareableResources}
+      shareLink={props.shareLink}
+    />
+  );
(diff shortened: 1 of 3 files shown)


src/components/TeacherComponents/LessonShareCardGroup/LessonShareCardGroup.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+  );
+};
+
+describe("lesson share card group", () => {
+  it("should render", () => {
+    renderWithTheme(
+      <ComponentWrapper shareableResources={[]} shareLink="www.fake.com" />,
+    );
+
+    const previewButton = screen.getByText("Preview as a pupil");
```
lioneljingles, line 35: Recommend using `data-testid` and `getByTestId` instead of text

src/components/TeacherComponents/LessonShareCardGroup/LessonShareCardGroup.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+    expect(previewButton).toBeInTheDocument();
+  });
+  it("should render with resources", () => {
+    const shareableResources = [
+      {
+        exists: true,
+        type: "video" as const,
+        metadata: "5min",
```
lioneljingles, line 43: Out of scope, but `metadata` should really be an object, not a string

src/components/TeacherComponents/LessonShareCardGroup/LessonShareCardGroup.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+      },
+    ];
+    renderWithTheme(
+      <ComponentWrapper
+        shareableResources={shareableResources}
+        shareLink="www.fake.com"
+      />,
+    );
+
+    const resourceCardLabel = screen.getByText("Video");
```
lioneljingles, line 54: Using `data-testid` would be clearer

lioneljingles (approved):

src/components/TeacherComponents/LessonShareCardGroup/LessonShareCardGroup.test.tsx
njblore: oh interesting, how come? I try and avoid using data-testids unless totally necessary because they're not 'testing as a user would', and you end up adding otherwise unnecessary props to components that often don't get cleaned up when tests change etc!

PR #2257, opened 2024-02-15
fix: remove margin from subject card [LESQ-625]

njblore:
## Description

Music year: 1950

- ~remove padding on subject listing cards so full PSED title fits~
- decrease heading size on subject cards to 20px

## How to test

1. Go to https://deploy-preview-2257--oak-web-application.netlify.thenational.academy
2. Should look the same except for reduced font size on subject cards

## Screenshots

How it should now look:


Monika-Koziol (changes requested):
I'd say the better way to fix it would be to drop the heading size to 20px, which would give titles more breathing space. And it would work for all of the subjects. 

Also not part of this ticket but the subtext under areas of learning has too many words per line. Can we limit to half a page on desktop only?

njblore:
thanks @Monika-Koziol I've made that change to the font size, and made a ticket for the subheading width

Monika-Koziol (approved):

PR #2269, opened 2024-02-20
PUPIL-348: Serve quiz images from new Cloudinary distribution

carlmw:
## Description

Music year: [1971](https://www.youtube.com/watch?v=MV0F_XiR48Q)

With the new pupil pages we made a change to serve quiz images directly from Cloudinary instead of passing them through Next's image processing proxy. This was done to improve image loading which has been greatly improved. Unfortunately this had the unintended side-effect of moving requests for those images off `thenational.academy` domain. This increased the chances of a request being blocked by a school network policy and possibly incurring additional tariffs for some users.

This change moves those requests back under a `thenational.academy` origin mitigating those issues.

## How to test

1. Go to https://deploy-preview-2269--oak-web-application.netlify.thenational.academy/pupils/programmes/english-primary-ks2/units/developing-reading-preferences-in-year-5/lessons/developing-reading-preferences-in-year-5-through-appreciation-of-characters
2. Take the quiz
3. Inspect each image and you should see that it is served from a Cloudinary distribution under `thenational.academy` and it should load no differently than before.
4. Go to https://deploy-preview-2269--oak-web-application.netlify.thenational.academy/pupils/programmes/science-secondary-ks3/units/moving-by-force/lessons/reading-distance-time-graphs
5. Take the starter quiz
6. Inspect the images in the answers and they should also be served from a Cloudinary distribution under `thenational.academy`

Screenshot 2024-02-20 at 12 50 06
Screenshot 2024-02-20 at 12 51 04

## Checklist

- [x] Added new env vars to development and production configs.


src/browser-lib/cloudinary/extractCloudinaryIdFromURL.ts
```diff
@@ -0,0 +1,28 @@
+import OakError from "@/errors/OakError";
+
+/**
+ * Extracts the cloudinary id (public id and version if present)
+ * from a URL and returns it.
+ *
+ * `CldImage` and `OakCloudinaryImage` can automatically extract the Cloudinary id from a URL
+ * if it is under the target distribution origin. However currently the curriculum data has the full
+ * URL baked in which points to an origin under `cloudinary.com`. In order to construct a new URL under the
+ * new distribution inside `thenational.academy` we need to extract the Cloudinary id from the original URL
```
carlmw, line 10: ☝🏼 not sure where this work belongs/when. Right now all other images inside OWA appear to be going through a vanilla Next `Image` so they are already being served from `thenational.academy` by virtue of them being proxied through Next.

jimCresswell: do you think it's okay to do the transformation in the image component? I don't like client-side data munging, but if the only ID source is the original URL I feel like it's that, or we add a field to the database, or we do it in the data layer (which would add complexity)

jimCresswell (approved):
First, excellent music choice 🤘 

This looks really good! I'm not sure what the right answer is with regard to constructing the URL from the DB data, I'm happy to chat it through, I'm also happy for the squad to make a decision

src/browser-lib/cloudinary/extractCloudinaryIdFromURL.ts
carlmw: 🤔 I think i'm actually coming down on doing it in the data layer probably as a Zod transform (at least for now until we can decide what to do with the data). It would make this change more surgical and hide this detail from the client entirely.

src/components/PupilComponents/QuizMCQMultiAnswer/QuizMCQMultiAnswer.tsx
```diff
@@ -71,17 +71,18 @@ export const QuizMCQMultiAnswer = (props: QuizMCQMultiAnswerProps) => {
...
-              height={answerImageData.height}
-              $minWidth={"all-spacing-19"}
-              placeholder="oak"
-              sizes={getSizes(["100vw", 1200])}
-            />
-          ) : undefined;
+          const answerImage =
+            answerImageData && answerImageData.public_id ? (
+              <OakCloudinaryImage
+                cloudinaryId={answerImageData.public_id}
```
carlmw, line 77: Turns out the cloudinary id was there the whole time 🤷🏼‍♂️

kimon-satan (approved):

rightsaidjames (approved):
Looks great. I tested this by inspecting quiz images to confirm they were being served from the correct domain, and also confirmed that if the `res.cloudinary.com` domain was blocked in DevTools then the quiz images (but not the SVG icons etc., as discussed) would still load.

Screenshot 2024-02-22 at 16 56 44

Screenshot 2024-02-22 at 16 57 14

Screenshot 2024-02-22 at 16 53 26

PR #2292, opened 2024-03-01
Fix/hide copyright slide deck

k-huggs:
## Description

Music year: {owa_music_year}

- Adds additional check to only display slide decks if they're downloadable
- 

## Issue(s)

Fixes #LESQ-676

## How to test

1. Go to {owa_deployment_url}
2. Click on \_\_\_\_\_\_

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist

- [ ] Design sign-off

### DIFF
--- a/src/components/TeacherComponents/helpers/lessonHelpers/lesson.helpers.ts
+++ b/src/components/TeacherComponents/helpers/lessonHelpers/lesson.helpers.ts
@@ -214,6 +214,7 @@ type GetPageLinksForLessonProps = Pick<
   | "starterQuiz"
   | "exitQuiz"
   | "hasCopyrightMaterial"
+  | "hasDownloadableResources"
 >;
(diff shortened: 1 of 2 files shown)


JBR90 (approved):

jack-skerrett-oak (commented):
Just looking at the code, because this is just changes on the front end, the lesson listing page won't know to exclude the slide deck right? So there will still be a graphic stating the lesson has a slide deck? Not the end of the world, if that can be "logic-ed" away without MV changes I think we should do it

k-huggs:
> Just looking at the code, because this is just changes on the front end, the lesson listing page won't know to exclude the slide deck right? So there will still be a graphic stating the lesson has a slide deck? Not the end of the world, if that can be "logic-ed" away without MV changes I think we should do it

I think for this page it will require an MV change, looking into it now 🕵️

PR #2295, opened 2024-03-04
[ENG-675] fix(policy pages): make internal links work

jimCresswell:
Internal links on policy pages weren't working, now they should.

To test, open the privacy policy page, and check that the first paragraph has a working, underlined link to the child-friendly version of the privacy policy.


njblore (approved):
this also turns `www.thenational.academy` into a link on the accessibility statement page, there may be other similar links in the legal pages as well - presumably that's what we want!

PR #2336, opened 2024-03-27
Feat/cur 330/add subject tags for science years 7 8 9

BTitterington:
## Description

CUR-330 - https://www.notion.so/oaknationalacademy/Add-subject-tags-for-Science-years-7-8-9-e65d572090ad43e78c5e0a76bf2fae06

This feature adds science subject tabs for KS3 (Years 7, 8 & 9) to allow users to filter and find relevant units.

They have been added in CAT as tags on the units with the category type "Discipline"

An MV was created to bring in the tag information for each unit, and this ticket uses those tags to add the filters/tabs when a unit has those attached.

## How to test

1. Go to https://deploy-preview-2336--oak-web-application.netlify.thenational.academy/teachers/curriculum/science-secondary-aqa/units
2. You should see these tabs on year 7 - All, Biology, Chemistry & Physics
3. Selecting these should filter the units as expected (check expected results in CAT)

## Screenshots

**How it used to look:**

image
 
**How it should now look:**

image

## Checklist

- [ ] Design sign-off


src/components/CurriculumComponents/UnitsTab/UnitsTab.test.tsx
```diff
@@ -467,6 +475,62 @@ describe("components/pages/CurriculumInfo/tabs/UnitsTab", () => {
     });
   });
 
+  test("user can filter units by discipline", async () => {
```
BTitterington, line 478: I would rather we had smaller components so that the year section could be tested in isolation. But for the moment I've selected the year within this test to reduce the noise.

src/node-lib/curriculum-api-2023/fixtures/curriculumUnits.fixture.ts
```diff
@@ -359,7 +375,7 @@ const curriculumUnitsTabFixture = (
         connection_future_unit_title: "Composition of numbers 6 to 10",
         connection_prior_unit_title: "Composition of numbers",
         domain: null,
-        domain_id: null,
+        domain_id: 17,
```
BTitterington, line 378: This was intentional, as I had a bug that even though the feature was broken, all unit tests were still passing. I narrowed it down to needing at least one unit that had a domain_id. In summary - with this domain_id I believe it covers more edge cases.

src/components/CurriculumComponents/UnitsTab/UnitsTab.tsx
```diff
@@ -238,11 +258,21 @@ const UnitsTab: FC<UnitsTabProps> = ({ data, examboardSlug }) => {
...
+      if (filters.disciplines.length >= 2) {
+        filters.disciplines.unshift(allDisciplineTag);
+      }
+
       initialYearSelection[year] = {
         subject:
           filters.childSubjects.find(
             (s) => s.subject_slug === "combined-science",
           ) ?? null,
+        discipline: allDisciplineTag,
```
BTitterington, line 275: Took a slightly different approach to the existing pattern. The selected discipline for the year will initially be 'all' (-1), even if there are no disciplines. In the isVisibleUnit function (which looks after the filtering of visible units) will return true for disciplines if the value is 'all' (-1).

BTitterington:
I accidentally used a normal commit instead of a merge commit on this branch. I think it will be easier to create a new branch with the changes, rather that work through the git changes required or re-write the history.

PR #2346, opened 2024-04-03
Fix preselected downloads that don't exist

JBR90:
## Description

Music year: {owa_music_year}

## Acceptance Criteria:

- [ ]  No download button should show for resources that don’t exist for download

`teachers/programmes/history-primary-ks2-l/units/20th-century-conflict-world-war-ii-ce4e/lessons/was-world-war-two-inevitable-6gv34d#worksheet
`
I've not been able to meet this criteria, we determine if a worksheet is available for download on the from end on lesson overview by looking at the lesson.worksheetUrl. 

```
 {
      label: "Worksheet",
      anchorId: "worksheet",
      condition: (lesson) => Boolean(lesson.worksheetUrl),
    },
```

This lesson has a worksheetUrl and hasDownloadableResources is true , so we download the download button.

On the data is has a worksheet asset object - this is what we use to determine if it has a download. (which is does)

So the database thinks there is a worksheet available to download.

However on the download api 

```
{
"error": {
"message": "Selected resources not found: worksheet-pdf, worksheet-pptx"
}
}
```

- [ ]  The download button should be disabled when 0 resources are selected (this is an edge case as it usually would be, but the preselected resource does not exist)

- List of changes

## Issue(s)

Fixes #

## How to test

1. Go to {owa_deployment_url}
2. Click on \_\_\_\_\_\_

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist

- [ ] Design sign-off


src/components/TeacherComponents/hooks/downloadAndShareHooks/useResourceFormState.test.tsx
```diff
@@ -154,11 +152,8 @@ describe("useResourceFormState", () => {
           query: { preselected: "worksheet" },
         });
         const { result } = renderHook(() => useResourceFormState(shareProps));
-
-        expect(result.current.selectedResources).toEqual([
-          "worksheet-pdf",
-          "worksheet-pptx",
-        ]);
+        console.log(result);
```
njblore, line 155: 👁️

njblore (approved):

jack-skerrett-oak (approved):

PR #2364, opened 2024-04-17
Fix/pupil 613/404 reloading

kimon-satan:
## Description
Currently we overwrite the url to add a path for section when it's not provided. When the lesson isn't found we go to the 404 . However that causes the overwrite action to happen again and we land back at the original page and rinse and repeat

Music year: 1977

- changed rewrites to redirects
- these are permanent redirects

## How to test

1. Go to https://deploy-preview-2364--oak-web-application.netlify.thenational.academy/pupils/lessons/transverse-waves
See that the page loads as expected and /overview is appended to the url
2. Go to https://deploy-preview-2364--oak-web-application.netlify.thenational.academy/pupils/lessons/transverse-waves-bad
See that the 404 loads but without perpetual reloading

## ACS

[x] page no longer perpetually reloads

### DIFF
--- a/CHANGE_LOG.md
+++ b/CHANGE_LOG.md
@@ -1,3 +1,42 @@
+# [1.407.0](https://github.com/oaknational/Oak-Web-Application/compare/v1.406.0...v1.407.0) (2024-04-16)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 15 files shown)


jimCresswell (approved):
Approving, pending discussed perm/temp clarification.

PR #2369, opened 2024-04-18
[CUR-456] fix: added aria-live="polite" to "units highlighted" text

orangemug:
## Description

Music year: 1996

- Added `aria-live="polite"` to `"units highlighted"` text

## Issue(s)
Fixes `CUR-456`

## How to test

When a user selects a unit, the number of highlighted units is displayed in text, and when a user filters through the years the results update, but this information isn’t conveyed automatically to screen reader users.

1. Go to `/teachers/curriculum`
2. Select a subject and school phase
3. Filter by the thread (highlight a thread)
   - The number of units highlighted is announced to screen reader
4. Select a year group
   -  The number of units highlighted should be re-announced

## Checklist

- [ ] Design sign-off


BTitterington:
@orangemug - Code looks good/safe, but I couldn't get it to read aloud the number of units highlighted (It just excluded what was in the span). I'll let James check as he has more experience here. Feel free to merge if he approves.

rightsaidjames:
The current solution works well on desktop - the number of units highlighted is announced when switching between threads, years and domains/disciplines/child subjects.

However on mobile it seems that no announcement is made when switching threads. I think this because the live region (in the sticky header) is not visible at the point at which the threads are switched. Might it make sense to add an additional live region to the thread selector inside the modal?

image

When switching between child subjects/domains/disciplines on mobile the updated number of units highlighted is announced, so I think we'll need the live region in both places. Year isn't a factor because on mobile the year buttons are shortcut links not filters.

rightsaidjames (approved):
Looks good now. The number of units highlighted is now announced after every action that affects the number of highlighted units (assuming there is a change) - switching threads, filtering by year (desktop and tablet only), filtering by domain/discipline/child subject. This applies on both mobile and desktop, because the `aria-live="polite"` attribute has been added to the thread option labels in the desktop sidebar **and** inside the mobile-only modal, and the `X units highlighted` text in the mobile sticky header also has this property.

Tested in VoiceOver on macOS and iOS

PR #2380, opened 2024-04-22
Feat/lesq 746/a11y filters [LESQ-746]

njblore:
## Description

- wrap filter groups in `fieldset` element and put the headings into `legend` elements
- add the `region` role to search results section and add `aria-live` tag
- move filters so they appear before results in tab order
- store last filter inputted by keyboard and autofocus on page reload - ended up using localStorage for this, I don't think this requires cookie consent as we're not storing anything to do with the user but would be good to get confirmation from someone who knows 100% 

## How to test

1. Go to https://deploy-preview-2380--oak-web-application.netlify.thenational.academy/teachers/search?term=macbeth
2. use keyboard to apply filters
3. You should see autofocus on page reload
4. clicking with a mouse should remove autofocus
5. clicking on anything else (search bar, active filters, results) should remove autofocus even with keyboard
6. check html for correct elements and attributes

## Checklist
- []  When you apply a filter using keyboard controls, the focus should remain on the filter you were on after the results have refreshed
- []  Semantic HTML should be used for the filters
    - []  We need to implement fieldset, legends and set the inputs to have type=”checkbox”
    
    ```jsx
    <fieldset>
    <legend>Exam boards</legend>
    ...
    </fieldset>
    ```
    
- []  Filters should be moved above the search results in tab order
- []  “aria-live=polite” should be used to announce that search results have been updated
    - This is the one I don’t fully understand how it should work [[Returned search results are not communicated as a status message.](https://www.notion.so/Returned-search-results-are-not-communicated-as-a-status-message-8efe8dac16eb47f9ba5494d49748b609?pvs=21)](https://www.notion.so/Returned-search-results-are-not-communicated-as-a-status-message-8efe8dac16eb47f9ba5494d49748b609?pvs=21)


jack-skerrett-oak (changes requested):
It looks good, however when using Mac voice over to navigate around, after searching for something or updating a field, the aria-live=polite kicks in and starts reading you every search result which isn't what you would necessarily want. James' suggestion is to add in an invisible div, or something to that effect, that states "Showing x results".. would that be possible?

src/components/TeacherComponents/SearchFilters/SearchFilters.tsx
```diff
@@ -1,87 +1,118 @@
...
 import { UseSearchFiltersReturnType } from "@/context/Search/search.types";
 import { FilterTypeValueType } from "@/browser-lib/avo/Avo";
 
 type SearchFiltersProps = {
   searchRefined: (filterType: FilterTypeValueType, filterValue: string) => void;
+  onFilterClick: (filter: string | undefined) => void;
+  lastFocussedFilter: string | null;
 } & UseSearchFiltersReturnType;
 
+const StyledFieldset = styled.fieldset`
```
JBR90, line 15: Could this be reusable and live in components ?

JBR90 (approved):

src/components/TeacherComponents/SearchFilters/SearchFilters.tsx
njblore: yeah for sure if we need it again we should do that, for now I don't think it's worth the effort though

jack-skerrett-oak (changes requested):
- We talked yesterday about removing the current solution of remembering which filter you had previously selected.
- With "Showing 62 results" being visibility:hidden, it cannot be seen by screen readers. We could either have that text display (might want to talk to Monika) or find some other way of visually hiding it that isn't using "visibility:hidden". Happy to do a quick guide of how to test with a screen reader!

jack-skerrett-oak (approved):
I'll resolve the relevant a11y issues in the audit table, but make a note that we need to revisit the tab focus thing in some way.

PR #2386, opened 2024-04-24
Feat/lesq 747/skip filters [LESQ-747]

njblore:
## Description

Music year: 2013

- Add button to skip filters to main content
- update search page and tests to use theme provider from oak components

## How to test

1. Go to https://deploy-preview-2386--oak-web-application.netlify.thenational.academy/search?term=waves
2.  Tab to skip filters button and click enter
3. You should be moved into the results, and tabbing will take you through the results

## Screenshots

How it should now look:
Screenshot 2024-04-24 at 14 43 09


njblore:
@jack-skerrett-oak ok just to make sure i'm understanding properly, the button should be invisible but appear when someone tabs to it? and there is just an empty space until it appears? and when they move away does it disappear again?

jack-skerrett-oak:
> @jack-skerrett-oak ok just to make sure i'm understanding properly, the button should be invisible but appear when someone tabs to it? and there is just an empty space until it appears? and when they move away does it disappear again?

@njblore to be honest, I thought this had figma designs/the spec was on the ticket. The solution I was suggesting is like "skip nav" elements on other sites. So the button should just appear as you tab through and not cause content to shift around, and then disappear when you tab away. Looking at the figma designs though they do suggest the content shifts.

jack-skerrett-oak (approved):

PR #2393, opened 2024-04-25
feat: added `<Autocomplete/>` component

orangemug:
## Description

Music year: 1974

- adds `<Autocomplete />` in a staging area ready to be added to [oak-components](https://github.com/oaknational/oak-components) once mature.
- adds `src/components/CurriculumComponents/OakComponentsKitchen/` as a staging area for new oak-components

Usage is now as simple as

```jsx
<Autocomplete
  inputProps={{
    label: "Pick an animal",
    id: "animals",
  }}
  onChange={setState}
  value={state}
>
  {["Cat", "Cave Bear", "Caterpillar"].map((animal) => {
    return <AutocompleteItem key={animal}>{animal}</AutocompleteItem>;
  })}
</Autocomplete>
```

## Issue(s)

Related to `CUR-428`

## How to test

1. Go to storybook `/?path=/story/components-curriculumcomponents-oakcomponentskitchen-autocomplete--autocomplete`
2. Test works as expected

## Screenshots

Screenshot 2024-04-29 at 10 45 46

## Checklist

- [ ] Design sign-off


kimon-satan (commented):
This is a cool idea but it might be confusing as these are OWA components and not OakComponents. The rule is that only components exported from https://github.com/oaknational/oak-components can be prefixed with Oak. OWA components can either be prefixed with OWA or not prefixed at all.

The aim is for this storybook to reduce in size over time with the bulk of OWA components being deprecated.

Generally we have a straight-to-production approach (ie. anything that is merged should be production ready). The preview branches effectively act as the staging environment. This isn't without problems but it does keep us on our toes. However, if you prefer this approach I would be happy for it to be added to OakComponents. Maybe we should add some CI/CD rules to forbid exports from the waiting room ? I can just see us ending up with perenial waiting room components which are secretly already in prod.

orangemug:
> Generally we have a straight-to-production approach (ie. anything that is merged should be production ready). The preview branches effectively act as the staging environment. This isn't without problems but it does keep us on our toes. However, if you prefer this approach I would be happy for it to be added to OakComponents. Maybe we should add some CI/CD rules to forbid exports from the waiting room ? I can just see us ending up with perenial waiting room components which are secretly already in prod.

The reason being is I think there will probably be some changes to the API over the short term, as I use it. So I want to put it somewhere in OWA while I refine it.

The idea of "waiting room" is that _it is intended_ to be used in production, but not ready for the `oak-component` library just yet. While we develop them and make sure they meet our use cases.

In the past I've taken this approach.

 - Develop components in the app in an isolated place
 - Once a component is used in more than 2 places, extract out into the component library. Reason being until you've used something `<3` times, it isn't generalised enough to be useful outside the app
 
I've got a PR coming behind this one, using this component for the curriculum downloads page.

kimon-satan:
> > Generally we have a straight-to-production approach (ie. anything that is merged should be production ready). The preview branches effectively act as the staging environment. This isn't without problems but it does keep us on our toes. However, if you prefer this approach I would be happy for it to be added to OakComponents. Maybe we should add some CI/CD rules to forbid exports from the waiting room ? I can just see us ending up with perenial waiting room components which are secretly already in prod.
> 
> The reason being is I think there will probably be some changes to the API over the short term, as I use it. So I want to put it somewhere in OWA while I refine it.
> 
> The idea of "waiting room" is that _it is intended_ to be used in production, but not ready for the `oak-component` library just yet. While we develop them and make sure they meet our use cases.
> 
> In the past I've taken this approach.
> 
> * Develop components in the app in an isolated place
> * Once a component is used in more than 2 places, extract out into the component library. Reason being until you've used something `<3` times, it isn't generalised enough to be useful outside the app
> 
> I've got a PR coming behind this one, using this component for the curriculum downloads page.

Ah I see . That makes sense. It doesn't quite align with my rationale for oak-components which goes beyond reusability (we should probably speak about this soon). However, if you want to work that way in curric it's absolutely fine as we are allowed / encouraged to develop our Wows independently in squads.

BTitterington:
Code looks good. I would also second that the naming might be confusing, how do you feel about removing the Oak prefix until the component is actually moved over. Can still have folder `OakComponentsWaitingRoom`

orangemug:
> Code looks good. I would also second that the naming might be confusing, how do you feel about removing the Oak prefix until the component is actually moved over. Can still have folder OakComponentsWaitingRoom

Yeah that makes sense, I'll remove the prefix. Simon also asked us to move it into a curriculum subdirectory. I'll create a little `src/components/CurriculumComponents/OakComponentsKitchen/` and remove the `Oak` prefix from the actual component.

BTitterington (approved):

PR #2396, opened 2024-04-29
LESQ-767: Update copyright year published lessons 

k-huggs:
## Description

Music year: 2004

- Updates year of new published lessons and test files

## Issue(s)

Fixes #LESQ-767

## Acceptance Criteria:

- [ ]  Update licence date on new lessons depending on when they were last published (both lesson overview and downloads)
- [ ]  Legacy content licence copy should remain unchanged

## How to test

1. Go to https://deploy-preview-2396--oak-web-application.netlify.thenational.academy

New lessons should have same year as the updated_at from internal releases table date on both lessons and downloads pages

## Screenshots

Screenshot 2024-04-29 at 15 09 31

How it should now look:
Screenshot 2024-04-29 at 15 09 09


njblore (changes requested):
the year needs to come from the data, we want to keep content published in 2023 saying 2023 but content published this year should say 2024 etc - see the linked slack thread in the ticket

src/components/TeacherComponents/CopyrightNotice/CopyrightNotice.tsx
```diff
@@ -53,39 +54,48 @@ const StyledLink = styled.a`
...
-    <OwaLink
-      page={"legal"}
-      legalSlug="terms-and-conditions"
-      $isInline
-      htmlAnchorProps={{
-        target: props.openLinksExternally ? "_blank" : "_self",
-        "aria-label": `Terms and conditions${
+  props: FontProps & { openLinksExternally: boolean; updatedAt?: string },
+) => {
+  let currentYear;
```
njblore, line 59: can we rename this something more accurate, `copyrightYear` perhaps?

src/components/TeacherComponents/CopyrightNotice/CopyrightNotice.tsx
```diff
@@ -10,6 +10,7 @@ import Icon from "@/components/SharedComponents/Icon";
 type CopyrightNoticeProps = FontProps & {
   showPostAlbCopyright: boolean;
   openLinksExternally: boolean;
+  updatedAt?: string;
```
njblore, line 13: should this be optional?

njblore: aha, in that case it might better to have this component accept a copyrightYear param, and have the logic higher up? because it's not clear here that this is handling that one specific edge case

jack-skerrett-oak (approved):
I think it looks good, bit hard to test because everything has 2024 (a side effect of adding cohorts earlier this year). We could counter that by taking the created_at date of the associated release in interal_releases, rather than the updated_at date, but I think we can live with this for now.

src/components/TeacherComponents/CopyrightNotice/CopyrightNotice.tsx
k-huggs: So this component is also being utilised by Pupils and Curriculum squad, which is another reason why I think I made it optional as it would mean changing their MVs, would it still be worth refactoring?

njblore: I think it would be worth raising that in the cross squad channel, presumably they also need to show the correct copyright year? 
personally I would suggest we update their components so that for now they just pass a hardcoded year and they can update later when it's a priority for them

src/components/TeacherComponents/TermsAgreementForm/TermsAgreementForm.tsx
```diff
@@ -47,6 +48,7 @@ const TermsAgreementForm: FC<TermsAgreementFormProps> = ({
   showSavedDetails = false,
   handleEditDetailsCompletedClick = () => {},
   showPostAlbCopyright = true,
+  updatedAt,
```
njblore, line 51: sorry to nitpick but can this be copyrightYear as well, bit confusing otherwise!

k-huggs: No probs, have just pushed up the change

njblore (approved):

PR #2416, opened 2024-05-09
Feat sticky nav for new plan a lesson - LESQ-782

JBR90:
## Description

Music year: {owa_music_year}

- Adds plan a lesson page nav to page
- Connects to cms
- Crudely adds some content without styling

## Acceptance Criteria:

- [x]  Using “sanity data” create sticky contents menu
- [x]  Ensure bullet-point style exists in Oak Components
- [x]  On mobile it is just displayed at the top
- [x]  Click an item in the contents menu and it takes you to the section
- [x]  Focus moves to the section you have selected
- [x]  At the bottom of each section there is a “Return to contents” link

## Issue(s)

Fixes #

## How to test

1. Go to https://deploy-preview-2416--oak-web-application.netlify.thenational.academy/lesson-planning-new
- Test nav and back to contents buttons

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist

- [ ] Design sign-off

### DIFF
--- a/src/__tests__/pages/lesson-planning-new.test.tsx
+++ b/src/__tests__/pages/lesson-planning-new.test.tsx
@@ -23,6 +23,11 @@ describe("pages/lesson-planning.tsx", () => {
 
     expect(screen.getByRole("heading", { level: 1 }).textContent).toBe("hero");
   });
+  it("Renders a nav", () => {
+    render(<PlanALesson pageData={testPlanningPageData} />);
+
+    expect(screen.getAllByText("Contents")).toHaveLength(2);
+  });
 
   describe("SEO", () => {
     it.skip("renders the correct SEO details", () => {
--- a/src/__tests__/pages/lesson-planning-new.test.tsx
+++ b/src/__tests__/pages/lesson-planning-new.test.tsx
@@ -25,8 +25,11 @@ describe("pages/lesson-planning.tsx", () => {
   });
   it("Renders a nav", () => {
     render(<PlanALesson pageData={testPlanningPageData} />);
-
+    const nav = screen.getByRole("navigation", {
+      name: "plan a lesson contents",
+    });
     expect(screen.getAllByText("Contents")).toHaveLength(2);
+    expect(nav).toBeInTheDocument();
   });
 
   describe("SEO", () => {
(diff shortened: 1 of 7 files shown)


src/__tests__/pages/lesson-planning-new.test.tsx
```diff
@@ -23,6 +23,11 @@ describe("pages/lesson-planning.tsx", () => {
 
     expect(screen.getByRole("heading", { level: 1 }).textContent).toBe("hero");
   });
+  it("Renders a nav", () => {
+    render(<PlanALesson pageData={testPlanningPageData} />);
+
+    expect(screen.getAllByText("Contents")).toHaveLength(2);
```
njblore, line 29: feels like getByRole 'nav' could work here?

njblore (approved):

Monika-Koziol (approved):

src/__tests__/pages/lesson-planning-new.test.tsx
njblore: hmm just looked and it's actually role 'navigation', is that what you tried?

jack-skerrett-oak (approved):
Some a11y bits!

- Because the nav has been given `tabindex=0`, when tabbing through the page (with and without a screen reader) you are navigated to it even though you cannot interact with it and we don't want to be able to focus non-interactive elements
- I think it would be good to add [`aria-current=page`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current) to the selected link in the nav to indicate to a screen reader user that this element is where you currently are. We are going to be doing a similar thing for the l[esson overview nav](https://www.notion.so/oaknationalacademy/I-want-the-lesson-overview-page-nav-to-be-accessible-6bbde63b08fb499e852b48454dbb4067?pvs=4)
- There are `aria-label`s on the lis, I am not sure what they are doing. This means that a screen reader is currently reading "Lesson planning is important, list item 1". If we wanted to make it really clear these are items in a nav list, we could say `aria-label="Scroll to lesson planning is important"` but I don't think it is needed per se

jack-skerrett-oak (changes requested):
I approved but meant to request changes, whoops!

jack-skerrett-oak (approved):

PR #2438, opened 2024-05-21
Feat/lesq 792/unit listing page query [LESQ-792]

njblore:
## Description

Music year: {owa_music_year}

- update `unitListing` query to use the new browse MV
- add query to fetch threads for each unit
- add query to get lesson counts for units
- add queries to get tiers and tier counts
- restructure unitListing query folder so that each query for data and its related schemas, fns and tests are in subfolders: `threads`, `tiers` and `units`
- update to use latest verson of curriculum schema which includes `optionality` in programmeFields
- remove some unused data points from the schema - `quizCount` and individual thread data on a unit - these were always set to null and not used anywhere from what I can tell

## How to test

I'd pay close attention to programmes with tiers, optionality or expired lessons, and ensure thread filtering is working as expected, lesson counts are correct and the order of units & threads has not changes

1. Go to {owa_deployment_url}/teachers/programmes/english-secondary-ks4-aqa/units for a programme with optionality
2. It should appear the same as production

## Screenshots

How it should now look:
Optionality:  
Screenshot 2024-05-21 at 08 28 37
  
Expired:  
Screenshot 2024-05-21 at 08 29 00. 
  
Tiers:  
Screenshot 2024-05-21 at 08 30 52


jack-skerrett-oak (changes requested):
I make it my mission to always find something :D 

- [ ] The key stage in the breadcrumb and above the unit title is no longer sentence case

PR #2440, opened 2024-05-21
Lesq 754] add aria live=polite 

JBR90:
## Description

Music year: {owa_music_year}

- adds aria-live  "polite" to blog copy to clipboard and share page

## Issue(s)

Fixes #

## How to test

1. Go to {owa_deployment_url}
2. Click on \_\_\_\_\_\_

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist

- [ ] Design sign-off

### DIFF
--- a/src/components/SharedComponents/Button/LoadingButton/LoadingButton.tsx
+++ b/src/components/SharedComponents/Button/LoadingButton/LoadingButton.tsx
@@ -21,6 +21,7 @@ type LoadingButtonProps = {
   disabled: boolean;
   success?: boolean;
   ariaLabel?: string;
+  ariaLive?: "off" | "polite" | "assertive";
   onClick: MouseEventHandler<HTMLButtonElement | HTMLAnchorElement>;
 } & (
   | {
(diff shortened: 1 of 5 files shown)


src/components/SharedComponents/Button/LoadingButton/LoadingButton.tsx
```diff
@@ -94,7 +95,7 @@ const ButtonContent: FC<LoadingButtonProps> = (props) => {
   return (
     <>
       <OakFlex $gap="all-spacing-2" $justifyContent="center">
-        <ButtonLabel $color="white">
+        <ButtonLabel aria-live={props.ariaLive} $color="white">
```
njblore, line 98: do we need a default? or does `undefined` equate to `off`?

JBR90: Good spot, prop best to set to off

jack-skerrett-oak (changes requested):
It's a shame, but I don't think there is anyway to get it to keep saying "copied to clipboard" after the first one... oh well!

- [x] I think the blog and webinar buttons should be changed so that their titles specify "link"... "Copy link to clipboard" -> "Link copied to clipboard"
- [x] I don't think the blog and webinar buttons need the aria-label, if I delete it everything still seems to work fine, worth checking on your end as well before we commit (I think we keep the one on the lesson share page because the aria label is different)

jack-skerrett-oak (approved):
beautiful

PR #2466, opened 2024-06-04
[CUR-642] Updated `jose` dependency

orangemug:
## Description

- Updated `jose` as per https://github.com/oaknational/Oak-Web-Application/security/dependabot/91

## Issue(s)

`CUR-642`

## How to test
Check app can make/generate graphql, also mux works correctly. As `jose` is deps of `@mux/mux-node` and `@mux/mux-node`


njblore (approved):
graphql codegen still works, videos still play 👍

PR #2473, opened 2024-06-05
Fix/docx pathways

orangemug:
## Description

Added `pathway` to docx document generation. 

## How to test

1. Go to {owa_deployment_url}/teachers/curriculum/docx-poc/select
2. Generate a document for a subject with pathways
3. Document should be generated correctly


sonalisin (approved):

rightsaidjames (approved):
Overall this looks good. It displays the pathways if they exist for a given subject + year, and the tiers if they exist:

Screenshot 2024-06-06 at 08 37 17
Screenshot 2024-06-06 at 08 26 26
Screenshot 2024-06-06 at 08 26 47

... but in the case of Science it only displays the child subjects and not the tiers.
Screenshot 2024-06-06 at 08 28 29

... however given that we're not distributing Cycle 1 docs anywhere, I don't think this should block merge.

PR #2511, opened 2024-06-17
LESQ-886: New content banner

k-huggs:
## Description

Music year: 1951

- Creates new content banner in OWA using components from component lib
- Adds content banner on unit listing, lesson listing and lesson overview pages
- Links to correct programme or unit listing page as prescribed in this doc: https://docs.google.com/spreadsheets/d/1pE0VtBgSsokQbCrhIhNJrhHfZrnS0U7xy57zsqexmmw/edit#gid=0 

## Issue(s)

Fixes #LESQ-886

## How to test

1. Go to https://deploy-preview-2511--oak-web-application.netlify.thenational.academy
2. Click on any subject with new content
3. You should see banner should be rendered on unit listing, lesson listing and lesson overview pages of all subjects apart from Maths, which is lesson listing and lesson overview
4. Go to resource button should navigate to the new resource unit listing page, unless their are programme factors then programme listing page

## Screenshots

How it should look now: 
Screenshot 2024-06-18 at 10 32 02

Screenshot 2024-06-18 at 10 33 10

Screenshot 2024-06-18 at 10 37 50

## Checklist

- [ ] Design sign-off

### DIFF
--- a/src/components/TeacherComponents/NewContentBanner/NewContentBanner.tsx
+++ b/src/components/TeacherComponents/NewContentBanner/NewContentBanner.tsx
@@ -10,13 +10,6 @@ import {
 
 import { resolveOakHref, OakPageType } from "@/common-lib/urls";
 
-/**
- * ! - Render on the correct pages ✅
- * ! - Link to the correct page - remove - l ✅
- * ! - Fix type error ✅
- * ! - Add images to CMS ✅
- */
-
 const StyledOakFlex = styled(OakFlex)`
   box-shadow: 0px 0px 8px 0px rgba(0, 0, 0, 0.2);
 `;
@@ -26,6 +19,7 @@ type NewContentBannerProps = {
   programmeSlug: string;
   keyStageSlug: string;
   subjectSlug: string;
+  isUnitListing?: boolean;
 };
 
 const renderContentBannerArray = [
@@ -64,6 +58,7 @@ const NewContentBanner: FC<NewContentBannerProps> = ({
   programmeSlug,
   keyStageSlug,
   subjectSlug,
+  isUnitListing,
 }) => {
   const renderComponent = renderContentBannerArray.some((item) => {
     return (
@@ -93,7 +88,7 @@ const NewContentBanner: FC<NewContentBannerProps> = ({
     progSlug = "english-primary-ks2";
   }
 
-  if (!renderComponent) {
+  if (!renderComponent || (isUnitListing && subjectSlug === "maths")) {
     return null;
   }
 
--- a/src/components/TeacherComponents/NewContentBanner/NewContentBanner.tsx
+++ b/src/components/TeacherComponents/NewContentBanner/NewContentBanner.tsx
@@ -8,7 +8,14 @@ import {
   OakImage,
 } from "@oaknational/oak-components";
 
-import { resolveOakHref } from "@/common-lib/urls";
+import { resolveOakHref, OakPageType } from "@/common-lib/urls";
+
+/**
+ * ! - Render on the correct pages ✅
+ * ! - Link to the correct page - remove - l ✅
+ * ! - Fix type error ✅
+ * ! - Add images to CMS ✅
+ */
 
 const StyledOakFlex = styled(OakFlex)`
   box-shadow: 0px 0px 8px 0px rgba(0, 0, 0, 0.2);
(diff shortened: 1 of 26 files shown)


src/components/TeacherComponents/NewContentBanner/NewContentBanner.tsx
```diff
@@ -0,0 +1,162 @@
...
+  { keyStageSlug: "ks4", subjectSlug: "biology" },
+  { keyStageSlug: "ks4", subjectSlug: "chemistry" },
+  { keyStageSlug: "ks4", subjectSlug: "combined-science" },
+  { keyStageSlug: "ks1", subjectSlug: "maths" },
+  { keyStageSlug: "ks2", subjectSlug: "maths" },
+  { keyStageSlug: "ks3", subjectSlug: "maths" },
+  { keyStageSlug: "ks4", subjectSlug: "maths" },
+];
+
+const removeSuffix = (str: string) => {
```
njblore, line 49: we have a method for this in utils `removeLegacySuffix`

src/components/TeacherViews/LessonOverview/LessonOverview.view.tsx
```diff
@@ -222,6 +223,12 @@ export function LessonOverview({ lesson }: LessonOverviewProps) {
         showDownloadAll={showDownloadAll}
       />
       <MaxWidth $ph={16} $pb={80}>
+        <NewContentBanner
+          keyStageSlug={keyStageSlug ?? ""}
```
njblore, line 227: in what cases is keyStageSlug undefined? what happens if the banner receives empty strings for the props?

src/components/TeacherComponents/NewContentBanner/NewContentBanner.tsx
```diff
@@ -0,0 +1,162 @@
...
+        $width={"all-spacing-18"}
+        $height={"all-spacing-15"}
+        $alignItems={"center"}
+      >
+        <OakImage
+          $display={"block"}
+          $width={"all-spacing-18"}
+          $height={"all-spacing-15"}
+          alt="new content banner image"
+          src={`https://${process.env.NEXT_PUBLIC_OAK_ASSETS_HOST}/${process.env.NEXT_PUBLIC_OAK_ASSETS_PATH}/v1718639479/teacher-journey/content-banner.png`}
```
njblore, line 155: think you can use `getBrowserConfig` for these values

src/components/TeacherComponents/NewContentBanner/NewContentBanner.tsx
k-huggs: ✅

src/components/TeacherViews/LessonOverview/LessonOverview.view.tsx
k-huggs: I believe it's just Specialist pages that don't have keyStageSlug available both canonical and pathway pages should have it available

src/components/TeacherComponents/NewContentBanner/NewContentBanner.tsx
```diff
@@ -0,0 +1,162 @@
...
+  { keyStageSlug: "ks3", subjectSlug: "history" },
+  { keyStageSlug: "ks3", subjectSlug: "science" },
+  { keyStageSlug: "ks4", subjectSlug: "biology" },
+  { keyStageSlug: "ks4", subjectSlug: "chemistry" },
+  { keyStageSlug: "ks4", subjectSlug: "combined-science" },
+  { keyStageSlug: "ks1", subjectSlug: "maths" },
+  { keyStageSlug: "ks2", subjectSlug: "maths" },
+  { keyStageSlug: "ks3", subjectSlug: "maths" },
+  { keyStageSlug: "ks4", subjectSlug: "maths" },
+];
```
njblore, line 49: don't need to change since this is a temporary component (I think?) but just pointing out this seems like kind of a strange data structure to use for this! I would have thought a record would be make looking up subjects by keystage straightforward

src/components/TeacherViews/LessonOverview/LessonOverview.view.tsx
njblore: ok, should be safe here because the banner component has an early return if the keystage isn't in the lookup, but I'd usually prefer we handle undefined values rather than defaulting to empty strings!

njblore (approved):

src/components/TeacherComponents/NewContentBanner/NewContentBanner.tsx
k-huggs: Changed this to a record ✅

Monika-Koziol (changes requested):
Hey one change:

- [ ] on the unit listing page: can the banner be the same width as the unit box?
Questions: I'm not sure if this part of this ticket but when I clicked on the link it took me to the new listing page (great) but there was another banner there. The banner shouldn't be on the new unit listing pages.

Monika-Koziol (approved):

jack-skerrett-oak (changes requested):
I am assuming Monika caught all the design related bits, so these are just some extras from me because I always have more!

- [ ] Could the images (chevron and main banner image) have alt="". Currently they have annoying alt text and I don't think either require any
- [ ] Could the heading be changed to a h2. Currently on all pages (unit listing, lesson listing) it skips from a h1 to a h3, which we don't like
- [ ] Bit fiddly, but for the English links, could we have the subject capitalised so it appears "English" and not "english". All the other subjects are fine. (The rule extends to the other languages, and also abbreviations like PE/RE, but English is the only one in this set)

jack-skerrett-oak (changes requested):
We also want links from legacy lesson listing and lesson overview to new unit listing/programme listing as per the gsheet. I'm assuming the combined maths unit pages will be at the "new" url (/maths-secondary-ks4-foundation/units, no -l) so these should remain compatible even when the combined unit listing page exists

Screenshot 2024-06-19 at 16 28 58

jack-skerrett-oak (approved):

PR #2516, opened 2024-06-18
LESQ-892: combined programme listing page

k-huggs:
## Description

Music year: 2012

- Combined programme listing page for ks4 Maths
- Adds tag to programme listing page

## Issue(s)

Fixes #LESQ-892

## How to test

1. Go to https://deploy-preview-2516--oak-web-application.netlify.thenational.academy

## Screenshots

How it should now look:
Screenshot 2024-06-19 at 11 48 31


njblore:
I'm not sure we're actually doing the new tag on the tier buttons, I asked monika on the ticket and as I understood it we are only having it on the subject icon!
also the non-ks4 maths are linking to programme listing page instead of units which would be good to resolve here

src/pages/teachers/key-stages/[keyStageSlug]/subjects/index.tsx
```diff
@@ -168,7 +168,8 @@ export const getStaticProps: GetStaticProps<
 
         const combinedMaths: KeyStageSubjectData = {
           programmeSlug: newMaths.programmeSlug,
-          programmeCount: newMaths.programmeCount + legacyMaths.programmeCount,
+          programmeCount:
+            newMaths.programmeCount + legacyMaths.programmeCount - 1,
```
Comment (njblore) [inline src/pages/teachers/key-stages/[keyStageSlug]/subjects/index.tsx:172]: thinking about this I think it should actually be that we take the higher programme count? otherwise if each have 2 it will become 3, if each had 6 it would become 11 but there are still only 6 programmes? Might get weird if they have different programes...but I don't think we need to worry about that

Comment (k-huggs) [reply src/pages/teachers/key-stages/[keyStageSlug]/subjects/index.tsx]: Changed

njblore (approved):

PR #2519, opened 2024-06-19
[CUR-682] Added a11y skip link to curriculum visualiser

orangemug:
## Description
Added skip links to curriculum visualiser and a `<SkipLink/>` component to `./CurriculumComponents/OakComponentsKitchen`

## Issue(s)

`CUR-682`

## How to test

1. Go to https://deploy-preview-2519--oak-web-application.netlify.thenational.academy/teachers/curriculum/science-secondary-aqa/units
2. Tab through content until you see the "Skip to units" should appear when highlighted.
3. Upon hitting enter, focus should jump to top of units 

Screenshot 2024-06-19 at 17 50 14

### DIFF
--- a/src/__tests__/__helpers__/renderWithProviders.tsx
+++ b/src/__tests__/__helpers__/renderWithProviders.tsx
@@ -19,6 +19,7 @@ import { MemoryRouterProvider } from "next-router-mock/MemoryRouterProvider";
 import { ThemeProvider } from "styled-components";
 import { OverlayProvider } from "react-aria";
 import { MemoryRouterProviderProps } from "next-router-mock/dist/MemoryRouterProvider/MemoryRouterProvider";
+import { OakThemeProvider, oakDefaultTheme } from "@oaknational/oak-components";
 
 import "../../browser-lib/oak-globals/oakGlobals";
 import ErrorBoundary from "../../components/AppComponents/ErrorBoundary";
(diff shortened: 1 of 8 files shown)


src/__tests__/__helpers__/renderWithProviders.tsx
```diff
@@ -58,6 +60,7 @@ const providersByName: {
 } = {
   cookieConsent: [MockedCookieConsentProvider],
   theme: [ThemeProvider, { theme }],
+  oakTheme: [OakThemeProvider, { theme: oakDefaultTheme }],
```
BTitterington, line 63: This seems reasonable. It did occur to me that others may have used renderWithProviders() and an OakThemeProvider, meaning there could be 2 OakThemeProviders providers now. But I don't think that would cause any issues anyway.

BTitterington (approved):

src/__tests__/__helpers__/renderWithProviders.tsx
orangemug: Yeah it took me a while to work out why the tests weren't working 🙂

rightsaidjames (approved):
Looks good - tested with a keyboard and also with VoiceOver, the skip link is discoverable and skips past the thread and year filter options.

@HelenHarlow I note that there is no skip link on mobile, I think this is okay because the threads are inside a modal so the only thing you have to tab past is the years, but is this intentional?

HelenHarlow:
> Looks good - tested with a keyboard and also with VoiceOver, the skip link is discoverable and skips past the thread and year filter options.
> 
> @HelenHarlow I note that there is no skip link on mobile, I think this is okay because the threads are inside a modal so the only thing you have to tab past is the years, but is this intentional?

Yess I left out mobile for exactly the reason you stated there, there are far fewer buttons because the threads are contained in a modal

HelenHarlow (approved):
Looks good to me!

PR #2532, opened 2024-06-24
[CUR-685] Pass in current `tier_slug` because it isn't set on all units

orangemug:
## Description
`tier_slug` isn't set on all units because some are a part of multiple tiers. Therefore we must pass in `tier_slug` into `createProgrammeSlug(...)` 

## Issue(s)

`CUR-685`

## How to test
See `CUR-685`


njblore (approved):

rightsaidjames (approved):
Looks good 👍

PR #2538, opened 2024-06-26
Remove WDIO

jimCresswell:
We created some example [WDIO](https://webdriver.io/) end-to-end browser tests quite a while ago to serve as an example of what was possible. Because of Percy we haven't had much need of them, and the code is now quite out of date.

Any potential future personalisation work may introduce journeys complex enough to warrant E2E browser tests, but if so it would be better to reevaluate the available options.

This PR also updates a few dependencies that were causing security warnings.

To test: the app should build and PR checks should pass.


orangemug (approved):
Looks good, only note is this PR updates deps outside of removing WDIO. This would name it hard to roll back just this change.

It's a minor issue though

PR #2546, opened 2024-06-27
[PUPIL-816] Add min widths to prevent cumulative layout shift on year selection page

BTitterington:
## Description

Music year: 1993

Added an extra oak box around the year links that has a min-width slightly larger than the rendered text size. 

Too be honest, I'm not 100% certain this is going to fix the web vitals score, I've used the performance tab to look for layout shifts and I still see one when the font loads, although it's not the whole white box now, it's just the "Welcome to Oak!" div.

I think we merge this and see if the score has improved in a week, if not we can look at this in more detail and maybe draft in some more opinions.

## How to test

1. Go to https://deploy-preview-2546--oak-web-application.netlify.thenational.academy/pupils/years
2. Refresh the page using Cmd (Command) + Shift + R
3. You should see the page reload without the white box changing size

| Current | New |
|----------|----------|
| cls years before | cls years after |

## Checklist

- [ ] Design sign-off
- [x] Does this PR update a package with a breaking change


RBrownsett:
@BTitterington I like the before and after videos! One thing I did notice on refreshing the font would sometimes take time to load through (only half a second or so). I've attached a video to show - around the 19 second mark is at its worst (takes around a second) the other times its not that noticeable

https://github.com/oaknational/Oak-Web-Application/assets/30384846/916e6b1f-de6c-4466-850a-d0b05b2a6695

BTitterington:
@RBrownsett You can go into the performance tab of developer tools and then settings and throttle the network connection to "Slow 3G" to see the font load slowly, but I don't think we can fix that without impacting performance across the whole site. There are different font loading options, we are set to `swap` which is generally the best, although there may be an argument to use `fallback`. You can read this on the different options if you're interested - https://blog.sentry.io/web-fonts-and-the-dreaded-cumulative-layout-shift/#use-the-font-display-property

RBrownsett:
@BTitterington just checked mobile view via browserstack and it looks like its had an effect when viewing via mobile- the white space to the right of the screen has been added
Old:
image

New:
image

Dev Tools:
image

RBrownsett (approved):
Approved - the display on mobile is now correct. Although not caused by this PR I will raise a ticket to look at the options to mitigate against the font load issue on the page

benprotheroe (approved):
lgtm!

PR #2559, opened 2024-07-03
fix: only combine unit listing pages for ks1-4 maths [LESQ-949]

njblore:
## Description

Music year: 2003

- EYFS was being excluded in the logic for combining new and legacy maths, as there is only legacy content
- this includes EYFS for maths subject only and doesn't combine counts

## Issue(s)

Fixes #
Maths EYFS gone missing from the site

## How to test

1. Go to https://deploy-preview-2559--oak-web-application.netlify.thenational.academy/teachers/key-stages/early-years-foundation-stage/subjects
3. You should see a card for maths
4. Click through and the counts should be correct
5. There should be no 'new' tag on any of the content

## Screenshots

Screenshot 2024-07-03 at 08 34 33

How it should now look:
Screenshot 2024-07-03 at 08 34 44

### DIFF
--- a/src/__tests__/pages/teachers/key-stages/[keyStageSlug]/subjects/index.test.tsx
+++ b/src/__tests__/pages/teachers/key-stages/[keyStageSlug]/subjects/index.test.tsx
@@ -195,5 +195,39 @@ describe("pages/key-stages/[keyStageSlug]/subjects", () => {
       expect(maths?.old).toBeNull();
       expect(maths?.new?.subjectSlug).toBe("maths");
     });
+    it("should not combine counts for EYFS maths", async () => {
+      const mockEyfsMathsResponse = {
+        keyStageSlug: "early-years-foundation-stage",
+        keyStageTitle: "Early years foundation stage",
+        subjects: [
+          {
+            subjectSlug: "maths",
+            subjectTitle: "Maths",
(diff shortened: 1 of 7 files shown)


jack-skerrett-oak (changes requested):
Shame this made it live, thought I had checked EYFS! 
- [x] Seems to be suffering from the same legacy breadcrumb issue. Clicking "Maths" in the breadcrumb takes a user to the legacy listing page, which doesn't exist

rightsaidjames (approved):

rightsaidjames (approved):

jack-skerrett-oak (commented):
@njblore the solution we have gone for is to have a legacy unit listing slug on the new style page (maths-foundation-early-years-foundation-stage-l), which is different to what it was whilst it had the breaking breadcrumb issue (maths-foundation-early-years-foundation-stage). I assume that is just an easier solution, and as we won't have new content it doesn't matter? Just felt a bit weird maintain the -l

njblore:
@jack-skerrett-oak yeah it's basically easier to leave it on the -l url, otherwise we'd have to change how new tags are applied throughout the journey - plus this way we don't need to redirect anyone

jack-skerrett-oak (approved):

PR #2621, opened 2024-07-16
fix: ensure review section started is tracked

kimon-satan:
## Issue(s)

Since April 5th, the number of users who reach the lesson review is zero (sequentially through the core journey). Looking just 2 steps: step 1 (lesson start) and step 2 (lesson review), then since 18th April there has been zero users hitting the lesson review page. The numbers have gone from 100s to 10s. This might still be working, though concerning.

Fixes #

this was for several interconnected issues:
- Avo did not have a valid section for "review"
- The LessonEngineProvider state reducer changes currentSection to review and this is picked up by a useEffect hook and so the other methods don't pick up the event
- completeSection had a check for all sections being completed but this didn't work because it relied on state which only changes on the subsequent render

- I have made a new avo tracking plan, fixed the issues in  LessonEngineProvider  and added code to completeSection to track when the user is redirected to review 

## How to test

1. Go to https://deploy-preview-2621--oak-web-application.netlify.thenational.academy/pupils/lessons/transverse-waves
2. Complete the lesson with the console open
3. You should see that the lessonSectionStarted event is sent to posthog with the property `Pupil Experience Lesson Section: "review"`

## ACS

[x] Lesson Review events should be tracked in a similar way to other events properties: Lesson Start, Lesson Quiz, etc.


RBrownsett (approved):
Approved - property appears as expected

image

BTitterington (approved):

PR #2636, opened 2024-07-24
[CURR-736] Child subject & tier filter downloads journey

sonalisin:
## Description

Music year: 2020

- Added child subject/tier filtering for relevant subjects in the curriculum downloads journey

## How to test

1. Go to https://deploy-preview-2635--oak-web-application.netlify.thenational.academy
2. Navigate to Secondary Maths - you should only see the tier filter
3. Navigate to Secondary Science (all exam boards) - you should see both filtering by child subject and tier
4. No primary programmes should have the selector

## Figma reference
https://www.figma.com/design/3cz9fb0ObFDLv1c212OtJv/Final-designs---Curriculum-squad?node-id=1903-44116&t=PBV2mNxaHpW0w1ju-0

## Checklist

- [ ] Design sign-off


src/components/CurriculumComponents/CurriculumDownloadTab/index.tsx
```diff
@@ -65,7 +101,15 @@ const CurriculumDownloadTab: FC<CurriculumDownloadTabProps> = ({
         schools: [],
       });
     }
-  }, [localStorageData]);
+
+    // Set the subject tier selector as visible when tiers & child_subjects are present
+    if (
```
orangemug, line 106: Any reason this needs to be in the same `useLayoutEffect` hook?

src/components/CurriculumComponents/CurriculumDownloadTab/index.tsx
```diff
@@ -107,9 +164,15 @@ const CurriculumDownloadTab: FC<CurriculumDownloadTabProps> = ({
...
-    ].join("/");
+      tierSelected,
+      childSubjectSelected,
+    ]
+      .filter(Boolean)
+      .join("/");
+
     const downloadPath = `/api/curriculum-downloads/${mvRefreshTime}/${slug}`;
 
+    console.log(slug);
```
orangemug, line 175: `console.log(...)`

src/components/CurriculumComponents/CurriculumDownloadTab/index.tsx
sonalisin: I had them initially in two different useLayoutEffects - but thought it looked a bit neater. Would it make more sense for there to be two separate ones?

src/components/CurriculumComponents/CurriculumDownloadTab/index.tsx
```diff
@@ -148,7 +211,14 @@ const CurriculumDownloadTab: FC<CurriculumDownloadTabProps> = ({
   return (
     <OakThemeProvider theme={oakDefaultTheme}>
       <Box $maxWidth={1280} $mh={"auto"} $ph={18} $pb={[48]} $width={"100%"}>
-        {!isLoading && (
+        {subjectTierSelectionVisible === true && (
```
orangemug, line 214: Minor: This is probably clearer if you use a "mode" or something like that. So `mode` would have two type `"tier_selection" | "download"` and then you'd render the children based off of that.

**Note**: No need to fix this now, but just for reference.

PR #2649, opened 2024-07-30
[ENG-798] Renew the security.txt key

tweakster:
## Description

- New public key
- Use of the renewed key to sign message

## How to test

Test 1

1. Create a dummy message
2. Encrypt it with the public key
3. Download the private key and use it to decrypt the message

Test 2

1. Use the public key to validate the signed security.txt

# Checklist
Upload the public key to Openpgp


mantagen (approved):
Code looks good, though wasn't able to test functionality

PR #2670, opened 2024-08-06
Added missing `aria-describedby` to phase/examboard errors

orangemug:
## Description
Added missing `aria-describedby` to phase/examboard errors

## Issue(s)

`CUR-457`

## How to test

1. Go to https://deploy-preview-2670--oak-web-application.netlify.thenational.academy/teachers/curriculum
2. Test phase/examboard error messages are announced to screen reader users


rightsaidjames (approved):
Now, if you trigger the error messages then open up the subject modal and tab to one of the buttons, NVDA (Firefox) says this:
> Curriculum plans grouping required select a subject to view a curriculum English radio button 1 of 6

And with exam boards:
> Choose an exam board for KS4 grouping required select an exam board to view the curriculum AQA radio button not checked 1 of 3

It's only this verbose for the first button you land on in a grouping, if you move between them it doesn't read out the group or the error message again until you leave the group and come back to it.

In short, this has been implemented according to the spec and I can confirm that it works with NVDA. There's something funky about radiogroup support in macOS VoiceOver, but it's not our job to fix AT bugs.

sonalisin (approved):
Not super important now, but some of the error states might benefit from comments to see when they're triggered

PR #2672, opened 2024-08-06
Fix/pupil 856/legacy content order

kimon-satan:
## Description

We need to make sure that back navigation makes sense in all usecases of the lesson listing page. These are as follows

- Direct navigation
  - Non-legacy
  - Legacy - equivalent exists in non-legacy
  - Legacy - no equivalent in non-legacy & non-legacy has programme-factors
  - Legcay - no equivalent in non-legacy & non-legacy doesn't have programme-factors
- Navigation via Pupil Browse

The expected outcomes are as follows

- Direct navigation
  - Non-legacy => back to programme units page
  - Legacy - equivalent exists in non-legacy => back to non-legacy units page
  - Legacy - no equivalent in non-legacy & non-legacy has programme-factors => back to non-legacy/options page
  - Legacy - no equivalent in non-legacy & non-legacy doesn't have programme-factors =>  back to non-legacy units page
- Navigation via Pupil Browse
 -  back to the non-legacy page you navigated from

## ProgrammeSlug Validation

Running regex on programmeSlugs leaves us vulnerable to DDOS so src/utils/validateProgrammeSlug.ts calculates the maximum length of programmeSlug and throws an error if it's exceeded. Then it validates slugs against our zod schema.
This could be useful in getStaticProps of other pages too

## How to test

1. Go to https://deploy-preview-2672--oak-web-application.netlify.thenational.academy/pupils/programmes/combined-science-secondary-year-10-foundation-l/units/cells-biology-3p8njnj/lessons
2. hit the back button you should end up at pupils/programmes/combined-science-secondary-year-10/options
3. choose aqa and foundation then navigate back to the legacy unit cell biology 
4. Hit the back button 
5. you should arrive at pupils/programmes/combined-science-secondary-year-10-foundation-aqa/units

## Checklist

- [x] direct navigation handled (all use cases)
- [x] browse navigation handled
- [x] tests


messmike (approved):

src/utils/validateProgrammeSlug.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    jest.spyOn(console, "error").mockImplementation(() => {});
+  });
+
+  it("should throw an error if the programmeSlug is too long", () => {
+    expect(() => {
+      validateProgrammeSlug("abracodabra-".repeat(100));
+    }).toThrow("programmeSlug is too long. Max length is 100");
+  });
+
+  it.only.each([
```
orangemug, line 14: Assume you don't want this `.only` here

RBrownsett (approved):
Approved - tested the following.

- [x] legacy back to legacy content
- [x] legacy back to cycle 1 and legacy
- [x] cycle 1 back to cycle 1
- [ ] direct legacy back to examboard page (subject category filters)

src/utils/validateProgrammeSlug.test.ts
kimon-satan: doh

orangemug (approved):

PR #2675, opened 2024-08-07
Feat/lesq 975/sign out button [LESQ-975]

njblore:
## Description

Music year: 2007

- adds a sign out button to the hamburger menu when a user is signed in
- it is not behind the feature flag as I didn't think it was necessary when you can only sign in behind the feature flag - lmk if you disagree on this

## How to test

1. Go to https://deploy-preview-2675--oak-web-application.netlify.thenational.academy (any page) while signed in
3. You should see the sign out button
4. Click on sign out and reopen the hamburger menu
5. You should not see the sign out button
6. When the feature flag is not active and you are not signed in you should not see the sign out button

## Screenshots
How it should now look:
Screenshot 2024-08-07 at 09 04 17

## Checklist

- [ ]  A button in the ~~navbar~~ hamburger menu should show when a user is logged in
- [ ]  When no user is logged in there is no button
- [ ]  The sign out button should appear across OWA products - curriculum, pupils, teachers etc


JBR90 (approved):

k-huggs (approved):

src/components/AppComponents/AppHeaderBurgerMenuSections/AppHeaderBurgerMenuSections.tsx
```diff
@@ -22,8 +27,15 @@ const AppHeaderBurgerburgerMenuSections: FC<
   AppHeaderBurgerburgerMenuSectionsProps
 > = (props) => {
   const { burgerMenuSections } = props;
+  const { user } = useUser();
+
   return (
     <OakFlex $flexDirection="column" $gap="all-spacing-7">
+      {user && (
+        <OakSecondaryButton element="a" href="/api/auth/oidc-logout">
```
carlmw, line 35: This can be `/api/auth/logout`. I only added the `/oidc-logout` endpoint as a testing convenience. The difference between the two is subtle. `/logout` will clear the user's session in OWA, while `/oidc-logout` will clear the user's session in OWA and Auth0. It gives us a clean slate in Auth0 making it easier to test with different accounts.

carlmw (approved):
LGTM, just a point about which endpoint to use to log the user out.

src/components/AppComponents/AppHeaderBurgerMenuSections/AppHeaderBurgerMenuSections.tsx
carlmw: It only really becomes relevant when users can login to multiple applications.

This is a really helpful guide to the sessions https://auth0.com/docs/authenticate/login/logout

Probably helps to think of Auth0 as an auth platform and not login for a single app. 

Going forward we might want users to be able to login to Aila or to give 3rd parties access to APIs etc. and that can all go through Auth0. If the user has a session in Auth0 and the user has previously given the app access (I.e. logged into it before) then those logins can be silent with no prompt. If we trash that session on logout of any one app then the user will have to login again for any other app.

We can make that decision and alter the default behaviour of `/logout` to do a "full" logout if we want to. Or we could configure `/login` in OWA to always display the login prompt if we want to go that route.

That was why I added the `/oidc-logout` because when testing it's not helpful to be automatically logged in if you want to switch to another test account!

jack-skerrett-oak (approved):
Sign out appears only if you are signed in and have posthog configured. It renders throughout OWA. Selecting the button signs the user out and they have to log back in when navigating to the downloads page. 

We might want to look at iterating on where the user is moved to once logging out, moving back to the home page is slightly jarring.

src/components/AppComponents/AppHeaderBurgerMenuSections/AppHeaderBurgerMenuSections.tsx
njblore: thanks that's really helpful!

PR #2677, opened 2024-08-07
Update to `published_mv_curriculum_units_0_6`

orangemug:
## Description
Update to `published_mv_curriculum_units_0_6` which filters out `year = "All"` in MV (see [here](https://github.com/oaknational/Database-Tools/blob/6c7492cd39b2fe143f3ec0d2f96fdd2ead311e59/sql-and-graphql-library/SQL/owa-materialized-views/mv_curriculum_units.sql#L245))

## Issue(s)

Fixes `CUR-799`

## How to test
Check that the curriculum visualiser is still in working order (no regressions)


rightsaidjames (approved):
Checked all of the lots and all of the curriculum visualiser features - everything seems to be in order!

PR #2689, opened 2024-08-12
Import function rather than module for `lodash` to improve performance

orangemug:
Import `lodash` with ES6 destructuring to improve performance

Before
Screenshot 2024-08-12 at 14 07 32

After
Screenshot 2024-08-12 at 14 09 00


rightsaidjames (approved):
Tested the subject category filtering in the pupil journey - working as before 👍

sonalisin (approved):

PR #2691, opened 2024-08-13
Feat/lesq 1006/onboarding role selection [LESQ-1006]

njblore:
## Description

Music year: {owa_music_year}

- refactors onboarding page into `role-selection` and `school-selection` sub routes
- extracts onboarding form into wrapper component
- create schema file for form types
- create new form for role selection

## How to test

With the feature flag enabled

1. Go to https://deploy-preview-2691--oak-web-application.netlify.thenational.academy/onboarding/role-selection and /school-selection
2. Click on the options, including other, see error states etc
3. There is a console log on successful submit
4. The newsletter checkbox should still be present, as well as Ts and Cs, across both onboarding pages
5. The journey should work from `/onboarding` -> select no -> arrive at `/onboarding/role-selection`

## Screenshots
Screenshot 2024-08-13 at 11 18 00

## Checklist

- [ ] Nothing selected by default, if a user selects “Continue” without selecting an option, error should say 'select a role'
- [ ] If other is selected, required text input displayed
- [ ] If a use selects continue without inputting a role into the text box, error should say 'please tell us what your role is'
- [ ] error should clear when a role is selected or input
- [ ] newsletter checkbox toggles


src/pages/onboarding/role-selection.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { useFeatureFlagEnabled } from "posthog-js/react";
+import { useEffect } from "react";
+
+import RoleSelectionView from "@/components/TeacherViews/Onboarding/RoleSelection/RoleSelection.view";
+
+const RoleSelectionPage = () => {
+  const ffEnabled = useFeatureFlagEnabled("use-auth-owa");
+  const router = useRouter();
+
+  // TODO: extract into shared hook
```
njblore, line 16: I saw @JBR90 had done this in his PR so leaving this here for now

src/components/TeacherViews/Onboarding/RoleSelection/RoleSelection.view.tsx
```diff
@@ -0,0 +1,92 @@
...
+        formState.errors.role !== undefined &&
+        formState.errors.other !== undefined
+      }
+      onSubmit={() =>
+        getValues().role === "other" &&
+        !getValues().other &&
+        setError("other", { message: "Please tell us what your role is" })
+      }
+      control={control as Control<OnboardingFormProps>}
+      trigger={trigger as UseFormTrigger<OnboardingFormProps>}
```
carlmw, line 63: 🏄🏻 nothing draws my attention like a cast

Looking at the code `OnboardingForm` is only concerned with the newsletter sign-up field so I reckon it could be made generic with something like this:

```typescript
const OnboardingForm = <FieldValues extends { newsletterSignUp: boolean }>(props: {
  children: React.ReactNode;
  handleSubmit: UseFormHandleSubmit<FieldValues>;
  formState: UseFormStateReturn<FieldValues>;
  heading: string;
  continueDisabled: boolean;
  onSubmit?: () => void;
  control: Control<FieldValues>;
  trigger: UseFormTrigger<FieldValues>;
})
```

that would spare the casting of `control` and `trigger` leave `OnboardingForm` unconcerned with the fields for the other forms. 

but tbh it's a lot of typescript for very little gain. It's a fun exercise though.

src/components/TeacherViews/Onboarding/RoleSelection/RoleSelection.view.tsx
```diff
@@ -0,0 +1,92 @@
...
+const roleOptions: Record<string, string> = {
+  "teacher-training": "Training to become a teacher",
+  "teacher-trainer": "Teacher trainer",
+  "private-tutor": "Private tutor",
+  "adult-helping-child":
+    "Adult helping a child, e.g. with revision, homeschooling",
+  "mat-or-lea": "Working at multi-academy trust or local educational authority",
+  nonprofit: "Working at an educational nonprofit",
+  other: "Other",
+};
```
carlmw, line 23: ackroydperfect

src/components/TeacherViews/Onboarding/SchoolSelection/SchoolSelection.view.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  return (
+    <OnboardingForm
+      control={control as Control<OnboardingFormProps>}
+      trigger={trigger as UseFormTrigger<OnboardingFormProps>}
+      formState={formState}
+      heading="Select your school"
+      handleSubmit={handleSubmit}
+      continueDisabled={
+        formState.errors?.school !== undefined || !formState.isValid
```
carlmw, line 65: This is just for me because i'm new to `react-hook-form`, but what does this line actually do? What does `formState.errors?.school !== undefined` do that  `!formState.isValid` doesn't?

Also the negation is really hard on my brain. Could it be flipped for easier comprehension?

```typescript
const canSubmit = formState.errors?.school === undefined && formState.isValid
...
continueDisabled={!canSubmit}
```

carlmw (approved):
No blockers for me. Just some reckons and a question.

src/components/TeacherViews/Onboarding/RoleSelection/RoleSelection.view.tsx
njblore: you're totally right to call it out! it's so unpleasant - it doesn't quite work with making it generic though as other parts of the form then complain that they're missing fields. I found a [few](https://github.com/react-hook-form/react-hook-form/issues/9287) [similar](https://github.com/react-hook-form/react-hook-form/issues/6726) sounding issues on the library with no solution 🫠

src/components/TeacherViews/Onboarding/RoleSelection/RoleSelection.view.tsx
njblore: not sure how to interpret this 😂

src/components/TeacherViews/Onboarding/RoleSelection/RoleSelection.view.tsx
carlmw: I had a go and got it working with generics all the way up to the point I encountered that issue with `Path<T>` you mentioned.

src/components/TeacherViews/Onboarding/SchoolSelection/SchoolSelection.view.tsx
njblore: hmm good question, so `isValid` will be false from the start as the required fields are not filled out, and I think it will also be false if there are errors set, but the errors will only be set at certain times eg when attempting to continue when the form is invalid - so I'm not sure we do need the check for errors here!

src/components/TeacherViews/Onboarding/RoleSelection/RoleSelection.view.tsx
carlmw: it's good! I meant to suggest mapping the roles to keys and only storing the key in planning. The copy will probably change 😉

src/__tests__/pages/onboarding/onboarding.test.tsx
JBR90, line 1: Why was this test file deleted ?

njblore: looks like it was a casualty of the merge! I'll reinstate

JBR90 (approved):

src/components/TeacherComponents/OnboardingForm/OnboardingForm.tsx
```diff
@@ -0,0 +1,139 @@
...
+                id="newsletterSignUp"
+              />
+            );
+          }}
+        />
+      </OakFlex>
+
+      <OakP $font="body-2" color="text-primary" $textAlign="center">
+        By continuing you agree to{" "}
+        <OwaLink
```
JBR90, line 93: Does this want to be an Oak link ?

jack-skerrett-oak (changes requested):
Screenshot 2024-08-14 at 13 47 30

- [x] The spacing above and below the error message is too big, it looks like it should be 24px
- [x] Could the field label be updated to "Please specify (required)" as per the figma
- [ ] I'm not sure how you would go about this but we will need to include the "required" attribute here, both for the radio buttons and also for the text input when "other" is selected
- [ ] I've also asked Monika if we can change "please specify" to "your role" as it is more informative for users
- [x] There is no place holder text in the "other" text input
- [ ] Not sure if it is a problem, but I noticed that the error that occurs when you don't type anything in the text input has an id of "other-error"
- [x] If you select "Other", don't enter a value, press continue, get the error, select another option from the list (without entering anything in the other text input), you still cannot proceed. I assume this is because of the hidden, erroring other input.
- [x] When the error for the "other" text input is present, the error can be cleared by toggling the marketing checkbox (though the error reappears when you try to continue again in an erroring state).

The issues with the h2 etc still exist but I think we expect them to so ignoring!

JBR90 (approved):

jack-skerrett-oak (approved):
The grouping of the radio buttons is being funny but I know Joe has been looking at that on another PR so best to wait I imagine.

PR #2720, opened 2024-08-20
Feat/i want a user to be geolocated at sign up - LESQ-1026

JBR90:
## Description

Music year: 2012

- Uses "x-country" header to set location 
- Sets DEVELOPMENT_USER_REGION  env for development (added to dev config)

## How to test

1. Clear any public meta data in clerk
2. Go to https://deploy-preview-2720--oak-web-application.netlify.thenational.academy/onboarding
3. Click through onboarding steps
4. Check clerk meta data - It should show "region" data
`{ "owa": { "isTeacher": false, "isOnboarded": true }, "region": "GB", "sourceApp": "https://deploy-preview-2720--oak-web-application.netlify.thenational.academy" }`

5. Clear public meta data in clerk
6. Use proxy to use different location
7. repeat steps above - you should see different country code
`{ "owa": { "isTeacher": false, "isOnboarded": true }, "region": "US", "sourceApp": "https://deploy-preview-2720--oak-web-application.netlify.thenational.academy" }`

8. disable proxy and logout/login - original signup region should still exist

## Checklist

- [ ] Design sign-off


src/app/api/auth/onboarding/route.ts
```diff
@@ -13,9 +14,22 @@ export async function POST(req: Request) {
   try {
     const owaData = onboardingSchema.parse(await req.json());
     const sourceApp = user.publicMetadata.sourceApp ?? getReferrerOrigin(req);
+    const region =
+      req.headers.get("x-country") ?? getBrowserConfig("developmentUserRegion");
```
carlmw, line 18: Do we know if Netlify will set a value if it can't geolocate a user? I'd hate for onboarding to fail and lock out someone because their IP isn't mapped in the ip=>geo database. I know that Cloudflare set `XX` when it can't locate a user, perhaps Netlify do something similar?

In case they don't set a fallback value I don't think this should block onboarding. There are still likely to be things a user can do legitimately after signing in that don't require us knowing their region. Maybe this could be logged instead?

carlmw (commented):
LGTM, just worried about locking users out if they can't be geolocated.

src/app/api/auth/onboarding/route.ts
JBR90: I couldn't find out a lot about x-country, nfortunately we cant use the cloudflare one as is seems to be filtered out by netlify. logged instead? - do you mean report to bugsnag in an Oak Error but not thrown?

carlmw: Yeah reported to BugSnag with the user id but handled, would help us get an idea of how many user's have a missing region while not blowing up their onboarding.

JBR90: I've not managed to get the errorReporter to work in the route file but no longer locks user in.

rightsaidjames (approved):
Looks good. I couldn't find a way to test what happens when your region can't be identified (anyone fancy a trip to the Arctic Circle?), but other than that I've tested all permutations I can think of. Once the region has been set, it won't be overwritten for that user, even if you repeat onboarding, unless you clear it out of the public data first.

carlmw (approved):
:shipit:

PR #2723, opened 2024-08-21
Update `react-aria`

orangemug:
## Description
This is a big update, see changes https://github.com/adobe/react-spectrum/compare/react-aria%403.33.1...react-aria%403.34.3

We should get QAs to smoke test this prior to merge


orangemug:
@jack-skerrett-oak @rightsaidjames no rush on this, but we'd need a smoke test.

Unfortunately I can't find a changelog or what's actually changed just a massive diff. This will likely improve experience of a11y users by fixing some edge-case-y issues

kimon-satan (approved):

rightsaidjames (approved):
I've done some regression testing on the ARIA features added to different parts of the site, as well as a general smoke test. All looks good.

PR #2727, opened 2024-08-21
fix: add a not found page to catch 404s and redirect to 404 page

njblore:
## Description

Music year: {owa_music_year}

- We have added pages for auth in the app router, and this takes precedence over the pages router, so all 404s are currently using the app router's default `not-found` page without any oak theming
- Creating the 404 page in the app router requires a bit more work, so this adds a little hack to redirect to the existing themed 404 page we have in pages router, until we can get that work done
- A tech debt ticket will be added to upgrade the `not-found` and remove the 404 page

## How to test

1. Go to {owa_deployment_url} anywhere
2. Update the url with an invalid slug eg {owa_deployment_url}/teachers/lessons/not-a-real-lesson
3. You should see the oak 404 page

## Screenshots

Screenshot 2024-08-21 at 13 21 27

How it should now look:
Screenshot 2024-08-21 at 13 20 49

## Checklist

- [ ] Oak themed 404 page should be shown when an invalid url is accessed
- [ ] the 'back' button should go back to the last valid owa url (if applicable)
- [ ] the 'home' button should take you home

### DIFF
--- a/src/app/not-found.tsx
+++ b/src/app/not-found.tsx
@@ -1,0 +1,7 @@
+"use client";
+import { useRouter } from "next/navigation";
+
+export default function NotFound() {
+  const router = useRouter();
+  return router.replace("/404");
+}


k-huggs (approved):

JBR90 (approved):

jack-skerrett-oak (approved):
Being redirected to 404s whenever the URL is invalid.

PR #2731, opened 2024-08-21
feat: add auth to download requests

carlmw:
## Description

- Adds "X-Should-Authenticate-Download" header - which is set to  `true` if feature flag is enabled.
- Adds auth token from clerks `'.useAuth()'` - are a they a teacher / geo authenticated metadata
- Updates feature flag clerk.

## How to test

Downloads should work as they did before on auth-downloads and downloads pages.

`teachers/lessons/under-attack-70rpcr` has geo restrictions set to` true` in the downloads api for testing purposes, this will only be checked if the feature flag is on.

 **Test Case 1: Feature Flag Off**

- [x] Download a file with the feature flag turned off. It should download as expected.

**Test Case 2: Feature Flag On, Logged In geo authorised****

- [x] Download a file with the feature flag turned on and the user logged in. Verify that the user metadata in Clerk has isRegionAuthorised set to true. The downloads should work 

**Test Case 3: Feature Flag On, Logged In not geo authorised****

- [x] Download a file with the feature flag turned on and the user logged in. Verify that the user metadata in Clerk has isRegionAuthorised set to false. The downloads should not work  statusCode = 403;

deploy preview: [https://deploy-preview-2731--oak-web-application.netlify.app/](https://deploy-preview-2731--oak-web-application.netlify.app/)


src/components/TeacherComponents/hooks/downloadAndShareHooks/useResourceFormSubmit.ts
```diff
@@ -48,10 +51,12 @@ const useResourceFormSubmit = (props: UseResourceFormProps) => {
       setTermsInLocalStorage(terms);
     }
     if (props.type === "download") {
+      const accessToken = await auth.getToken();
```
carlmw, line 54: Will be `null` if there is no logged in user.

src/components/SharedComponents/helpers/downloadAndShareHelpers/createDownloadResourcesLink.tsx
```diff
@@ -34,7 +35,16 @@ const createDownloadResourcesLink = async (
     selection,
     isLegacyDownload,
   };
-  const res = await fetch(downloadEnpoint);
+  const res = await fetch(
+    downloadEnpoint,
+    authToken
```
carlmw, line 40: The token has been threaded through to the underlying request to the curriculum downloads API

src/components/TeacherComponents/hooks/downloadAndShareHooks/useResourceFormSubmit.test.ts
```diff
@@ -128,6 +128,8 @@ describe("useResourceFormSubmit", () => {
         "lesson",
         ["intro-quiz-questions"],
         true,
```
JBR90, line 130: We should maybe refactor this to take an object, so it's clear that the values are.

rightsaidjames:
@carlmw / @JBR90 Is this testable on OWA yet? If not, I don't think it's worth my time testing it at this point, I (or the next QA) can look at it when it is properly integrated.

RBrownsett:
@carlmw @JBR90 when I signed up using this PR I noticed the meta data for the user I had was missing the region property in the public section- is that expected? Example meta data:
image

JBR90:
@RBrownsett @ @rightsaidjames - Sorry for the radio silence on this one, the curriculum downloads PR has been merged and i've updated the comments in this PR on how to test with a hard coded geo restricted lesson. Let me know if there are any further questions>

@RBrownsett - Yes this is expected , we are storing actual region in private meta data and isRegionAuthorised in public.

RBrownsett:
@JBR90 as mentioned test cases 1 and 2 have both passed.

Test case 3 I am still able to download resources with this account below when using a VPN set to be outside of the UK (I have cleared my cache and deleted my clerk jwt tokens) 

image

Unless there is another cookie that needs to be cleared from this list? 

image

RBrownsett (approved):
Tested all 3 scenarios, worked as expected.

Test case 3 needed to be a specific lesson to ensure 403 was displayed 

image

PR #2732, opened 2024-08-22
Feature branch: teacher squad events upgrade

njblore:
## Description

Feature branch containing updates to all teacher squad's events to use the new format
- Browse events
- Lesson events
- Search events

## How to test

1. Go to https://deploy-preview-2732--oak-web-application.netlify.thenational.academy
2. Trigger events below and view them in the console

## Checklist
- [ ]  Add new `browseRefinedAccessed` event
    - triggered from keystage selection on the homepage
- [ ] Update events to use `browseRefined`
    - keystage selection keypad on subject listing pge
    - subject selection on subject listing page
    - programme selection on programme listing page
    - thread filter selection on unit listing page
- [ ]  Update fields of unit accessed
    - unit selection on unit listing page
- [ ] Update fields of `lessonResourcesDownloaded`
    - triggered from download button on downloads page
- [ ] Update fields of `lessonShared`
    - triggered from share buttons on lesson share page
- [ ] Update fields of `lessonResourceDownloadStarted`
    - triggered from lesson overview page download buttons
- [ ]  Replace lessonSelected event with `lessonAccessed`
    - triggered from lesson listing page lesson selection
- [ ] Add new `searchAccessed` event
	- triggered when search button clicked
- [ ] Update fields of `searchRefined`
	- triggered from search page when using filters
- [ ] Update `searchResultsExpanded`
	- triggered from search page when using drop down on search results
- [ ] Removes `displaySearchResults`


JBR90 (approved):

jack-skerrett-oak (changes requested):
Sorry, there's a lot to get through so have only gotten through the browse/lesson things. I will pick up the remaining search ones tomorrow. I have also included places where I feel events should be triggered from but currently aren't. I know we said otherwise in slack but given there is potentially a fair bit of rework, it might make sense.

- [x] When I select a key stage on the homepage I see a “Browse Refined” event, and not a “Browse refined accessed” event (though I have just seen a comment on a ticket about this)
- [x] browseRefinedAccessed events should also be triggered when selecting from the footer or the hamburger menu
- [x] Browse refined on the programme listing pages do not work properly
    - They only take into consideration the tier selected
    - If you select an exam board programme (ks4 history) then no browse refined event is sent
    - If you select an tier + exam board programme (ks4 science) then a browse refined event is sent, but it is only containing information on the tier selected
    - If you select a tier programme (ks4 maths) then the event is fine and wonderful
- [ ] On UnitAccessed, posthog complains of a type mismatch for yearGroupName and yearGroupSlug. They have been defined as numeric but the values provided are strings
- [x] On UnitAccessed, when you select a unit from within a programme that has programme factors (tier, exam board, tier & exam board, then the “tierName” and “examBoard” values are always returned as “null”
- [x] On LessonAccessed, posthog complains of a type mismatch for yearGroupName and yearGroupSlug. They have been defined as numeric but the values provided are strings
- [x] On lessonShared, I think the engagementType should be “advocate”. https://www.notion.so/oaknationalacademy/Product-Squad-Technical-Requirements-Summer-2024-60f931984867449a90d6f6f5d46a7308?pvs=4#c87e2e12cbb248fd8cf465749c26fdf9

There are no browse/access events anywhere within specialist, will have to talk to Ewan about that one. Also, lessonShareStarted hasn’t been updated, another one for Ewan.

jack-skerrett-oak (changes requested):
OK, I think this is the rest of it

- [x] We might want to double check that searchAccessed is “explore”, reading the descriptions I feel it should be refine still https://www.notion.so/oaknationalacademy/Product-Squad-Technical-Requirements-Summer-2024-60f931984867449a90d6f6f5d46a7308?pvs=4#0b4632ee6a704e46b41b6061cdec0040
- [ ] Because of the way that search works, whenever you refine a search, a searchAccessed event is also triggered
- [ ] When you apply a filter in search and get a searchRefined event, the component associated with the event is “keystage_keypad_button” but I think it should be “filter_link” (that’s the best option from the list given)
- [ ] It feels like the searchRefined event is triggered at the wrong time. If I apply my first filter, the refined event says the correct filter value but then says I have no active filters and the search result count is the same as before I added the filter. When I add another filter, the active filter now contains the previously added filter (good) and the search results are how many were returned by the addition of the previous filter. It seems to be lagging behind, I don’t know how valuable that will be. What I imagine would be more useful is to see “this filter was applied and now there are this many search results” rather than “this filter was applied and this is how many results there were last time)
- [ ] For searchResultsExpanded, when I click on a unit result (search for surds or something) it populates the lesson name with “Surds”, which it shouldn’t as I have selected a unit. The lesson slug is an empty string. Should these be nullable?
- [x] searchResultsExpanded has been given the engagementIntent “Use”, it doesn’t seem relevant when you look at the list in notion, should probably be refine https://www.notion.so/oaknationalacademy/Product-Squad-Technical-Requirements-Summer-2024-60f931984867449a90d6f6f5d46a7308?pvs=4#6025daa18e3c4f618b4f0059e8f0c8e0
- [ ] searchResultsExpanded has the component type “select_oak_lesson” when you click to expand a unit search result, should probably be something like “unit_card” but might be worth checking
- [ ] searchResultsExpanded has the component type “select_oak_lesson” when you click to expand a lesson search resut, should probably be something like “lesson_card” but might be worth checking

Another for Ewan, Search Journey Initiated exists but it hasn’t been updated on the Avo branch, I assume we don’t want it.

jack-skerrett-oak:
> > It feels like the searchRefined event is triggered at the wrong time. If I apply my first filter, the refined event says the correct filter value but then says I have no active filters and the search result count is the same as before I added the filter. When I add another filter, the active filter now contains the previously added filter (good) and the search results are how many were returned by the addition of the previous filter. It seems to be lagging behind, I don’t know how valuable that will be. What I imagine would be more useful is to see “this filter was applied and now there are this many search results” rather than “this filter was applied and this is how many results there were last time)
> 
> @jack-skerrett-oak on this point it's tricky because the page reloads when a filter is applied, so we can either send the event right as the button is clicked and capture the triggering filter in the event, or we can send the event after the page reloads and send along the updated result count etc, but we won't know what filter was clicked to trigger this. We could do some extra work to encode the last applied filter in the URL or something, but then there are still complexities with filters being removed rather than applied and how we capture that. What do you think?

@njblore yeah I didn't think that would be simple. I guess it is something that we can feed back to Ewan and leave as is for now? There is mounting pressure on making search better, especially around the whole page loading of results (first a11y things, now events)

jack-skerrett-oak (changes requested):
Not over yet! :D 

- [x] The unit name field in searchResultExpanded when opening a lesson result has the lesson name, rather than the unit name (the unit slug is fine)
- [ ] Then possibly a harder one to fix... It seems that if you expand a search result and select an option pretty quickly, the URL the event is sent from is the page you'd land on, not the search page (search result opened isn't all that important because that might not exist soon anyway)
Screenshot 2024-08-30 at 10 31 38

njblore:
@jack-skerrett-oak I'm struggling to find a reasonable way to fix the second issue, the URL that is sent is not something we control ourselves and it looks like in our analytics set up there is a slight delay between the event being sent to avo and sent to posthog, which allows the url to change before it is captured by posthog. How much of an issue do you think this is?

jack-skerrett-oak:
> @jack-skerrett-oak I'm struggling to find a reasonable way to fix the second issue, the URL that is sent is not something we control ourselves and it looks like in our analytics set up there is a slight delay between the event being sent to avo and sent to posthog, which allows the url to change before it is captured by posthog. How much of an issue do you think this is?

@njblore I'll raise it in the channel :)

jack-skerrett-oak (approved):
Given the outstanding issue pre-exists this ticket I don't see a reason to block the PR based on that.

As part of re-testing I didn't fully check through only events, I only focused on the areas that needed fixing due to time constraints.

PR #2737, opened 2024-08-23
LESQ-980: Unauthorised download notice

carlmw:
## Description

Music year: {owa_music_year}

- List of changes

## Issue(s)

Fixes #

## How to test

1. Go to {owa_deployment_url}
2. Click on \_\_\_\_\_\_

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist

- [ ] Design sign-off


src/__tests__/__helpers__/mockClerk.ts
```diff
@@ -0,0 +1,45 @@
...
+  spy?.mockRestore();
+  spy = null;
+}
+
+export function setUseUserReturn(newUseUserReturn: UseUserReturn) {
+  if (!spy) {
+    enableMockClerk();
+  }
+  useUserReturn = newUseUserReturn;
+}
```
carlmw, line 45: I was mocking Clerk a lot so it was time to make it more convenient. This will also make it a little easier to rewrite tests when we remove the feature flag.

clerk.d.ts
```diff
@@ -31,6 +23,21 @@ declare global {
        * Indicates that the user has been onboarded
        */
       isOnboarded?: boolean;
+      /**
+       * Indicates that the user is authorised to download region restricted content
+       */
+      isRegionAuthorised?: boolean;
```
carlmw, line 29: Is this a good name for the thing?

src/components/TeacherViews/LessonDownloads.view.tsx
```diff
@@ -288,74 +293,87 @@ export function LessonDownloads(props: LessonDownloadsProps) {
...
+                lessonTitle={lessonTitle}
+                unitSlug={unitSlug}
+                unitTitle={unitTitle}
+                programmeSlug={programmeSlug}
+                data-testid="downloads-confirmation"
+                isCanonical={props.isCanonical}
+                nextLessons={lesson.nextLessons}
+                onwardContentSelected={onwardContentSelected}
+                isSpecialist={isSpecialist}
               />
```
carlmw, line 318: I was initially going to create a whole other page, but bringing in the breadcrumbs and everything else would have made this a lot more work. Adding a third state to the downloads page made a lot more sense.

src/node-lib/curriculum-api-2023/queries/lessonDownloads/lessonDownloadsCanonical.schema.ts
```diff
@@ -11,6 +11,8 @@ export const lessonDownloadsCanonicalSchema = baseLessonDownloadsSchema.extend({
 
 export type LessonDownloadsCanonical = z.infer<
   typeof lessonDownloadsCanonicalSchema
->;
+> & {
+  isDownloadRegionRestricted: boolean;
+};
```
carlmw, line 16: Once this is coming from the query all these unions can go away.

src/node-lib/curriculum-api-2023/queries/lessonDownloads/lessonDownloads.query.ts
```diff
@@ -126,12 +128,19 @@ const lessonDownloadsQuery =
...
+      lessonData = lessonDownloadsSchema.parse({
         ...lessonDownloads,
         isLegacy: false,
         isSpecialist: false,
       }) as T;
     }
+
+    return {
+      ...lessonData,
+      isDownloadRegionRestricted: ["account-security-68rkee"].includes(
```
carlmw, line 140: Hard coding a single lesson for now so that this can be tested.

JBR90 (approved):
Can see the message as expected, but auth downloads is just a blank page when feature flag is off and in incognito mode.

rightsaidjames:
Just a thought, but would it be better if the user's region was checked on the fly against the list of approved regions (and using the outcome of this check to determine what they can access), rather than setting the `isRegionAuthorised` flag at onboarding (and using this value to determine what they can access)?

rightsaidjames:
Also, it doesn't really make sense to me that `/downloads` can be reached while logged out (or with the FF not enabled), but whilst logged in I can only access it if I'm in an authorised region. Likewise, with `/downloads-auth` I would expect it to be the regular downloads page (with no auth required) when the FF is not enabled rather than a blank page. Why isn't there a single `/downloads` page that behaves differently depending on the FF state? Or is all that to come in a later PR?

njblore:
@rightsaidjames Carl's away so I'll try and answer these for you!

For the first point it's definitely a reasonable thought, and that's how we set it up initially. However there were concerns that we'll need to check against the whitelisted regions in multiple places, so this solution is intended to avoid us having to maintain multiple versions of the whitelist, or find a reasonable way to share it, and instead just have a single whitelist in one place at onboarding - access checks are also then simplified to checking a single field for true/false.

On your second point - yes once we promote this feature we will consolidate to a single downloads page again, having multiple pages is just a necessity right now due to how the combination of auth & FF works. Agree it would be good if behaviour was consistent, but that's not a high priority right now as long as each page works as expected (and if it's unclear what is 'expected' we can iron that out!)

rightsaidjames (approved):
That all sounds reasonable!

For the first point, my main concern with the `isRegionAuthorised` flag is that if a specific region was added or removed from the list of authorised regions then there would need to be some sort of migration/script to change the value for those users? e.g. if Jersey was removed from the list of allowed regions, we would need to search for all users with a `region` of `JE`, and then set their `isAuthorisedRegion` flag to `false`. Probably not an issue in the short to medium term, but who knows?

Anyway, I'm happy that this is working as expected in its current state - LGTM!

jack-skerrett-oak:
> Just a thought, but would it be better if the user's region was checked on the fly against the list of approved regions (and using the outcome of this check to determine what they can access), rather than setting the `isRegionAuthorised` flag at onboarding (and using this value to determine what they can access)?

@rightsaidjames There is also a more "product-y" answer to this that is definitely going to be a majority of users. What we care about on the geo-restriction side of things is not that a user is in the UK when they access the content, but more that they are a teacher in the UK. Their exact location when accessing the content is much less important to us. So in the tiny number of scenarios where a UK teacher is trying to access our content from outside the UK, we don't want them blocked (though they will of course be blocked if they are signing up for the first time outside of the UK).

PR #2743, opened 2024-08-29
Update MVs which filter out cycle 2 programmes

orangemug:
## Description
Update the curriculum MVs which filter out cycle 2 programmes from curriculum visualiser until ready for launch.

## Issue(s)

Fixes `CUR-788`

## How to test
Test that there are no functional changes to the visualiser. Also @rightsaidjames are you able to find out if we can publish a programme on staging double check it all works correctly prior to production publish?

FYI: I can setup, a test env for staging locally and proxy it over to you.


sonalisin (approved):

rightsaidjames (approved):
Based on our testing in the preview environment and Jamie's ngrok environment that's pointing at staging data (with a D&T programme published), I think we're good to go. Cycle 1 subjects are unaffected, and when pointed at staging with published cycle 2 data there are no cycle 2 lots displayed. So good to go!

PR #2749, opened 2024-09-03
fix: enable local development in chrome [LESQ-1052]

njblore:
## Description

Music year: 1983

[Ticket](https://www.notion.so/oaknationalacademy/I-want-to-develop-locally-on-Chrome-d26e3f1f31c94c5c8b89605f3f86e584?pvs=4)
- Adds reverse proxying to posthog calls when developing locally - this gets around the cors block that happens otherwise on localhost, preventing interaction with feature flags (see https://posthog.com/docs/advanced/proxy/nextjs) 
- Will only add this on dev mode, production builds should still use the current posthog set up

## Issue(s)

Fixes #
Unable to develop feature flagged features locally in chrome

## How to test

This should only affect local development, it should not affect how events are reported to posthog or how feature flags work

1. Go to https://deploy-preview-2749--oak-web-application.netlify.thenational.academy
2. Test some events are still working as expected

For devs:
1. pull down the branch and run on chrome
2. see that you can access feature flagged downloads-auth page


k-huggs (approved):
Seems to work fine on chromium browser!

JBR90 (approved):

jack-skerrett-oak (commented):
When I add a number of search filters one after another, I am only getting one search refined event. When I do the same on prod I am getting individual search refined events (though their bodies might be wrong). Happy to quickly pair to see if the issue is relevant, else I will raise this problem else where.

jack-skerrett-oak:
> When I add a number of search filters one after another, I am only getting one search refined event. When I do the same on prod I am getting individual search refined events (though their bodies might be wrong). Happy to quickly pair to see if the issue is relevant, else I will raise this problem else where.

You can see in this screenshot I only have two search refined events, even though I quickly added four, and then after 30 seconds or so added the type filter
Screenshot 2024-09-03 at 11 44 06

njblore:
@jack-skerrett-oak seems to be working as expected for me, although I do have to refresh in posthog a lot and they're sometimes slow to come through. Are you clicking them at super speed? or just regular?  
Screenshot 2024-09-03 at 12 02 49

jack-skerrett-oak (approved):
My issue is just an issue in general so we can ignore it :D

PR #2777, opened 2024-09-12
chore: add some notes to server config about some values needing to be set in Netlify

carlmw:
## Description

Music year: 1954

- List of changes

## Issue(s)

Fixes #

## How to test

1. Go to {owa_deployment_url}
2. Click on \_\_\_\_\_\_

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist

- [ ] Design sign-off


src/node-lib/getServerConfig.ts
```diff
@@ -193,13 +193,17 @@ const envVars = satisfies<Record<string, EnvVar>>()({
     default: null,
   },
   clerkPublishableKey: {
+    description:
+      "Note: this value must also be set in Netlify's environment vars since it is needed in middleware",
```
carlmw, line 197: This is annoying but the oak config .env files are not sourced in middleware (which runs in deno) and the only way to set them is to follow Netlify conventions and set them in their UI.

netlify/edge-functions/subdomain-redirect.ts
```diff
@@ -63,9 +63,8 @@ async function redirectNetlifySubdomains(
   console.log(`URL is on allow list: ${isOnAllowList}`);
 
   if (subdomain && !redirected && !isOnAllowList) {
-    const redirectTargetUrl = new URL(
-      `https://${subdomain}.netlify.thenational.academy/`,
-    ).href;
+    const redirectTargetUrl = new URL(request.url);
+    redirectTargetUrl.host = `${subdomain}.netlify.thenational.academy`;
```
carlmw, line 67: Ok! This was the issue all along.

Netlify expect you to either let them manage your DNS or to [delegate a sub-domain for them to manage for deploy previews](https://docs.netlify.com/domains-https/custom-domains/automatic-deploy-subdomains/#:~:text=setup%20steps.-,Want%20to%20use%20a%20subdomain%20without%20bringing%20the%20apex%20domain%20to%20Netlify%20DNS%3F,-You%20can%20bring).

☝🏻 we're not doing that, instead we've rolled our own solution in Cloudflare of wildcard DNS and the subdomain-redirect edge function (above).

This means that internally Netlify is unaware of the `netlify.thenational.academy` sub-domain on deploy previews. Instead it treats all requests as being made to `netlify.app` — which is why the redirects inside the Clerk middleware were always made through `netlify.app`. This redirect would trigger the above subdomain-redirect edge function. 

This function however was severely limited! It would always redirect to the root of the app, discarding both the pathname and search params.

My change above alters the behaviour of the subdomain redirect function (which is only applied to deploy previews). It now only rewrite the host portion of the URL,  preserving the pathname and search params.

This means that deploy previews go through an additional redirect, but eventually settle on the proper domain and everything gets where it needs to go.

Ideally we'd do the proper Netlify way to stay on the rails — however that would be a chunk of ops work. 



dance3

src/middleware.ts
```diff
@@ -5,7 +5,8 @@ export default clerkMiddleware();
 export const config = {
   matcher: [
     // Skip Next.js internals and all static files, unless found in search params
-    // "/((?!_next|[^?]*\\.(?:html?|css|js(?!on)|jpe?g|webp|png|gif|svg|ttf|woff2?|ico|csv|docx?|xlsx?|zip|webmanifest)).*)",
-    "/api/auth/onboarding",
+    "/((?!_next|[^?]*\\.(?:html?|css|js(?!on)|jpe?g|webp|png|gif|svg|ttf|woff2?|ico|csv|docx?|xlsx?|zip|webmanifest)).*)",
+    // Always run for API routes
+    "/(api|trpc)(.*)",
```
carlmw, line 10: This now matches the stock/recommended Clerk config.

RBrownsett:
@carlmw when I follow the link, sign in I am still sent to the homepage. I've have also tried to navigate via the lesson mentioned and download, I still get re-directed to the homepage. Should I see the downloads page?

https://github.com/user-attachments/assets/dac3ee3c-b63a-4cf9-a3a1-a6336787d1be

carlmw:
> 3\. https://deploy-preview-2777--oak-web-application.netlify.thenational.academy/teachers/programmes/citizenship-secondary-ks3-l/units/citizenship-whats-it-all-about-e038/lessons/what-is-citizenship-c4t38d/downloads-auth?preselected=all

@RBrownsett 
Ahah! Yes. Looks like you're hitting this bug that @njblore is looking at https://www.notion.so/oaknationalacademy/I-want-to-be-redirected-to-onboarding-after-signing-up-through-a-direct-download-link-d8dc72cc35574b5381cb04bafc9bba75?pvs=4 which seems to be specifically related to the sign-in page.

I was hoping this change might fix that bug too but no such luck. Though now i've got a whiff of what that other issue could be.

This PR should work if you click the Google button on the sign-up page (without clicking the "Sign-in" link first). 

I'll take a look at the sign-in issue while Tash is away today because I might have the scent of the cause.

RBrownsett (approved):
Approved - Tested the following from the Sign up page:
Teacher flow - Continue button / Skip button
Non Teacher flow
User is re-direct to the download page for the lesson

netlify/edge-functions/subdomain-redirect.ts
JBR90: Nice one! Maybe worth flagging this in slack too ?

JBR90 (approved):

PR #2784, opened 2024-09-16
Feat/lesq 1046/threads styling [LESQ-1046]

njblore:
## Description

Music year: 1953

- Refactor unitsLearningThemeFilters to use radio buttons for threads instead of links
- Add a component RadioTile, will extract this to components as a separate ticket

## How to test

1. Go to https://deploy-preview-2784--oak-web-application.netlify.thenational.academy/teachers/programmes/chemistry-secondary-ks4-foundation-ocr/units
2. Click on Themes along the right
3. You should see the inputs toggle and units update

## Screenshots

Screenshot 2024-09-17 at 08 13 19

How it should now look:

Screenshot 2024-09-17 at 08 43 29

## Ticket

https://www.notion.so/oaknationalacademy/I-want-threads-to-be-restyled-as-radio-buttons-25be5006e6084a9a9986a261f8f9b54a?pvs=4


RBrownsett:
@njblore  I'm having a few issues when I try to move through the radio list using the keyboard? I'm able to select the top item in the list but when I press down the keyboard I'm not able to got to the next thread?

Monika-Koziol (changes requested):
Just a couple of small changes:

- [x] Can we change the copy to say "All' instead of "All in suggested order"
- [ ] Can we add 32px padding at the bottom to avoid threads touching the footer when there are few units
Screenshot 2024-09-17 at 10 02 42

src/components/TeacherComponents/UnitsLearningThemeFilters/RadioTile.tsx
```diff
@@ -0,0 +1,117 @@
...
+  onChange: (theme: RadioTheme) => void;
+  onFocus: (value: string | undefined) => void;
+}) => {
+  return (
+    <OakBox
+      $borderColor="border-neutral"
+      $ba="border-solid-m"
+      $borderRadius="border-radius-s"
+      $pa="inner-padding-s"
+      key={theme.slug}
```
carlmw, line 66: `key` is redundant here

src/components/TeacherComponents/UnitsLearningThemeFilters/UnitsLearningThemeFilters.tsx
```diff
@@ -91,28 +113,29 @@ const UnitsLearningThemeFilters = ({
...
+      >
+        {[
+          { slug: "all", label: "All in suggested order" },
+          ...learningThemesMapped,
+        ].map((theme) => {
+          const isChecked = activeThemeSlug === theme.slug;
+          const isFocussed = focussedThemeSlug === theme.slug;
+          return (
+            <RadioTile
+              theme={theme}
```
carlmw, line 129: this does need a `key` here

src/components/TeacherComponents/UnitsLearningThemeFilters/UnitsLearningThemeFilters.tsx
```diff
@@ -25,56 +26,77 @@ export type UnitsLearningThemeFiltersProps = {
...
+        eventVersion: "2.0.0",
+        analyticsUseCase: "Teacher",
+        filterType: "Learning theme filter",
+        filterValue: theme.label,
+        activeFilters: { keyStage: [keyStageSlug], subject: [subjectSlug] },
+      });
+    }
+
+    const query = theme.slug === "all" ? {} : { "learning-theme": theme.slug };
+    router.push(
```
carlmw, line 90: 🏄🏻 I think this should be a `replace` and not a `push`. I wouldn't expect to be taken back to the last filter results if I hit the back button, instead i'd expect to go to the previous page.

src/components/TeacherComponents/UnitsLearningThemeFilters/UnitsLearningThemeFilters.tsx
```diff
@@ -25,56 +26,77 @@ export type UnitsLearningThemeFiltersProps = {
...
-            } else if (b?.slug === "no-theme") {
-              return -1;
-            } else {
-              return 0;
-            }
-          },
-        )
+        .filter((theme) => isRadioTheme(theme))
+        .map((theme) => theme as RadioTheme)
+        .sort((a: RadioTheme, b: RadioTheme) => {
```
carlmw, line 50:
```suggestion
        .filter(isRadioTheme)
        .sort((a, b) => {
```

the extra `=>`, `map` and `sort` typing should be unnecessary as type info is supposed to be forwarded through `isRadioTheme`.

src/components/TeacherComponents/UnitsLearningThemeFilters/RadioTile.tsx
```diff
@@ -0,0 +1,117 @@
...
+const HiddenRadioButtonInput = styled.input.attrs({
+  type: "radio",
+})`
+  position: absolute;
+  opacity: 0;
+  cursor: pointer;
+`;
+
+const RadioTileFocus = styled(OakBox)<OakBoxProps>`
+  box-shadow: ${`inset 0 0 0 0.15rem #ffe555`};
```
carlmw, line 27: Is there a reason to use a template for the box-shadow here?

carlmw (approved):
LGTM, just some minor things

src/components/TeacherComponents/UnitsLearningThemeFilters/UnitsLearningThemeFilters.tsx
njblore: agreed!

src/components/TeacherComponents/UnitsLearningThemeFilters/UnitsLearningThemeFilters.tsx
njblore: you would think! this drives me mad, but it won't infer the type without an assertion here - actually though I think we can add that as the filter return type, which would be a bit neater: `.filter((theme): theme is RadioTheme => isRadioTheme(theme))`

src/components/TeacherComponents/UnitsLearningThemeFilters/RadioTile.tsx
njblore: nope! it's because I had originally tried to get the colour hex from the token name but that fn isn't exported from components library 🫠

src/components/TeacherComponents/UnitsLearningThemeFilters/UnitsLearningThemeFilters.tsx
carlmw: Weird. Works on my machine 🤣 

Screenshot 2024-09-17 at 11 42 49

njblore: now hang on, I've been tricked!
Screenshot 2024-09-17 at 11 45 23
Screenshot 2024-09-17 at 11 45 54

Monika-Koziol (approved):

PR #2785, opened 2024-09-16
feat(UnitsTab): add unit count feedback for screen readers

assadk88:
## Description

- Added voice over announcement when year group filters are selected

## Issue(s)

Fixes #

## How to test

1. Go to {owa_deployment_url}
2. Click on \_\_\_\_\_\_

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist

- [X] Design sign-off


assadk88:
@orangemug Sorry that this took longer than anticipated, but it took some digging into. What I've found was that the unit test wasn't working because `getNumberOfSelectedUnits` and `isVisibleUnit` were defined in the same module. So, when I tried to mock isVisibleUnit in the unit tests, it didn’t affect the internal call to `isVisibleUnit` inside `getNumberOfSelectedUnits`, which continued to use the original implementation. This was resolved by moving `isVisibleUnit` into a separate module.

src/utils/curriculum/getNumberOfSelectedUnits.ts
```diff
@@ -0,0 +1,26 @@
...
+  selectedYear: string | null,
+  yearSelection: YearSelection,
+): number {
+  let count = 0;
+
+  Object.keys(yearData).forEach((year) => {
+    const units = yearData[year]?.units;
+    if (units && (selectedYear === "" || selectedYear === year)) {
+      units.forEach((unit) => {
+        isVisibleUnit(yearSelection, year, unit) ? count++ : count;
```
orangemug, line 20: This is a little odd, hence the [sonar cloud warning](https://sonarcloud.io/project/issues?severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&types=BUG&pullRequest=2785&id=oaknational_owa&open=AZICpe558EdTur4Z4FPb)  

Clearer would be

```js
count = isVisibleUnit(yearSelection, year, unit) ? count+1 : count
```

src/utils/curriculum/isVisibleUnit.ts
```diff
@@ -0,0 +1,24 @@
+import { YearSelection } from "@/components/CurriculumComponents/CurriculumVisualiser";
+import { CurriculumUnitsTabData } from "@/node-lib/curriculum-api-2023";
+
+export function isVisibleUnit(
```
orangemug, line 4: Not strictly a part of this work, just an FYI we also need some tests of this function. You can leave that for now though

orangemug (approved):
One small issue to resolve (see comments), then over to @rightsaidjames for QA

rightsaidjames (changes requested):
Adding visually-hidden/sr-only text for screen readers is only half of the requirements, sorry if this was unclear in the ticket. 

The number of units displayed needs to be _announced_ to users whenever they choose a year group filter, by adding an `aria-live="polite` and `aria-atomic="true"` to each of these spans. Furthermore the spans need to be present on page load (empty is fine) so that the live regions can be 'registered' with the screen reader and monitored for changes.

References:
#2369
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-live
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-atomic

rightsaidjames:
So... adding the `aria-live` to the sr-only span has had the unintended effect of the unit count being read out twice - once as part of the label, and then an an announcement. The solution that me and @assadk88 agreed on was to move the sr-only text outside of the span, so it was no longer part of the label. This would help avoid confusion for speech control users and sighted screen reader users (e.g. text to speech users with dyslexia).

@assadk88 also noticed that having a live region for each label meant that VoiceOver would not re-announce the unit count if you went back to a year group that you'd previously selected - this makes sense, as the contents of each live region isn't really changing once it has been populated once.

I also spotted that the unit count is incorrect at KS4 for [Secondary Science](https://deploy-preview-2785--oak-web-application.netlify.thenational.academy/teachers/curriculum/science-secondary-aqa/units) and [Secondary Maths](https://deploy-preview-2785--oak-web-application.netlify.thenational.academy/teachers/curriculum/maths-secondary/units), it looks like it's counting the units before the final filtering step (by tier or child subject) has taken place.

src/utils/curriculum/isVisibleUnit.ts
assadk88: Added a placeholder test file for this and created a ticket in the backlog

src/utils/curriculum/getNumberOfSelectedUnits.ts
assadk88: To increment in ternary form requires the `++` operator I believe – as Sonar is flagging this I've reverted it back it to it's previous if statement form and used the `+=1` assignment

src/utils/curriculum/isVisibleUnit.test.ts
```diff
@@ -0,0 +1,10 @@
+// import { isVisibleUnit } from "./isVisibleUnit.test";
+
+// import { YearSelection } from "@/components/CurriculumComponents/CurriculumVisualiser";
+// import { CurriculumUnitsTabData } from "@/node-lib/curriculum-api-2023";
+
+// TODO: Add a unit test for isVisibleUnit utility function
+
+describe("isVisibleUnit", () => {
+  it.todo("This function needs to be tested");
```
orangemug, line 9: Can you add a ticket to notion for this, and mark it as tech debt and add a comment above this line.

assadk88: Done

src/components/CurriculumComponents/UnitsTab/UnitsTab.tsx
```diff
@@ -77,18 +85,34 @@ const UnitsTab: FC<UnitsTabProps> = ({ trackingData, formattedData }) => {
...
+    yearData,
+    selectedYear,
+    yearSelection,
+  );
+
+  // Screen reader announcements
+
+  const [unitAnnouncement, setUnitAnnouncement] = useState("");
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
assadk88: Good point, this was from an approach I was trying before to force the aria-live region to announce correctly, but it can now be removed

rightsaidjames (approved):
Looks good - it's working reliably with all of the various filters (year, child subject, tier, subject category) and now provides an accurate unit count (in line with what's displayed on screen). It also works nicely with the pre-existing thread highlighting functionality.

orangemug (approved):

PR #2795, opened 2024-09-18
Feat/lesq 1046/threads styling wrap up

njblore:
## Description

Music year: 2011

- Replace radio tile component with oakRadioTile from components library

## How to test

1. Go to https://deploy-preview-2795--oak-web-application.netlify.thenational.academy/teachers/programmes/english-primary-ks1/units
3. Click on thread filters on the right
4. You should see the list of units updating on selection
5. Navigating with a keyboard should apply filters as they are moved through, without reloading the page
6. Look and behaviour should match production


src/pages/teachers/programmes/[programmeSlug]/units.tsx
```diff
@@ -222,7 +222,9 @@ const UnitListingPage: NextPage<UnitListingPageProps> = ({
                     <UnitsLearningThemeFilters
                       labelledBy={learningThemesId}
                       learningThemes={learningThemes}
-                      selectedThemeSlug={themeSlug ? themeSlug : "all"}
+                      selectedThemeSlug={
+                        selectedThemeSlug ? selectedThemeSlug : "all"
+                      }
```
Comment (carlmw) [inline src/pages/teachers/programmes/[programmeSlug]/units.tsx:227]: 🏄🏻 

```suggestion
                      selectedThemeSlug={selectedThemeSlug ?? "all"}
```

src/components/TeacherComponents/UnitsLearningThemeFilters/UnitsLearningThemeFilters.tsx
```diff
@@ -114,23 +112,18 @@ const UnitsLearningThemeFilters = ({
...
-                id={`${theme.slug}-${idSuffix}`}
-              />
-            );
-          },
-        )}
+        {[{ id: "all", label: "All" }, ...themeTileItems].map((theme) => {
+          const isChecked = activeThemeSlug === theme.id;
+          return (
+            <OakRadioTile
+              tileItem={theme}
```
carlmw, line 119: :surfer: not something to action now, but we should be cautious about passing non-primitives as props (particularly to components in the library). Theoretically it will make components harder to optimise later should we need to since the value is more likely to change between renders and harder to test equality. In reality React is slow anyway so it's not the end of the world. I'd usually prefer to pass each value as an individual prop when there are only a couple of things.

carlmw (approved):
LGTM, just my preference for the nullish operator and a vague thing about complex props for thinking.

mikeritson-oak (changes requested):
I've taken a look at this change across browsers, for style/design and functionality. Chrome, Firefox and Edge all function as expected and match the production variant. In Safari, when the user navigates to a particular thread (with the up/down arrow keys) then the selection is given a solid black background but does not show the yellow highlight border, as the other browsers do.
image
This does match production and if it is acceptable, or can be picked up later, then this PR can be approved.

njblore:
thanks @mikeritson-oak I'll open a separate ticket for that issue as it will be a fix in the components library. Also apologies but there have been a lot of new changes merged into this since I've been off so re-requesting a review, lmk if you'd rather I pass it to another QA (not sure who is on teacher squad duties this sprint!)

mikeritson-oak (approved):
Rechecked now that further work from main has been merged into the preview branch. All still looks good. One thing that was raised previously and @njblore has [confirmed](https://github.com/oaknational/Oak-Web-Application/pull/2795#issuecomment-2397001109) that will be handled separately. 👍

PR #2797, opened 2024-09-18
Fix/lesq 1074/onboarding hubspot signup [LESQ-1074]

njblore:
## Description

Music year: 1976

- add a default value (false) for newsletter sign up on role selection page

## Issue(s)

Fixes #
Unable to proceed past onboarding unless you check newsletter sign up

## How to test

1. Go to https://deploy-preview-2797--oak-web-application.netlify.thenational.academy/onboarding _*while signed in*_
2. Click on 'No' to go to role selection page
3. You should be able to select a role and proceed without checking the newsletter sign up checkbox
4. You should also be able to check the newsletter sign up and see that value reflected [in the hubspot form submission](https://app-eu1.hubspot.com/submissions/24947711/form/50399960-674f-4e23-ad66-367fbdbcdbfb/submissions/)

## Screenshots
Screenshot 2024-09-18 at 15 45 11

## Checklist

- [ ] Can continue past onboarding without signing up to the newsletter
- [ ] Can still sign up to the newsletter

PR #2798, opened 2024-09-18
LESQ-1073: Remove restriction on manual school entry

carlmw:
## Description

Music year: 1971

Fixes #
1. Removes the minimum length requirement on the manually entered school name and address
2. Sets the manual school name, school address and school picker inputs as "required"
3. Fixes a regression in the role selection form caused by the above changes (this page should be retested 🙇🏻 )

## How to test

1. Go to https://deploy-preview-2798--oak-web-application.netlify.thenational.academy/onboarding
5. Select "Yes" when asked if you "Work in a school"
6. You should see a required field for the school picker
7. Click "Enter manually"
8. You should see required fields for "School name" and "School address"
9. You should be able to a enter a non-whitespace only string into both and proceed

## Screenshots

How it used to look

Screenshot 2024-09-19 at 09 40 18

How it should now look:

Screenshot 2024-09-19 at 09 41 10

## Checklist

- [ ] Design sign-off

### DIFF
--- a/src/components/TeacherComponents/ManualEntrySchoolDetails/ManualEntrySchoolDetails.tsx
+++ b/src/components/TeacherComponents/ManualEntrySchoolDetails/ManualEntrySchoolDetails.tsx
@@ -43,9 +43,7 @@ const ManualEntrySchoolDetails: FC<ManualEntrySchoolDetailsProps> = ({
         render={({ field: { onChange, onBlur, value } }) => {
           const onChangeHandler = (e: ChangeEvent<HTMLInputElement>) => {
             onChange(e.target.value);
-            if (e.target.value.length > 2) {
-              onManualSchoolInputChange(true, e.target.value);
-            }
+            onManualSchoolInputChange(true, e.target.value);
           };
           const onBlurHandler = () => {
             onBlur();
@@ -56,8 +54,9 @@ const ManualEntrySchoolDetails: FC<ManualEntrySchoolDetailsProps> = ({
             <Input
               label="School name"
               placeholder="Type school name"
-              value={value}
+              value={value ?? ""}
               isRequired
+              required
               id={"school-name"}
               onBlur={onBlurHandler}
               onChange={onChangeHandler}
@@ -78,9 +77,7 @@ const ManualEntrySchoolDetails: FC<ManualEntrySchoolDetailsProps> = ({
         render={({ field: { onChange, onBlur, value } }) => {
           const onChangeHandler = (e: ChangeEvent<HTMLInputElement>) => {
             onChange(e.target.value);
-            if (e.target.value.length > 2) {
-              onManualSchoolInputChange(false, e.target.value);
-            }
+            onManualSchoolInputChange(false, e.target.value);
           };
 
           const onBlurHandler = () => {
@@ -92,8 +89,9 @@ const ManualEntrySchoolDetails: FC<ManualEntrySchoolDetailsProps> = ({
             <Input
               label="School address"
               placeholder="Type school address"
-              value={value}
+              value={value ?? ""}
               isRequired
+              required
               id={"school-address"}
               onBlur={onBlurHandler}
               onChange={onChangeHandler}
--- a/src/components/TeacherComponents/OnboardingForm/OnboardingForm.schema.ts
+++ b/src/components/TeacherComponents/OnboardingForm/OnboardingForm.schema.ts
@@ -32,13 +32,8 @@ const ukSchoolSchema = z.object({
 export type UkSchoolFormValues = z.infer<typeof ukSchoolSchema>;
 
 const manualSchoolSchema = z.object({
-  manualSchoolName: z
-    .string()
-    .min(3, "School name must be at least 3 characters long"),
-
-  schoolAddress: z
-    .string()
-    .min(3, "School address must be at least 3 characters long"),
+  manualSchoolName: z.string().trim().min(1),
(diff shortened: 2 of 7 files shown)


src/components/TeacherComponents/ManualEntrySchoolDetails/ManualEntrySchoolDetails.tsx
```diff
@@ -56,8 +54,9 @@ const ManualEntrySchoolDetails: FC<ManualEntrySchoolDetailsProps> = ({
             <Input
               label="School name"
               placeholder="Type school name"
-              value={value}
+              value={value ?? ""}
```
carlmw, line 57: Prevents the input from moving between a controlled and uncontrolled state when initialised without a `value`

src/components/TeacherViews/Onboarding/SchoolSelection/SchoolSelection.test.tsx
```diff
@@ -89,39 +89,39 @@ describe("Onboarding view", () => {
         name: "Enter manually",
       });
 
-      userEvent.click(manualButton);
+      await userEvent.click(manualButton);
```
carlmw, line 92: Fixing a bunch of `act` errors in the test runner

src/components/TeacherViews/Onboarding/RoleSelection/RoleSelection.view.tsx
```diff
@@ -61,17 +63,6 @@ const RoleSelectionView = () => {
...
-          if (getValues().role === "Other" && !getValues().other) {
-            setError("other", {
-              message: "Please tell us what your role is",
-            });
-          } else if (!getValues().role) {
-            setError("role", {
-              message: "Please select what describes you best",
-            });
-          }
-        }}
```
carlmw, line 74: This started to break when I added `noValidate` to the form — I needed this to disable the default browser validation which was being triggered by me adding the `required` attribute to a few inputs.

It seemed to cause a race between the `onSubmit` and `useForm` validation where the validation logic here was being immediately replaced by the rules defined in the Zod schema. This only appeared in tests which i'm putting down to the synthetic nature of the event system/some quirk I don't really understand.

Anyway moving this into a `refine` on the Zod schema gives us the same behaviour avoiding the issue entirely.

src/components/TeacherViews/Onboarding/RoleSelection/RoleSelection.view.tsx
```diff
@@ -34,13 +34,15 @@ const RoleSelectionView = () => {
...
-    setError,
     getValues,
     control,
     trigger,
   } = useForm<RoleSelectFormProps>({
     resolver: zodResolver(roleSelectFormSchema),
     mode: "onBlur",
+    defaultValues: {
+      newsletterSignUp: false,
+    },
```
carlmw, line 45: I needed https://github.com/oaknational/Oak-Web-Application/pull/2797 for the `refine` validation to run (it was bailing early in the tests since `newsletterSignUp` was always invalid)

src/components/TeacherComponents/OnboardingForm/OnboardingForm.schema.ts
```diff
@@ -4,15 +4,17 @@ const baseSchema = z.object({
...
-  }),
-  other: z.string().optional(),
-});
+export const roleSelectFormSchema = z
+  .object({
+    ...baseSchema.shape,
+    role: z.string({ message: "Please select what describes you best" }),
+    other: z.string().trim().optional(),
+  })
+  .refine((input) => input.role === "Other" && !!input.other, {
```
njblore, line 13: oh this is neat! didn't know zod could do this

njblore (approved):

src/components/TeacherComponents/OnboardingForm/OnboardingForm.schema.ts
carlmw: Was new to me too! Though I wish this sort of inter relationship could be expressed declaratively and a more representative type inferred as a result. Right now the type doesn't respect the rules in the refine so `other` will always be optional.

I've never used it but [Yup has `when` ](https://github.com/jquense/yup?tab=readme-ov-file#schemawhenkeys-string--string-builder-object--values-any-schema--schema-schema) which is a bit more declarative and does the same job — though i'm not sure if it will infer a better type!

mikeritson-oak (commented):
I've checked this over, all looked OK but I noted on thing that we may or may not want to mitigate.

Both fields allow any combination of user input of alphnumeric characters, without a minimum restriction and whitespace alone not permitted (obviously spaces can be used within entries). It is though possible to prepend whitespace to any entered value, so having spaces at the  start of a field value for instance:

image

If we are happy that this can occur, based on free format user entry, then this PR can be approved.

carlmw:
> I've checked this over, all looked OK but I noted on thing that we may or may not want to mitigate.
> 
> Both fields allow any combination of user input of alphnumeric characters, without a minimum restriction and whitespace alone not permitted (obviously spaces can be used within entries). It is though possible to prepend whitespace to any entered value, so having spaces at the start of a field value for instance:
> 
> image
> If we are happy that this can occur, based on free format user entry, then this PR can be approved.

I think that is okay and IMO preferable. I hate to mess with user's keystrokes (it's weird if you try to type a space and nothing happens). Leading and trailing white-space is trimmed on submission so we won't actually store it anyway.

mikeritson-oak (approved):
All looked good during testing. 👍

PR #2803, opened 2024-09-19
Added "why this why now" and "description" to unit modal

orangemug:
## Description
Added unit "description" and "why this why now" for cycle 2 subjects into the UI 

## Issue(s)

Fixes `CUR-832`/`CUR-831`

## How to test
For cycle 1:

 1. Head to curriculum visualiser
 2. Check prior/future unit description is same as previous
 
Enable cycle 2 units

 1. Check "description" and "why this why now" is present.

## Screenshots
Cycle 1
Screenshot 2024-09-19 at 15 29 07

Cycle 2
Screenshot 2024-09-19 at 15 28 48

### DIFF
--- a/CHANGE_LOG.md
+++ b/CHANGE_LOG.md
@@ -1,3 +1,61 @@
+# [1.567.0](https://github.com/oaknational/Oak-Web-Application/compare/v1.566.0...v1.567.0) (2024-09-26)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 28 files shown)


assadk88 (approved):

rightsaidjames (approved):
Looks good. Tested in a range of cycle 2 subjects to ensure it's behaving as expected, including with units with options. I would assume that if the unit description and WTWN _were_* to vary based on optionality or exam board variant, then this would be reflected in the visualiser, but as far as I can tell no one has actually done this yet. One to keep an eye out for!

I also confirmed that, with the cycle 2 flag enabled, cycle 1 unit modals still display prior/future unit info and not WTWN + unit description.

PR #2804, opened 2024-09-19
Make order "nullable" and don't include cycle 1 in cycle 2 phase options

orangemug:
## Description

 - When cycle 2 is enabled exclude cycle 1 phase options (cycle 2 includes all of cycle 1)
 - Add `.default(0)` hack to `curriculumUnits.schema.ts` as order is sometime `null` (to be investiagted)
 
**Note**: The hack allows us to check over subjects/phases that were previously erroring (order of units will be incorrect though) 

## Issue(s)
Fixes `CUR-921`/`CUR-922`

## How to test

 - Check for regressions (cycle 1)
 - Check for correct subjects in picker (cycle 2)
 - Check you can access all subjects/phases via lot picker, even though there might be issues in the visualiser itself (cycle 2)


sonalisin (approved):
LGTM 🦖

rightsaidjames (approved):
Looking good, both in terms of regression testing on the preview environment and with cycle 2 enabled on my local environment.

- [X] No duplicate subjects in lot picker
- [X] Primary Music loads
- [X] Cooking and nutrition (both phases) loads

PR #2834, opened 2024-09-30
LESQ-1096: fix confirmation email sent twice

carlmw:
## Description

Music year: 1988

rendering two sign-up/sign-in forms was causing the email confirmation
to be sent twice. I believe this was intended to make it easier to
present an alternative mobile view of the sign-up UI.

it appears to ultimately be an implementation bug in Clerk, because this
only happens with the confirmation link flow and not with the confirmation
code flow. Also if you refresh the page it will send another link which
is not great.

## How to test

1. Go to https://deploy-preview-2834--oak-web-application.netlify.thenational.academy/sign-up
2. Sign up with an email address
3. You should see the "Verify your email" screen
4. Check your email
5. You should have received one confirmation email

🚨 **I had to make some tweaks to the way the responsive layout was implemented to fix this. Definitely worth checking that I haven't caused a regression in the page layout (I tested and it all looked correct)** 

🤦🏻‍♂️ 😮‍💨  **Reloading the "Verify your email" screen will also cause another email to be sent. This appears to be a bug in Clerk because these emails should be throttled to the countdown you can see on screen.**

**Also, it will continue to send 2 emails in development owing to strict mode which causes the component to double render. In fact I was getting 4 emails in development because of strict mode and the double form bug combined. Fun.**

## Checklist

- [ ] Design sign-off

### DIFF
--- a/src/app/styles-registry.tsx
+++ b/src/app/styles-registry.tsx
@@ -1,0 +1,29 @@
+"use client";
+
+import React, { useState } from "react";
+import { useServerInsertedHTML } from "next/navigation";
+import { ServerStyleSheet, StyleSheetManager } from "styled-components";
+
+export default function StyledComponentsRegistry({
+  children,
+}: {
+  children: React.ReactNode;
+}) {
+  // Only create stylesheet once with lazy initial state
+  // x-ref: https://reactjs.org/docs/hooks-reference.html#lazy-initial-state
+  const [styledComponentsStyleSheet] = useState(() => new ServerStyleSheet());
+
+  useServerInsertedHTML(() => {
+    const styles = styledComponentsStyleSheet.getStyleElement();
+    styledComponentsStyleSheet.instance.clearTag();
+    return <>{styles}</>;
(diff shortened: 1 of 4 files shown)


src/app/styles-registry.tsx
```diff
@@ -0,0 +1,31 @@
+/* istanbul ignore file */
+"use client";
+
+import React, { useState } from "react";
+import { useServerInsertedHTML } from "next/navigation";
+import { ServerStyleSheet, StyleSheetManager } from "styled-components";
+
+// https://nextjs.org/docs/app/building-your-application/styling/css-in-js#styled-components
+export default function StyledComponentsRegistry({
```
carlmw, line 9: I noticed this was missing — it's needed to ensure that styled-component styles are server side rendered. Until now there was a noticeable flicker of styles while the CSS loaded.

jack-skerrett-oak (approved):
Only one email!

k-huggs (approved):

PR #2846, opened 2024-10-03
[CUR-902] Maths KS4 tiers, removed duplicates and improved unit specificity

BTitterington:
## Description

Music year: 1966

- Added additional filtering to the getStaticProps curriculum units data.

## Issue(s)

Fixes #

## How to test

1. Go to https://deploy-preview-2846--oak-web-application.netlify.thenational.academy/teachers/curriculum/maths-secondary/units
2. Goto year 10 foundation
3. You should not see Circle theorems
4. Click higher filter
5. You should see Circle theorems

--

1. Go to https://deploy-preview-2846--oak-web-application.netlify.thenational.academy/teachers/curriculum/science-secondary-edexcel/units
2. Combined science and other units should be the same as CAT and the live site

## Checklist

- [ ] Design sign-off


src/pages/teachers/curriculum/[subjectPhaseSlug]/[tab].tsx
```diff
@@ -503,13 +503,65 @@ export function createDownloadsData(
   return downloadsData;
 }
```
Comment (BTitterington) [inline src/pages/teachers/curriculum/[subjectPhaseSlug]/[tab].tsx:505]: I think ideally this would be dealt with in the MV, but it also would have been too time consuming and complex to do that for this ticket.

This function ultimately filters the units to remove duplicates and retain only the most specific units. It identifies units with the same core slugs and checks if a more specific version exists by comparing optional fields (tier_slug, examboard_slug, pathway_slug). If a unit has null values for these optional fields while a more specific version has non-null values, the less specific unit is discarded. Additionally, it ensures only the first occurrence of each unique unit is kept in the final array.

rightsaidjames (approved):
Maths Foundation issues are fixed - the number of units in Y10 and Y11 are correct in both tiers, and units are displaying their Foundation variant (where appropriate) when the Foundation tier is selected. We will need a follow-up ticket to get the same fix into the downloadable documents, but I don't think this should block the merge of this PR.

The initial version of this fix had caused some serious regressions in Science, but as of the latest commit these are now resolved and has even fixed a problem with a unit that had been missing from the Combined Science view.

sonalisin (approved):

PR #2852, opened 2024-10-07
Implement subject header

assadk88:
## Description

- Implemented an updated `CurriculumHeaderTabNav` component for the subject header
- Refreshed the subject header styling
- Updated several components to support the new CurriculumHeaderTabNav
- Added a new `flat` button variant
- Modified CurriculumHeaderTabNav stories and tests to account for the updated tabular navigation.
- Visually removed redundant Curriculum tab H2 headings, so that they are now only visible to screen readers
    - A change was also made in the [oak-components](https://github.com/oaknational/oak-components/pull/298) repo to the `OakDownloadsJourneyChildSubjectTierSelector` component to facilitate this 
- @rightsaidjames please note that it was discussed with @HelenHarlow that for the 'selected tab' hover over should have no effect – so if in the course of your testing you're hovering over the currently selected tab and nothing occurs that is by design. Mouse hover over should however affect the remaining two unselected tabs however.

## Issue(s)

Fixes #CUR-869

## How to test

1. Go to https://deploy-preview-2852--oak-web-application.netlify.thenational.academy.
2. Navigate to the Teachers' Curriculum section.
3. Select a subject and phase.
4. You should see the new `CurriculumHeaderTabNav` component with the new navigation tabs

## Screenshots

image

How it should now look:

image

## Checklist

- [X] Design sign-off
- [x] Manually tested on Chrome
- [ ] Does this PR update a package with a breaking change – No


HelenHarlow (approved):
Looks great!

src/components/CurriculumComponents/CurriculumHeaderTabNav/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+      as="nav"
+      aria-label={label}
+      $mv={0}
+      $pv={0}
+      $overflowY={"hidden"}
+      $overflowX={"auto"}
+      {...flexProps}
+    >
+      {links.map((link, i) => {
+        return (
```
sonalisin, line 35: I think you can just return the component, e.g. 
```suggestion
        {links.map((link, i) => <ButtonAsLink
            {...link}
            size="large"
            ...
```

as this is more consistent with the maps elsewhere

src/components/SharedComponents/Button/ButtonAsLink/ButtonAsLink.tsx
```diff
@@ -42,22 +49,35 @@ const ButtonAsLink: FC<ButtonAsLinkProps> = (props) => {
...
+    ...transformedProps,
+    variant,
+  });
 
   // aria-label overrides label.
   // If labelSuffixA11y is provided, it is appended to the label.
   const defaultTitle =
     ariaLabel ?? (labelSuffixA11y ? `${label} ${labelSuffixA11y}` : label);
 
+  const handleClick = (e: React.MouseEvent<HTMLAnchorElement>) => {
```
sonalisin, line 67: Good stuff! Thanks for bringing this out into a proper handler function :)

src/components/SharedComponents/Button/ButtonInner.tsx
```diff
@@ -88,14 +89,14 @@ const ButtonInner: FC<ButtonInnerProps> = (props) => {
     labelSuffixA11y,
     shouldHideLabel,
     background,
-    variant,
+    variant = "minimal",
```
sonalisin, line 92: Will this impact other buttons across the app? Might be worth checking with @rightsaidjames

src/components/SharedComponents/Button/common.ts
```diff
@@ -19,7 +21,8 @@ export type ButtonVariant =
   | "brushNav"
   | "minimal"
   | "minimalNav"
-  | "buttonStyledAsLink";
+  | "buttonStyledAsLink"
+  | "flat";
```
sonalisin, line 25: Maybe not to do in this PR, but is "flat" the best description for the button variant? As it's only really used in the curric header it might be worth brainstorming with @HelenHarlow the best style name for this

src/utils/curriculum/constants.ts
```diff
@@ -1,4 +1,4 @@
-export const ENABLE_CYCLE_2 = false;
+export const ENABLE_CYCLE_2 = true;
```
sonalisin, line 1: This should be `ENABLE_CYCLE_2 = false` so we don't accidentally release Cycle 2 early 😜

sonalisin (approved):
Looks good to me - a few little comments but great work!

src/components/SharedComponents/Button/common.ts
assadk88: Good point, it could definitely do with being more descriptive

src/components/SharedComponents/Button/ButtonInner.tsx
assadk88: Good spot. I've defaulted this to `const DEFAULT_BUTTON_VARIANT: ButtonVariant = "brush";` – which can be found [here](https://github.com/oaknational/Oak-Web-Application/blob/51e8a50fddf428750848048a8e1c477a6a1624bc/src/components/SharedComponents/Button/common.ts#L270), so this now shouldn't affect anything.

src/components/CurriculumComponents/CurriculumHeaderTabNav/index.tsx
assadk88: Done

src/components/CurriculumComponents/OverviewTab/OverviewTab.tsx
```diff
@@ -252,9 +253,27 @@ const OverviewTab: FC<OverviewTabProps> = (props: OverviewTabProps) => {
           </OakBox>
         </Box>
       )}
-      <Box $maxWidth={1280} $mh={"auto"} $ph={16} $width={"100%"}>
+      <Box
+        id="curriculum-overview"
+        aria-labelledby="curriculum-overview-heading"
```
rightsaidjames, line 258: This is a cool idea, but unfortunately the `aria-labelledby` property is [not allowed](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby) on the `generic` role, which is the implicit role for `<div>`. Can the <Box> component have a `tag` of `article` or `section`? Otherwise, you could give it a role of [region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/region_role) which would then allow the `aria-labelledby` attribute to be used with no further modifications.

src/components/CurriculumComponents/UnitsTab/UnitsTab.tsx
```diff
@@ -204,15 +204,25 @@ const UnitsTab: FC<UnitsTabProps> = ({ trackingData, formattedData }) => {
...
-          tag="h2"
-          $mb="space-between-m"
-          $ml={["space-between-s", "space-between-none"]}
-          $font={["heading-5", "heading-4"]}
-        >
-          Unit sequence
-        </OakHeading>
+      <Box
+        id="curriculum-units"
+        aria-labelledby="curriculum-unit-sequence-heading"
```
rightsaidjames, line 209: See previous comment re: use of `aria-labelledby`

src/components/CurriculumComponents/CurriculumDownloadTab/index.tsx
```diff
@@ -321,7 +323,20 @@ const CurriculumDownloadTab: FC<CurriculumDownloadTabProps> = ({
 
   return (
     <OakThemeProvider theme={oakDefaultTheme}>
-      <Box $maxWidth={1280} $mh={"auto"} $ph={18} $pb={[48]} $width={"100%"}>
+      <Box
+        id="curriculum-downloads"
+        aria-labelledby="curriculum-downloads-heading"
```
rightsaidjames, line 328: See previous comment re: use of `aria-labelledby`

src/components/CurriculumComponents/OverviewTab/OverviewTab.tsx
```diff
@@ -307,14 +326,6 @@ const OverviewTab: FC<OverviewTabProps> = (props: OverviewTabProps) => {
                 $maxWidth={["100%", "100%", "65%"]}
                 $textAlign={"left"}
               >
-                <OakHeading
```
rightsaidjames, line 310: You've removed this heading but it should instead be visually hidden like the Cycle 2 enabled version above.

src/components/CurriculumComponents/CurriculumHeaderTabNav/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+      {...flexProps}
+    >
+      {links.map((link, i) => (
+        <ButtonAsLink
+          {...link}
+          size="large"
+          variant={variant}
+          aria-current={link.isCurrent ? "page" : undefined}
+          aria-controls={link.page}
+          key={`CurriculumHeaderTabNav-${link.page}-${i}`}
```
rightsaidjames, line 41: @HelenHarlow @assadk88 If we implemented the extra line I've added below on this component, then the following would apply for the currently active tab:
* The link would no longer be clickable or activatable using a keyboard.
* The link would not have any hover or focus styles (currently it has focus styles but not hover styles)
* The link would no longer be in the tab index.
* The link would be reported as a `dimmed` or `disabled` link by screen readers.
* The link would still be in screen readers' list of links.

Is this a feature we want, or is it better that the link be clickable even when it is pointing to the currently active tab? We could also tackle this as a separate issue, as it's not an accessibility issue to leave it as-is.

```suggestion
          aria-controls={link.page}
          disabled={link.isCurrent ? true : undefined}`
          key={`CurriculumHeaderTabNav-${link.page}-${i}`}
```

HelenHarlow: Hey! The reason it's missing hover styles currently is that we would need to add an additional hover state for the 'selected' tab, as the component looks different (it has a line on the top). This didn't feel necessary seeing as clicking on that tab in the selected state won't do anything but reload the page. As long as screen reader users can still obtain the full list of links in the tabs, as this is important for understanding the structure of the page, then I'm happy with whatever approach. Hopefully this extra context helps?

rightsaidjames: Yup, makes sense, so my question to you is would you like the currently active tab to be a clickable link, or would you prefer it to be a 'disabled'/'dimmed' link? There is no a11y impact either way, so it's more of a UX/Design question.

rightsaidjames: If we left it as is, the link still would be reported as 'current' or 'active' by screen readers, so even though it's clickable it should still be clear that activating the link won't do anything except refresh the page.

rightsaidjames: Great, thanks!

src/components/CurriculumComponents/OverviewTab/OverviewTab.tsx
```diff
@@ -252,9 +253,27 @@ const OverviewTab: FC<OverviewTabProps> = (props: OverviewTabProps) => {
...
           <OakFlex $gap={"all-spacing-16"} $alignItems={"flex-start"}>
+            <ScreenReaderOnly>
+              <OakHeading
+                tag="h2"
+                $font={["heading-5", "heading-4"]}
+                $mb="space-between-m"
+                data-testid="overview-heading"
+                id="curriculum-overview-heading"
+              >
+                Overview
```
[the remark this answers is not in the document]
assadk88: Yep, you're right, it's redundant

src/components/CurriculumComponents/OverviewTab/OverviewTab.tsx
assadk88: Good point – I've added the role attribute

src/components/CurriculumComponents/UnitsTab/UnitsTab.tsx
assadk88: As above,  I've added the role attribute

src/components/CurriculumComponents/CurriculumDownloadTab/index.tsx
assadk88: As above,  I've added the role attribute

src/components/CurriculumComponents/OverviewTab/OverviewTab.tsx
assadk88: Reinstated

rightsaidjames (approved):
Looks good, LGTM.

All a11y issues confirmed as fixed.

PR #2859, opened 2024-10-08
LESQ-1106: fix for broken canonical share pages

k-huggs:
## Description

Music year: 1963

- passes down lesson pathways to share button, in the even there's no programme slug, finds the first available slug to navigate to share page

#### N.B. Fixes to the canonical pages are probably required in a few areas

New ticket created to make a canonical share page - https://www.notion.so/oaknationalacademy/I-want-a-canonical-Share-page-11a26cc4e1b1801999f8e891ce8000ff

## Issue(s)

Fixes #LESQ-1106

## How to test

1. Go to https://deploy-preview-2859--oak-web-application.netlify.thenational.academy
2. Navigate to any canonical pages (including KS4 Maths and Science), share button should link to a share page

## Screenshots

https://github.com/user-attachments/assets/b73643ef-196b-4440-b6bf-ad18ac25b6a5

## Checklist

- [ ] Design sign-off


src/components/TeacherComponents/helpers/lessonHelpers/lesson.helpers.ts
```diff
@@ -52,6 +52,10 @@ export const getCommonPathway = (
           acc[key] = null;
```
njblore, line 52: > /**
 * Returns the intersection different pathways.
 * Where a property is different, the value is set to null.
 */
 seems like this is working 'as intended', but the ks4 subjects differ on so many PFs they're getting reset to null more often than not...has this just always been a bug? 👀  we really need a default rather than a null wherever this is being used to generate the share page?

k-huggs: Yeah I'm thinking this could have been a previous bug, it seems to mainly be affected by tiers as ks4 English and History both have examboards. This function returns out the common props, the programmeSlug is passed to the share button which is used to navigate to the share page

k-huggs:
[Ticket created](https://www.notion.so/oaknationalacademy/I-want-a-canonical-Share-page-11a26cc4e1b1801999f8e891ce8000ff) to fix this

rightsaidjames (approved):
Looks good. One question though - as a result of this change, canonical lesson pages have their own share pages. Previously, users were being redirected to the share page for an equivalent non-canonical lesson. I assume this behaviour is not an issue?

Also, does this clash with #2880 @njblore ?

k-huggs:
> Looks good. One question though - as a result of this change, canonical lesson pages have their own share pages. Previously, users were being redirected to the share page for an equivalent non-canonical lesson. I assume this behaviour is not an issue?
> 
> Also, does this clash with #2880 @njblore ?

Sorry meant to close this PR. We're going to continue with Tash' work as it's a better overall fix, rather than just a stop gap!

PR #2864, opened 2024-10-09
Remove order hack, now cooking and nutrition data has been fixed

orangemug:
## Description
Previously cooking and nutrition was missing the `order` field. That has now been resolved upstream.

## Issue(s)
Fixes `CUR-924`

## How to test
Enable cycle 2 and check opening any of the sequences doesn't "error out" with errors regarding missing order field.


sonalisin (approved):

rightsaidjames (approved):
Looks good - the page does not error, and the units in each year are now consistent with the CAT.

PR #2866, opened 2024-10-09
LESQ-1084: Registration tracking

carlmw:
## Description

Music year: {owa_music_year}

- List of changes

## Issue(s)

Fixes #

## How to test

1. Go to {owa_deployment_url}
2. Click on \_\_\_\_\_\_

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist

- [ ] Design sign-off


src/hooks/useTrackRegistration.ts
```diff
@@ -0,0 +1,89 @@
...
+export function useTrackRegistration() {
+  const { track } = useAnalytics();
+  const { user } = useFeatureFlaggedClerk().useUser();
+  const lastUserRef = useRef(user);
+
+  // Would prefer that this were dealt with in an event driven manner.
+  // However Clerk only offer webhooks for this which would be a lot of
+  // infrastructure work and require orchestration to tie the event back to
+  // the correct person in analytics. If we do decide to invest time in Clerk
+  // webhooks later this would be a great candidate to be lift off the client
```
carlmw, line 21: mybest

src/components/TeacherComponents/OnboardingForm/onboardingActions/getSubscriptionStatus.ts
```diff
@@ -0,0 +1,33 @@
+import OakError from "@/errors/OakError";
```
carlmw, line 1: No code changes, just extracted this from onboardingActions.ts

src/components/TeacherComponents/OnboardingForm/onboardingActions/onboardUser.ts
```diff
@@ -0,0 +1,47 @@
+import errorReporter from "@/common-lib/error-reporter";
```
carlmw, line 1: No code changes, just extracted this from onboardingActions.ts

src/components/TeacherComponents/OnboardingForm/onboardingActions/setOnboardingLocalStorage.ts
```diff
@@ -0,0 +1,45 @@
+import useLocalStorageForDownloads from "../../hooks/downloadAndShareHooks/useLocalStorageForDownloads";
```
carlmw, line 1: No code changes, just extracted this from onboardingActions.ts

src/components/TeacherComponents/OnboardingForm/onboardingActions/submitOnboardingHubspotData.ts
```diff
@@ -0,0 +1,64 @@
+import { OnboardingFormProps } from "../OnboardingForm.schema";
```
carlmw, line 1: No code changes, just extracted this from onboardingActions.ts

src/components/TeacherComponents/OnboardingForm/getQueryParamsFromOnboardingFormData.ts
```diff
@@ -1,34 +0,0 @@
-import { ParsedUrlQuery, encode } from "querystring";
-
-import {
-  ManualSchoolFormValues,
-  SchoolSelectFormProps,
-  UkSchoolFormValues,
-} from "./OnboardingForm.schema";
-
-export const getQueryParamsFromOnboardingFormData = (
```
carlmw, line 9: This was sort of doing what I needed to allow tracking of the accumulated form state through the onboarding flow. 

I've refactored it into `encodeOnboardingDataQueryParam` and `decodeOnboardingDataQueryParam` which collects the form state in base64 JSON making it easy to deserialise back into `OnboardingFormProps`

src/browser-lib/avo/getAvoBridge.ts
```diff
@@ -50,8 +50,21 @@ const getAvoBridge = ({ posthog }: AnalyticsServices) => {
...
+  const identify: CustomDestination["identify"] = (userId) => {
+    posthog.identify(userId, {});
+  };
+
+  const setUserProperties: CustomDestination["setUserProperties"] = (
+    userId,
+    properties,
+  ) => {
+    posthog.identify(userId, properties);
+  };
```
carlmw, line 62: This appears to be the first time that we're setting user properties through Avo and this wasn't supported by the existing analytics setup. Defining these functions ensures that the new methods route analytics calls to the correct place.

src/components/TeacherComponents/OnboardingForm/onboardingActions/collectOnboardingTrackingProps.ts
```diff
@@ -0,0 +1,124 @@
...
+} from "@/browser-lib/avo/Avo";
+
+/* Reported for manually entered schools or as a fallback when the URN cannot be extracted
+from the school picker data */
+const DUMMY_URN = "000000";
+
+/**
+ * Produces onboarding tracking properties from the user and onboarding form
+ */
+export function collectOnboardingTrackingProps(
```
carlmw, line 23: Why must tracking require so much code be delivered to the client?!

desmondwat

src/components/AppComponents/App/AppHooks.tsx
```diff
@@ -38,6 +39,7 @@ const useAppHooks = () => {
       !router.pathname.startsWith("/videos"), // Disable Gleap for standalone video pages
   });
   useAxe({ enabled: getBrowserConfig("axeA11yLogging") === "on" });
+  useTrackRegistration();
```
carlmw, line 42: Registering the hook to detect sign-ins, sign-ups and sign-outs

src/components/TeacherComponents/OnboardingForm/onboardingActions/collectOnboardingTrackingProps.ts
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * Pulls the URN/IRN from the school picker output
+ *
+ * English and Welsh URNs are 6 digits
+ * NI IRNs include a hypen `999-9999`
+ * Scottish URNs are 7 digits
+ */
+function extractUrn(school: string) {
+  return /^\d{7}|^\d{6}|^\d{3}-\d{4}/.exec(school)?.at(0);
```
njblore, line 142: ❗this has just made me realise there's a bug in the current school picker since adding NI schools! We get the URN by splitting at the hyphen... 🙃

src/browser-lib/avo/getAvoBridge.ts
njblore: I'm not 100% but it looks like the `withQueue` method that posthog is used with implements the identify method?

carlmw: That's what this eventually calls 🤣 took me a while to burrow through the abstractions to grok.

Screenshot 2024-10-11 at 09 58 21

`getAvoBridge` returns an interface that Avo's codegen expects. All it does is call methods on the posthog abstraction you mentioned. Without these wrappers in the bridge it was blowing up in the codegen.

src/components/TeacherComponents/OnboardingForm/onboardingActions/collectOnboardingTrackingProps.ts
carlmw: Always happy to cause issues

src/browser-lib/avo/getAvoBridge.ts
njblore: Ohhh... yeah I was sure i'd seen identify somewhere but couldn't parse what was going on at all on inspection 😂 fair enough!

njblore (approved):

src/components/TeacherComponents/OnboardingForm/onboardingActions/collectOnboardingTrackingProps.ts
ewan-wakeman: For my understanding is this an Avo issue or something separate. I'm building up a bit of an "evaluation" of how useful Avo is as a middle-man for handling Posthog events.

src/components/TeacherComponents/OnboardingForm/onboardingActions/collectOnboardingTrackingProps.ts
ewan-wakeman: To add to this. Do you know if we have LEA/Estab (for English/Welsh schools) as a possible field in the Aircury data. I think the reason for the oddness here is that NI/Scottish IDs are actually just representations of a `/\d{3}\d{4}/` code where the first three digits point to a local authority and the latter four point to an Establishment. This ID is typically a lot more stable than URN (which changes whenever schools convert to academy amongst other things). I wonder if would be good long term to only track URN for English Schools and to grab the 7 digit LEA/Estab code for all schools across the UK.

carlmw: @ewan-wakeman AIrcury get their data for English schools from GIS [which does appear to offer the LAESTAB](https://get-information-schools.service.gov.uk/Establishments/Search?SelectedTab=Establishments&SearchType=Location&SearchType=Location&LocationSearchModel.Text=London%2c+London&LocationSearchModel.AutoSuggestValue=51.50014877319336%2c+-0.12624000012874603&OpenOnly=true&b=1&b=4), but Aircury don't currently include it in their API responses.

E.g. this is what we currently get.
```json
{"urn":"100000","la":"City of London","name":"The Aldgate School","postcode":"EC3A 5DE","fullInfo":"100000, City of London, The Aldgate School, EC3A 5DE","status":"Open"}
``` 

It would likely be a chargeable piece of work for it to be added so that we can access it.

ewan-wakeman: Ok cool, it's not urgent at all, just might be one to consider longer term I suppose, seeing as how this seems to align more neatly with Scotland and NI. In the meantime we can deal with this nuance further down in the pipeline and just accept some slight more tricky regex for now.

carlmw:
@njblore re-requesting a review for the latest 2 commits. I was testing sign-ups on the deploy preview (now that I can) and noticed a bug where I was getting thrown out of the feature flag after sign-up. Because we use `distinct_id` to enable the FF we can't identify a user with any other id for now. Once we remove the feature flag we should be good to identify the user with whichever id we want (if we want).

src/browser-lib/posthog/posthog.ts
```diff
@@ -22,22 +22,27 @@ export const posthogToAnalyticsServiceWithoutQueue = (
...
+            }
+
+            resolve(client.get_distinct_id());
+          },
+          disable_session_recording: true,
+          capture_pageview: false,
+        });
+      } else {
+        resolve(client.get_distinct_id());
+      }
```
carlmw, line 45: Super subtle bug, but this promise would never resolve if Posthog had already been initialised. This happens in OWA currently because `AnalyticsProvider` is unmounted repeatedly on page load (not great).

This meant that `useAnalytics().posthogDistinctId` would always be `null`

src/components/TeacherComponents/OnboardingForm/onboardingActions/collectOnboardingTrackingProps.ts
```diff
@@ -0,0 +1,144 @@
...
+    .slice(0, 10);
+
+  return {
+    platform: "owa",
+    product: "user account management",
+    engagementIntent: "explore",
+    componentType: "continue_button",
+    eventVersion: "2.0.0",
+    analyticsUseCase: "Teacher",
+    userId_: distinctUserId,
```
carlmw, line 42: Continue to identify the user with their posthog id. Failing to do this causes users to be thrown out of the `use-auth-owa` feature flag at the start of onboarding.

njblore (approved):

jack-skerrett-oak (changes requested):
Largely it seems OK, some little things I noticed that could possibly be tidied? 

- [ ]  There is a mismatch in the event version. On the Avo event it is listed as 2.1.0, in posthog it comes through as 2.0.0 (though only a couple of the events have the event version written in avo (sign up, onboarding complete)
- [ ]  Do we still live in a world where “User onboarding progressed” is needed? I don’t think the event is fired, instead we are “Set”ting user properties (one for the avo side more than anything)
- [ ]  There are two “Set” events sent whenever one should be triggered. I’m not sure if both are essential, just seems wasteful
- [ ]  The “User Onboarding Completed” has a property “component_type” that is set to “continue_button” when going through the “Yes” journey whether you select the “Continue” button or the “Skip” button
- [ ]  On the “No” journey, the “User role type” is initially set to “Non-teacher”, this is then updated based on information in the following form, is there any value in updating it when the user selects no at on the “Do you work in a school” question?

carlmw:
> Largely it seems OK, some little things I noticed that could possibly be tidied?
> 
> * [ ]   There is a mismatch in the event version. On the Avo event it is listed as 2.1.0, in posthog it comes through as 2.0.0 (though only a couple of the events have the event version written in avo (sign up, onboarding complete)
> * [ ]   Do we still live in a world where “User onboarding progressed” is needed? I don’t think the event is fired, instead we are “Set”ting user properties (one for the avo side more than anything)
> * [ ]   There are two “Set” events sent whenever one should be triggered. I’m not sure if both are essential, just seems wasteful
> * [ ]   The “User Onboarding Completed” has a property “component_type” that is set to “continue_button” when going through the “Yes” journey whether you select the “Continue” button or the “Skip” button
> * [ ]   On the “No” journey, the “User role type” is initially set to “Non-teacher”, this is then updated based on information in the following form, is there any value in updating it when the user selects no at on the “Do you work in a school” question?

@jack-skerrett-oak 

> There is a mismatch in the event version
This appears to have been a bug in the description. Version 2.1.0 isn't currently an accepted value in the Avo codegen. I've change the description to read "2.0.0" which is all i'm allowed to send.

> Do we still live in a world where “User onboarding progressed” is needed? 
Yes it is needed. Something that isn't obvious about Avo is that Avo events do not map directly to destination (I.e. Posthog) events. They're a different concept. Instead Avo events perform one or more "actions" — up until now that appears to have been entirely the "Log event" action. I found this helpful https://www.avo.app/docs/data-design/avo-tracking-plan/events#actions

> There are two “Set” events sent
Yep, this was down to the fact that two "actions" were registered for the "User Onboarding Completed" and "User Onboarding Progressed" events. One was an "identify" (which can also set user props at the same time) and the other a regular "set user properties". The "identify" was redundant so i've removed it.

> The “User Onboarding Completed” has a property “component_type”
i've gone to great lengths to make this happen 😭. I've added a "skip_button" component type to Avo and started sniffing out which button the user clicked to trigger the submit. 

> On the “No” journey, the “User role type” is initially set to “Non-teacher”
Yes this is very deliberate. The "Non-teacher" value lets us detect when a user drops out of onboarding on the "role" selection form. The intent here is that this generic role is refined by the choice made on the role selection form. It also tells us when they completed that step. Our alternative was to add a single use user prop or log an event to tell use when a user had answered No on the first step.

jack-skerrett-oak (approved):

PR #2869, opened 2024-10-10
All years in mobile header and unit modal

orangemug:
## Description
Adds all-years alternative text in mobile header and unit modal

## Issue(s)

Fixes `CUR-959`

## How to test
 - Enable cycle 2 and test for correct text in mobile header and unit modal.
 - Check for regressions in cycle 1


rightsaidjames (changes requested):
@orangemug Mostly good but, as per the [Figma designs](https://www.figma.com/design/3cz9fb0ObFDLv1c212OtJv/Final-designs---Curriculum-squad?node-id=6242-93342&m=dev) on mobile, the subject and year group need to stack in the unit modal header so that they don't get crammed into a single row:

Screenshot 2024-10-10 at 17 37 40

Screenshot 2024-10-11 at 08 51 33

sonalisin (approved):

rightsaidjames (approved):
Looks good, wraps well on all mobile viewports.

PR #2870, opened 2024-10-10
chore: bump oak components

njblore:
## Description

Music year: 1960

- Bump oak components to include new subject icon `subject-cooking-nutrition`
- This icon does not exist in cloudinary yet (should be added this week) so will appear broken until then: 
Screenshot 2024-10-10 at 14 31 01

https://deploy-preview-2870--oak-web-application.netlify.thenational.academy/teachers/search?term=macbeth

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "@mdx-js/loader": "^3.0.1",
         "@mux/mux-node": "^8.8.0",
         "@mux/mux-player-react": "2.7.0-canary.0-7c57cdd",
-        "@oaknational/oak-components": "^1.38.2",
+        "@oaknational/oak-components": "^1.40.0",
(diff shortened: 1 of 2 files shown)


k-huggs (approved):

rightsaidjames (commented):
I suggest that we **don't** merge this until the icons are actually in Cloudinary, as doing so would put us in a worse position than we are on production:

rightsaidjames:
... on the other hand, if anyone else wants to bump oak-components as part of their PR then it will include this change anyway, so I guess we may as well merge this now!

njblore:
@rightsaidjames yeah it's not ideal, but it is in a quite tucked away place and hopefully won't be missing for long 🤞and I'm keen to clear out the error reports for this one in case there are any genuine missing icons getting lost in the spam!

PR #2891, opened 2024-10-16
Switch to using `subjectPhaseOptions` so we have `ks4_options`

orangemug:
## Description
Previously we were using `subjectPhaseOptionsIncludeNew` which doesn't include `ks4_options` changes. We can switch to `subjectPhaseOptions` as we no longer require the "new" state

## Issue(s)

Fixes `CUR-983`

## How to test

1. Go to https://deploy-preview-2891--oak-web-application.netlify.thenational.academy/teachers/curriculum/
2. Test downloads still work as expected with correct examboard/pathway in cover/filepath


sonalisin (approved):

rightsaidjames (approved):
Looks good:
- Tested with cycle 1 curricula with exam board, tiers and child subjects.
- Regression tested with cycle 1 curricula that don't have KS4 variation (e.g. Maths, Primary subjects)
- Tested lots of cycle 1 and cycle 2 lots locally with the feature flag enabled.
- Confirmed that pathways (Core, GCSE, GCSE + exam board) are being treated correctly in the filename and cover page.

PR #2893, opened 2024-10-16
Don't render non-valid examboard-less subjects/phases

orangemug:
## Description
Don't render (404) non-valid examboard-less subjects/phases. For example `/teachers/curriculum/english-secondary/units` now correctly 404's

## Issue(s)

Fixes `CUR-758`

## How to test

2. Test all subjects/phases are still accessible (including cycle 2)
3. Test that examboard-less slugs 404


sonalisin (approved):

rightsaidjames (approved):
Tested the redirects to preferred exam boards for each subject (e.g. `secondary-english` > `secondary-english-aqa`), it all works well. Nonsensical slugs (e.g. `secondary-maths-aqa` or `secondary-english-ssdgfd` lead to a 404, which is fine because they're not linked from anywhere.

Tested on the preview environment with Cycle 1 and locally with both cycle 1 and cycle 2.

PR #2895, opened 2024-10-16
More percy URLs for curriculum

orangemug:
## Description
Added remaining URLs into percy for curriculum squad 

## Issue(s)

Fixes `CUR-590`

## How to test
Check that percy is happy and CI completes successfully.


assadk88 (approved):

src/common-lib/urls/getDeploymentTestUrls.js
```diff
@@ -71,6 +71,11 @@ function getDeploymentTestUrls() {
     "/teachers/curriculum/english-secondary-aqa/units",
     "/teachers/curriculum/english-secondary-aqa/overview",
     "/teachers/curriculum/english-secondary-aqa/downloads",
+    "/teachers/curriculum/science-secondary-aqa/units",
+    "/teachers/curriculum/science-secondary-aqa/overview",
```
rightsaidjames, line 75: I think this is redundant because it's not significantly different to the English one. But we could add in a cycle 2 explainer page when that goes live?

src/common-lib/urls/getDeploymentTestUrls.js
```diff
@@ -71,6 +71,11 @@ function getDeploymentTestUrls() {
     "/teachers/curriculum/english-secondary-aqa/units",
     "/teachers/curriculum/english-secondary-aqa/overview",
     "/teachers/curriculum/english-secondary-aqa/downloads",
+    "/teachers/curriculum/science-secondary-aqa/units",
```
rightsaidjames, line 74: For this and the science downloads page, can we have a different exam board than the English example?

rightsaidjames (approved):
Other than my review comments, I think we're good for now 🎉

PR #2898, opened 2024-10-17
Simplify and fix tracking event for opening unit modal

orangemug:
## Description
Simplify and fix tracking event for opening unit modal in curriculum visualiser

Note this also fixes double events being triggered when opening the modal. Previously we got two events triggered rather than one.

## Issue(s)
Fixes `CUR-507`

## How to test

1. Go to {owa_deployment_url}
2. Navigate to the curriculum visualiser
3. Open a modal
4. Check only one tracking event is fired.


assadk88 (approved):

rightsaidjames (approved):
Looks good from my perspective - only one event is firing when I open a unit modal, and the count of events is still going up on the Avo events dashboard.

PR #2901, opened 2024-10-17
Added fixture generation via script for curriculum

orangemug:
## Description
Added fixture generation via script for curriculumsquad and updated some tests to use them, as an initial example

This PR adds a new command which updates fixtures based on real data.

```bash
./scripts/dev/curriculum/automate fixtures
```

The fixture processing is defined in `./scripts/dev/curriculum/_commands/fixtures.ts`

## Issue(s)
Fixes `CUR-987`

## How to test
No testing required, if automated test pass we should be good.


orangemug:
Added you all as reviewers @assadk88 @sonalisin @rightsaidjames mainly for feedback. If one of you could also review the PR that would be good.

assadk88 (approved):

sonalisin (approved):

PR #2909, opened 2024-10-22
feat: implement Homepage banner component for Curriculum page

assadk88:
## Description

Music year: 1997

- Created a `HomePageBanner` component

## Issue(s)

Fixes #[991](https://www.notion.so/oaknationalacademy/3bccfe060e4645fe97842977e365c2ad?v=d9869bab5f8a4e28a7a960d37f6ab31e&p=10d26cc4e1b180c5a1a8e58c7441b540&pm=s)

## How to test

1. Go to `/curriculum`
2. You should see the header banner

## Screenshots

How it should now look:

Desktop:
image

`@375px`:
image

## Checklist

- [X] Added or updated tests where appropriate
- [X] Manually tested across browsers / devices
- [X] Considered impact on accessibility
- [X] Design sign-off
- [X] Approved by product owner
- [ ] ~~Does this PR update a package with a breaking change~~

PR #2916, opened 2024-10-23
Suggestions to keep `<ButtonAsLink/>` generic

orangemug:
## Description
Suggested alterations for https://github.com/oaknational/Oak-Web-Application/pull/2909 comments in line

### DIFF
--- a/src/components/SharedComponents/Button/button.styles.ts
+++ b/src/components/SharedComponents/Button/button.styles.ts
@@ -30,13 +30,18 @@ import getColorByLocation from "@/styles/themeHelpers/getColorByLocation";
 import { HOVER_SHADOW_TRANSITION } from "@/styles/transitions";
 import opacity, { OpacityProps } from "@/styles/utils/opacity";
 import margin, { MarginProps } from "@/styles/utils/spacing";
-import { BackgroundIcon } from "@/components/SharedComponents/Icon/Icon";
+import {
+  BackgroundIcon,
+  IconName,
+} from "@/components/SharedComponents/Icon/Icon";
+import Flex from "@/components/SharedComponents/Flex.deprecated";
 import ButtonLabel from "@/components/SharedComponents/Button/ButtonLabel";
 
 export type ButtonStylesProps = OpacityProps &
   MarginProps & {
     size: ButtonSize;
     $iconPosition: IconPosition;
+    icon?: IconName;
     variant: ButtonVariant;
     background: ButtonBackground;
     $fullWidth?: boolean;
@@ -53,6 +58,7 @@ export const getButtonStylesProps = (
   const {
     variant = DEFAULT_BUTTON_VARIANT,
     $iconPosition = DEFAULT_ICON_POSITION,
+    icon,
     size = DEFAULT_BUTTON_SIZE,
     background = DEFAULT_BUTTON_BACKGROUND,
     $fullWidth,
@@ -63,6 +69,7 @@ export const getButtonStylesProps = (
   return {
     size,
     $iconPosition,
+    icon,
     variant,
     $fullWidth,
     background: disabled && variant !== "brushNav" ? "grey50" : background, // we don't discolor brushNav buttons when disabled
@@ -268,14 +275,15 @@ const buttonStyles = css<ButtonStylesProps>`
     props.variant === "buttonStyledAsLink" &&
     props.size === "medium" &&
     css`
-      & svg[name="underline-1"] {
-        top: 27px;
-        width: 96%;
+      & ${ButtonStyledAsLinkFocusUnderline} {
+        display: block;
+        bottom: -1px;
+        left: ${props.$iconPosition === "leading" ? "4px" : ""};
+        width: ${props.icon ? "calc(100% - 4px)" : "100%"};
       }
 
-      & div {
+      & ${Flex} {
         margin-left: 0;
-        font-weight: 600;
       }
     `}
 
--- a/src/common-lib/urls/urls.ts
+++ b/src/common-lib/urls/urls.ts
@@ -400,7 +400,7 @@ const EXTERNAL_PAGE_NAMES = [
 ] as const;
 type ExternalPageName = (typeof EXTERNAL_PAGE_NAMES)[number];
 
-type OakPages = {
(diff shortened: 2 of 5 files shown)


src/components/SharedComponents/Button/button.styles.ts
```diff
@@ -268,14 +275,15 @@ const buttonStyles = css<ButtonStylesProps>`
...
     props.size === "medium" &&
     css`
-      & svg[name="underline-1"] {
-        top: 27px;
-        width: 96%;
+      & ${ButtonStyledAsLinkFocusUnderline} {
+        display: block;
+        bottom: -1px;
+        left: ${props.$iconPosition === "leading" ? "4px" : ""};
+        width: ${props.icon ? "calc(100% - 4px)" : "100%"};
```
orangemug, line 282: Pass though `icon` so we know if it's got an icon or not and reduce the underline

src/components/SharedComponents/Button/button.styles.ts
```diff
@@ -268,14 +275,15 @@ const buttonStyles = css<ButtonStylesProps>`
     props.variant === "buttonStyledAsLink" &&
     props.size === "medium" &&
     css`
-      & svg[name="underline-1"] {
-        top: 27px;
-        width: 96%;
+      & ${ButtonStyledAsLinkFocusUnderline} {
+        display: block;
+        bottom: -1px;
+        left: ${props.$iconPosition === "leading" ? "4px" : ""};
```
orangemug, line 281: If leading icon push left by 4px

src/components/SharedComponents/Button/button.styles.ts
```diff
@@ -268,14 +275,15 @@ const buttonStyles = css<ButtonStylesProps>`
     props.variant === "buttonStyledAsLink" &&
     props.size === "medium" &&
     css`
-      & svg[name="underline-1"] {
-        top: 27px;
-        width: 96%;
+      & ${ButtonStyledAsLinkFocusUnderline} {
```
orangemug, line 278: Use names from `styled-components`

src/components/SharedComponents/Button/button.styles.ts
```diff
@@ -268,14 +275,15 @@ const buttonStyles = css<ButtonStylesProps>`
...
+        display: block;
+        bottom: -1px;
+        left: ${props.$iconPosition === "leading" ? "4px" : ""};
+        width: ${props.icon ? "calc(100% - 4px)" : "100%"};
       }
 
-      & div {
+      & ${Flex} {
```
orangemug, line 285: Use names from `styled-components`

src/components/SharedComponents/Button/button.styles.ts
```diff
@@ -268,14 +275,15 @@ const buttonStyles = css<ButtonStylesProps>`
...
         margin-left: 0;
-        font-weight: 600;
```
orangemug, line 278: I think this isn't required

assadk88: The designs have the font-weight as 600 in Figma, however I'm happy to leave this as 700 as it looks fine and would prevent us from adding further CSS for this exception

HelenHarlow: I think if it can match the font weight in Figma (600), then that would be ideal as this will match the styles used elsewhere. I'm unsure of what the built-in styles of that specific button component are & how much work that is though

orangemug: Yup my mistake @assadk88 is correct

src/components/SharedComponents/Button/button.styles.ts
assadk88: Good point

src/components/SharedComponents/Button/button.styles.ts
assadk88: @orangemug I have a question. 

In [`src/styles/utils/typography.ts`](https://github.com/oaknational/Oak-Web-Application/blob/main/src/styles/utils/typography.ts), the configurations `font-size: 16, font-weight: 600` (needed for the desktop version of the button as per Figma) and `font-size: 16, font-weight: 600` (needed for the mobile version of the button as per Figma) do not exist.

If we want to remain in alignment with the type scale already defined in the app we would have to choose from either:

```
  "body-2": [16, 24, 300, "-0.005em"],
  "body-3": [14, 20, 300, "-0.005em"],
  "body-2-bold": [16, 24, 700, "-0.005em"],
  "body-3-bold": [14, 20, 700, "-0.005em"],
```

but in all of these cases their font-weights are either too high or too low.

This is a long-winded way of saying, should I create a new configuration at font-weight 600 to match up with the Figma designs, or default to a pre-existing one? 

Adding custom CSS to `<ButtonAsLink/>` for font-weight would prevent the component from being implementation agnostic.

orangemug: Probably one for @HelenHarlow. What do you think is font weight `700` ok to keep inline with the other uses of this component?

HelenHarlow: I've had a quick sweep through of the bold text on the website and it all appears to be 600. Basically, I'm happy to align to whatever bold text we're using in the app, but I'm not sure what the correct value is?

assadk88: @HelenHarlow So, further from what we discussed earlier:

* **Font size and weight**: Remove responsiveness from the font-size so it stays a consistent 16px (1rem) for mobile and desktop – hence this will allow us to retain the 600 font-weight
* **Stacking behaviour:** Have the banner stack at screen widths where it can no longer reasonably fit on one line

Having discussed the above with @orangemug , in order to function within the constraints of the DS/component library the stacking behaviour required a bit of custom CSS which would require adding another breakpoint to get the behaviour that you wanted, so we decided against it.

Instead we've decided to to have desktop display on one line, and have it stack as it hits tablet widths and below. Essentially we decided to stack much earlier. I've attached a video to show you what I mean.
I hope that's fine with you?

https://github.com/user-attachments/assets/485ddff3-6de3-4b04-a33d-aed06d1c7a1d

PR #2917, opened 2024-10-23
LESQ-1117: onboarding login fixes

carlmw:
## Description

Music year: {owa_music_year}

## Issue(s)

Fixes
* Download button asking users to Sign in when they have signed in but have not onboarded
* Validation when clicking "Continue" or "Skip" on the "How can Oak support you" step of onboarding

### Scenario #1 — Validation of "How can Oak support you" page

1. Go to https://deploy-preview-2917--oak-web-application.netlify.thenational.academy/onboarding
2. Click on take the teacher journey through onboarding
3. On the "How can Oak support you" step
4. Don't select any options
5. Click "Continue"
6. You should see a validation error
7. Select some options
8. Click "Skip"
9. The form should submit
10. `how_can_oak_support_you_` should be omitted from the form sent to Hubspot

### Scenario #2

1. Sign in as a user who has not onboarded in OWA (`owa.isOnboarded` is not `true` in Clerk)
2. Go to https://deploy-preview-2917--oak-web-application.netlify.thenational.academy/teachers/programmes/citizenship-secondary-ks3-l/units/citizenship-whats-it-all-about-e038/lessons/what-is-citizenship-c4t38d#slide-deck
3. The download button should read "Download all resources"
4. Clicking "Download all resources" should take you to onboarding

## Checklist

- [ ] Design sign-off

### DIFF
--- a/src/components/TeacherViews/Onboarding/HowCanOakSupport/HowCanOakSupport.view.tsx
+++ b/src/components/TeacherViews/Onboarding/HowCanOakSupport/HowCanOakSupport.view.tsx
@@ -91,6 +91,7 @@ const HowCanOakSupport = () => {
             Skip
           </OakSecondaryButton>
         )}
+        continueButtonDescription={formState.errors.root?.message}
       >
         <OakFlex $flexDirection="column" $gap="space-between-s">
           <div
--- a/src/components/TeacherViews/Onboarding/HowCanOakSupport/HowCanOakSupport.view.tsx
+++ b/src/components/TeacherViews/Onboarding/HowCanOakSupport/HowCanOakSupport.view.tsx
@@ -6,7 +6,7 @@ import {
   OakSecondaryButton,
   OakSpan,
 } from "@oaknational/oak-components";
-import { Control, UseFormTrigger, useForm } from "react-hook-form";
+import { Control, Controller, UseFormTrigger, useForm } from "react-hook-form";
 import { useRouter } from "next/router";
 import { useEffect } from "react";
 import Link from "next/link";
@@ -37,7 +37,7 @@ export const oakSupportMap: Record<OakSupportKey, string> = {
 const HowCanOakSupport = () => {
   const router = useRouter();
   const onboardingState = decodeOnboardingDataQueryParam(router.query);
-  const { formState, setValue, handleSubmit, getValues, control, trigger } =
+  const { formState, setValue, handleSubmit, control, trigger } =
     useForm<UseOfOakFormProps>({
       resolver: zodResolver(extendedUseOfOakSchema),
       mode: "onBlur",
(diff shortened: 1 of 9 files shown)


src/components/TeacherViews/Onboarding/HowCanOakSupport/HowCanOakSupport.view.tsx
```diff
@@ -109,6 +115,7 @@ const HowCanOakSupport = () => {
             <FieldError id="missing-values" withoutMarginBottom>
               An error occurred. Please{" "}
               <OakLink
+                element={Link}
```
carlmw, line 118: Navigate through Next router to avoid a full page reload

src/components/TeacherViews/Onboarding/HowCanOakSupport/HowCanOakSupport.view.tsx
```diff
@@ -86,17 +81,28 @@ const HowCanOakSupport = () => {
...
+        }}
         secondaryButton={(isSubmitting) => (
           <OakSecondaryButton
             width="100%"
             disabled={hasMissingFormData || isSubmitting}
             name="skip"
+            onClick={() => {
+              setValue("submitMode", "skip");
+              trigger();
+            }}
```
carlmw, line 96: Flip `submitMode` between continue and skip to trigger the corresponding validation and transforms from the schema.

I'm not sure if this is idiomatic use of react-hook-form but the docs and discussions were unhelpful. I think another way would have been to fork the behaviour in the form `resolver` but that would still need someway to flip some state to say which button was clicked. I saw some examples that used an `onClick` on the secondary button to trigger `handleSubmit` but that would have required widespread refactoring of `OnboardingForm` which would have been a nightmare.

RBrownsett:
@carlmw a few notes on the accessibility of the error message, currently it is not clear what a screen reader user needs to do when they do not select an option (just that the continue button is disabled and greyed out). Possible suggestions:

- Use `aria-live="assertive"` rather than polite to inform users of the error message
- Have the focus move to the top of the checkbox when they try and click continue rather disabling the button (similar to how the contact us validation works here: https://www.thenational.academy/contact-us
- Update the error text so that it reads more like an action e.g. "Please select the ways how Oak can support you"

njblore (approved):

carlmw:
@RBrownsett i've made some changes based on your suggestions.

1.`aria-live` is now set to `assertive` — I tested this out in VO and it seems that the `FieldError` component doesn't actually work as expected since it doesn't add the region until the error is being rendered which isn't supported with `aria-live`. To get VO to announce I had to use it purely for rendering and combine `aria-live` with `aria-description` to get it to announce when validation failed.
2. I had the first checkbox receiving focus on error but this mean the error above was not announced. The focus announcement took precedence. 
3. I've updated the error text as suggested to be more actionable

Let me know if this is enough or if you have some alternatives to get the focus working.

jack-skerrett-oak (approved):

PR #2921, opened 2024-10-23
Add `<IconButtonFocusVisible/>` to allow `:focus-visible` instead of `:focus`

orangemug:
## Description
Changing `:focus` to `:focus-visible` in `<IconFocusUnderline/>` which effects all the core components.

## Issue(s)

Fixes `CUR-973`

## How to test
1. Go to https://deploy-preview-2921--oak-web-application.netlify.thenational.academy/teachers/curriculum/english-secondary-aqa/units
2. Open a unit modal using a mouse and check for lack of focus outline on the close button.
3. Close the modal.
4. Open a unit modal using a keyboard and check that the focus outline on the close button is still present.
5. Repeat steps 2-4 using an optionality modal.


sonalisin:
Might be worth asking @HelenHarlow if there'll be any impact for this... Not sure which buttons aren't the shared component ones

sonalisin (commented):
Might also be worth having another style with a class for the changes? E.g.

`.curriculumButton {
      :focus-visible ${IconFocusUnderline} {
          display: block;
          bottom: 0;
          left: 0;
       }
}`
etc.
So we can change it on our side but won't impact other buttons sitewide??

njblore:
can't see any issue from the teacher side, but would upgrading to [Tertiary button with no text ](https://lively-meringue-8ebd43.netlify.app/?path=/story/components-molecules-oaktertiarybutton--tertiary-button-with-no-text)component be an option here? or is the brush underline focus still required? 😬

orangemug:
> can't see any issue from the teacher side, but would upgrading to [Tertiary button with no text ](https://lively-meringue-8ebd43.netlify.app/?path=/story/components-molecules-oaktertiarybutton--tertiary-button-with-no-text)component be an option here? or is the brush underline focus still required? 😬

Yeah one for us to discuss with @HelenHarlow

orangemug:
@sonalisin you were right, I backed away from changing `<IconButton/>` everywhere to a single override instead.

@rightsaidjames @sonalisin review please 🙏

sonalisin (approved):

rightsaidjames (approved):
Looks good. The decision on whether to display focus styles for the close elements is in the hands of the browsers now, so it's not completely consistent. Sometimes the browser will take a few interactions to decide whether you should be shown focus styles or not, and certain sequences of keyboard interactions will cause focus styles to linger longer than one might expect, but it seems reasonably logical, so I will conclude that the CSS changes are definitely being applied correctly. Tested in Chrome, Safari and Firefox on macOS, Safari on iOS, Edge on Windows 11 (via BrowserStack) and Chrome/Chromium on Android 12.

PR #2926, opened 2024-10-24
Added in `<FocusWrap/>` component so we can deal with complex focus logic

orangemug:
## Description
Adds `<FocusWrap/>` component so we can deal with complex focus logic and movement across multiple modals.

This allows us to wrap some focusable elements with `<FocusWrap>{/*...*/}</FocusWrap>` and react to "focus wrapping" events.

This PR also adds the focus logic as described in `CUR-878`

@rightsaidjames are you ok to do an early smoke test for a11y issues? 

## Issue(s)

Fixes `CUR-878`

## How to test

1. Go to https://deploy-preview-2926--oak-web-application.netlify.thenational.academy/teachers/curriculum
2. Check tab behaviour is correct as per designs in `CUR-878`
3. Check for any accessibility issues created by the designs


orangemug:
@sonalisin @assadk88 I'll assign you for a proper review later on, but what do y'all think of the general approach?

HelenHarlow (approved):
Looks good to me!

sonalisin (approved):

rightsaidjames (approved):
Tested with Chrome, Firefox and Safari with and without VoiceOver, and on Windows with Edge + NVDA (via BrowserStack). Overall it's looking really good. The focus behaviour is the same, except that the user is now no longer trapped inside the modal - they can tab past the end of the modal to get to the next part of the subject-phase picker.

LGTM!

rightsaidjames (approved):
Re-approved following refactoring

PR #2953, opened 2024-10-31
fix(LESQ-1155): guard against duplicate sign-up events when hook is remounted

carlmw:
## Description

Music year: 1953

## Issue(s)

Fixes
- A race condition causing sign-up events to be tracked multiple times

## How to test

1. [With the `teacher-download-auth` feature flag set to `with-login` ](https://eu.posthog.com/project/124/feature_flags/35482) and while signed-out
2. Go to https://deploy-preview-2953--oak-web-application.netlify.thenational.academy/teachers/programmes/computing-secondary-ks3-l/units/impact-of-technology-collaborating-online-respectfully-35d0/lessons/account-security-68rkee
4. Click "Sign in to download"
5. With the dev console opened throttle your connection to "Fast 4G" (or sit near a source of interference to disrupt your WiFi a bit 🤸🏻‍♂️ maybe)
6. Complete the sign-up process
7. When you're redirected to `/onboarding` you should see only one "User Sign-Up Completed" event sent to Posthog
8. This is a race condition so this might need to be repeated several times to be sure (though I was fairly reliably able to reproduce the condition with the throttled connection)

## Checklist

- [ ] Design sign-off


src/hooks/useTrackRegistration.ts
```diff
@@ -25,14 +34,23 @@ export function useTrackRegistration() {
...
       user.unsafeMetadata.owa?.lastTrackedSignInAt ===
       user.lastSignInAt?.valueOf()
     ) {
       return;
     }
 
+    // Skip tracking when the hook has been re-mounted/re-rendered
+    // this guards against a race condition caused by the hook being unmounted
+    // and re-rendered while the Clerk user metadata is being updated
+    if (hasAlreadyTrackedEvent) {
```
carlmw, line 48: I was relying on the dependencies array to prevent the hook from firing again on subsequent renders while the user metadata update was taking place. I really should have known better than to trust `useEffect` in this scenario.

It took me a while to get to the bottom of what was going on but ultimately the hook is unmounted repeatedly.
1. Once on initial page load (when the Bugsnag error boundary is added to the tree)
2. On every navigation through the pages router (I didn't know this about the pages router, this is different in app router IIRC where the tree isn't thrown away).

It was on navigation that things were going wrong:

1. The user completes sign-up
3. We redirect to the lesson download page
4. The hook fires a few times
5. Dependencies settle when the user is loaded through `useUser`
    * The hook fires with all dependencies met
    * The tracking call is made
    * A request is made to update user metadata in Clerk with the last tracked sign-in (this is not an optimistic update)
6. In or around the same tick that the user is loaded we redirect to `/onboarding`
   * The hook is remounted throwing away all state
   * The hook fires immediately since all dependencies have been met
     * Critically the update user metadata request has not responded yet
   * The tracking call is made again
   * Another request is sent to update user metadata with the last tracked sign-in
 7. Both requests to update user metadata eventually respond and the last tracked date sign in settles.

src/hooks/useTrackRegistration.ts
```diff
@@ -5,6 +5,15 @@ import useAnalytics from "@/context/Analytics/useAnalytics";
 import type { UserResource } from "clerk";
 import { SingleSignOnServiceValueType } from "@/browser-lib/avo/Avo";
 
+let hasAlreadyTrackedEvent = false;
```
carlmw, line 8: dafoe

k-huggs (approved):

mikeritson-oak (approved):
After the initial issue with testing this change, no further instances were seen and outcomes were as expected. Even the initial results seem to have updated or changed at some point. 

The squad has decided to approve this ticket and monitor for any further instances occurring. 👍

PR #2966, opened 2024-11-07
[CUR-1031] fix: update sanity query for equality of subject title

sonalisin:
## Description

Music year: 2017

- List of changes
- Changed GQL query to search by title instead of slug (no slug available in sanity)
- Changed query to look for equality instead of matching

## Issue(s)
PE and RE had switched explainers in prod due to fuzzy matching during querying. Hadn't had the issue before as most subjects are single words and not phrases

## How to test

1. Go to https://deploy-preview-2966--oak-web-application.netlify.thenational.academy/teachers/curriculum
2. Click on Physical education primary/secondary
3. You should see the correct explainer
4. Same for RE

## Screenshots

Screenshot 2024-11-07 at 13 49 21

How it should now look:
Screenshot 2024-11-07 at 13 49 07

## Checklist

- [ ] Design sign-off

### DIFF
--- a/src/node-lib/sanity-graphql/generated/sdk.ts
+++ b/src/node-lib/sanity-graphql/generated/sdk.ts
@@ -3905,7 +3905,7 @@ export type CurriculumCorePageQuery = { __typename?: 'RootQuery', allCurriculumC
 
 export type CurriculumOverviewQueryVariables = Exact<{
   isDraftFilter?: InputMaybe<Sanity_DocumentFilter>;
-  subjectSlug?: InputMaybe<Scalars['String']['input']>;
+  subjectTitle?: InputMaybe<Scalars['String']['input']>;
   phaseSlug?: InputMaybe<Scalars['String']['input']>;
 }>;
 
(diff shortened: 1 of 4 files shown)


orangemug:
Questions, if no `slug` how was this producing any of the correct results?

orangemug (approved):

sonalisin:
> Questions, if no `slug` how was this producing any of the correct results?

Yeah, it had me stumped too. But it was using "match" on subjectSlug, so for one worded subjects ("maths") and ("Maths"), then it would be fine. stuff like art and design / cooking and nutrition are fine as there's no other likeness in other subjects. @SarahMcKinless' theory of them both having 'education' got pretty inconsistent results

Using `matches`:
Screenshot 2024-11-07 at 13 58 16

Using `eq`:
Screenshot 2024-11-07 at 13 58 35

orangemug:
If this works as a hot fix we should add a `slug` in after the fact, else changes to title will break in future updates to the content

sonalisin:
> If this works as a hot fix we should add a `slug` in after the fact, else changes to title will break in future updates to the content

I thought the same thing. Notion ticket here: https://www.notion.so/oaknationalacademy/Tech-debt-Create-slug-field-in-sanity-for-curriculum-overview-13726cc4e1b180df8395cc67e39c60fb?pvs=4

orangemug:
> Yeah, it had me stumped too...

But there is no `subjectSlug` still right? So is that don't a fuzzy match on all the fields or something?

sonalisin:
> > Yeah, it had me stumped too...
> 
> But there is no `subjectSlug` still right? So is that don't a fuzzy match on all the fields or something?

It's filtering just on the "Subject" field which is just the title 

`subject: { matches: $subjectSlug }`

rightsaidjames (approved):
Tested PE and RE - correct explainers are showing in the explainer tab and documents! Also regression tested some other cycle 1 and cycle 2 explainers to make sure that all of the key functionality was still working as before.

PR #2991, opened 2024-11-13
Add mobile lot picker modal

assadk88:
## Description

Music year: 1962

- Added a subject and phase picker modal which is only in use for mobile screen width
  - Note that the View button has been removed from the lot picker and is now incorporated in the Phase picker as a 'View curriculum'
- Fixed a small pre-existing bug where the subject icon colour wasn't inverted properly upon selection/focus
- Updated existing lot picker tests and added further tests for the mobile variants, also updated data-testids to consistently use kebab-case

## Issue(s)

Fixes #[782](https://www.notion.so/oaknationalacademy/Lot-picker-Mobile-modal-65bda90cfcb246f5a096014d4c31cc33?pvs=4)

## How to test

1. Go to https://deploy-preview-2991--oak-web-application.netlify.thenational.academy
2. Navigate to the `/teachers/curriculum`
3. Use Chrome DevTools to adjust the screen-width to a **mobile** preset
4. Interact with the mobile subject and phase pickers

## Screenshots

Video:

https://github.com/user-attachments/assets/e48e7868-c93d-455b-848a-a2b9433d668c

image
image

How it should now look:
image
image

## Checklist

- [x] Design sign-off
- [ ] ~~Does this PR update a package with a breaking change~~


HelenHarlow (commented):
Heya, looks great! I few things I spotted below:

- The sticky area should be up to the grey line, it seems to go a bit above 
Screenshot 2024-11-18 at 11 32 15
- The exam board buttons should have the same hover state as the other buttons, they seem to have an additional yellow drop shadow
- When using keyboard, the edge of the focus state seems to be clipped
Screenshot 2024-11-18 at 11 32 53
- It does seem to take a little while to load, so instead of the modal closing immediately when the user completes the flow, I wonder if we can make it so we stay in the modal and use the loading state of the primary button until the page has loaded?

src/components/SharedComponents/SubjectPhasePicker/SubjectPhasePicker.tsx
```diff
@@ -385,10 +536,36 @@ const SubjectPhasePicker: FC<SubjectPhasePickerData> = ({
     return title;
   };
 
+  const handleConfirmSubject = () => {
+    setShowPhases(true);
+    setTimeout(() => {
+      setIsMobileLotPickerModalOpen(false);
+    }, 50);
```
orangemug, line 543: Why do we need the `setTimeout` here?

src/components/SharedComponents/SubjectPhasePicker/SubjectPhasePicker.tsx
```diff
@@ -180,6 +187,109 @@ const SelectionDropDownBox = styled(Box)<object>`
...
+  showSubjectError,
+  onClick,
+}: SubjectContainerProps) {
+  const subjectErrorId = useId();
+  const subjectInputId = useId();
+  const isMobile = useMediaQuery("mobile");
+
+  return (
+    <Box
+      $maxHeight={["calc(100vh - 160px)", "auto"]}
```
orangemug, line 206: Since this is quite specialised now, rather than adding an additional type in https://github.com/oaknational/Oak-Web-Application/pull/2991/files#diff-98ee9ade28de4475b0c6b48173ce722b73bbc912e588c2fb751e8e2da0cfd117R139 maybe we can instead wrap in another `<div/>` or can we use flex to fill the vertical space.

src/styles/theme/types.ts
```diff
@@ -135,7 +135,8 @@ export type PercentSpacing =
   | "10%"
   | "5%"
   // vw units sometimes needed to fix edge cages
-  | "28vw";
+  | "28vw"
+  | "100vh";
```
orangemug, line 139: See note above

src/styles/utils/size.ts
```diff
@@ -5,7 +5,9 @@ import { NullablePixelSpacing, PercentSpacing } from "../theme/types";
 import responsive, { ResponsiveValues } from "./responsive";
 import { getRemUnits } from "./getRemUnits";
 
-type CalcValues = `calc(${PercentSpacing} - ${NullablePixelSpacing}px)`;
+type CalcValues =
+  | `calc(${PercentSpacing} - ${NullablePixelSpacing}px)`
+  | `calc(${PercentSpacing} + ${NullablePixelSpacing}px)`;
```
orangemug, line 10: Also see note above

orangemug (commented):
Mostly looks good, I'd like to keep away from adding additional spacing types if we can. I think we can probably use "flex"... we hopefully anyway.

src/components/SharedComponents/SubjectPhasePicker/SubjectPhasePicker.tsx
assadk88: I had initially added this for a better transition between the subject to phase picker, but you're correct it's no longer needed

rightsaidjames (approved):
Looks good, assuming that @HelenHarlow is happy with the design fixed! One tiny thing @assadk88 - please can the 'arrow-right' icon on the 'Confirm subject' and 'View curriculum' buttons be given an empty alt text value, as it's a redundant image next to the visible button text.

Screenshot 2024-11-19 at 17 07 37
Screenshot 2024-11-19 at 17 07 00

orangemug:
Ah wait I've got one for ya @assadk88. If the use hasn't accepted the cookies yet. The _cookie banner becomes un-clickable_ with the modal open, it also _covers the content_

Screenshot 2024-11-20 at 09 28 11
Screenshot 2024-11-20 at 09 27 36

rightsaidjames (changes requested):
Excellent spots @orangemug! And 🤦 to me for forgetting to test with the cookie consent widget. Firing the 'Request changes' klaxon...

HelenHarlow (approved):
Looks great from a design POV, thanks for making those updates!

orangemug (approved):

PR #3018, opened 2024-11-22
Fix/pupil 1119/share cookies to local storage

kimon-satan:
### Job Story

Switching from cookies to local storage to avoid hard limits and performance overheads

### Implementation details

Modify 

src/pages-helpers/teacher/share-experiments/createShareId.ts

to use local storage instead of cookies

update all tests relying on cookie mocks

### Acceptance criteria

- [x]  local storage used
- [x]  tests updated
- [x]  existing functionality uneffected

How to test
Go to https://deploy-preview-3018--owa-storybook.netlify.app/ and accept cookies and open the console
Go to  https://deploy-preview-3018--owa-storybook.netlify.app/teachers/lessons/atomic-structure-very-small-electron-mass
See that TeacherShareInitiated is sent
Copy the url into a different browser window (eg. a different google instance or safari)
See that both TeacherShareIniated and TeacherShareConverted are sent (accept cookies)
See that the url in the second browser is modified to a different id
Repeat the experiment with same urls and see that the events are not sent the second time
repeat for the various pages
try with feature flags disabled in either browser and note that conversions are still sent will initiated events are not

### DIFF
--- a/src/utils/localstorage.ts
+++ b/src/utils/localstorage.ts
@@ -39,6 +39,30 @@ export function getLocalstorageWithSchema<Schema extends z.ZodTypeAny>(
   return out!;
 }
 
+export function getNullableLocalstorageWithSchema<Schema extends z.ZodTypeAny>(
+  key: string,
+  schema: Schema,
+): z.infer<Schema> | undefined {
+  let value: unknown;
+  const valueRaw = localStorage.getItem(key);
+  if (valueRaw === null) {
+    return undefined;
+  }
+
+  try {
+    value = JSON.parse(valueRaw);
+  } catch (err) {
+    throw new ErrorInvalidJson(key);
+  }
+
+  const results = schema.safeParse(value);
+  if (!results.success) {
+    throw results.error;
+  }
+  const out = results.data as z.infer<Schema>;
+  return out!;
+}
+
 export function setLocalstorageWithSchema<
   Schema extends z.ZodTypeAny,
   DataType extends z.infer<Schema> | undefined,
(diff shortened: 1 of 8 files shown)


src/utils/localstorage.ts
```diff
@@ -39,6 +39,30 @@ export function getLocalstorageWithSchema<Schema extends z.ZodTypeAny>(
   return out!;
 }
 
+export function getNullableLocalstorageWithSchema<Schema extends z.ZodTypeAny>(
```
orangemug, line 42: Maybe we should remove the check from `getLocalstorageWithSchema` and just pass `undefined` to the schema, so something like (untested)

```ts
export function getLocalstorageWithSchema<Schema extends z.ZodTypeAny>(
  key: string,
  schema: Schema,
): z.infer<Schema> {
  let value: unknown;
  const valueRaw = localStorage.getItem(key);
  if (valueRaw !== null) {
    try {
      value = JSON.parse(valueRaw);
    } catch (err) {
      throw new ErrorInvalidJson(key);
    }
  } else {
    value = undefined;
  }
  const results = schema.safeParse(value);
  if (!results.success) {
    throw results.error;
  }
  const out = results.data as z.infer<Schema>;
  return out!;
}
```

The we can allow the schema to accept `null` as a value

orangemug (approved):
Looks good except the one suggestion

src/utils/localstorage.ts
kimon-satan: Yeah maybe . I just didn't want to rock the boat especially as this is all temporary until we get our user data api from planning squad

RBrownsett (approved):
Tested - no issues

PR #3028, opened 2024-11-26
Huge image on homepage affecting page performance

orangemug:
## Description
Fix to reduce huge image size on homepage

> [!NOTE]  
> Size in DOM are based on pixel ratio of your browser so might differ 

Reduction

```
243kb - 64kb = 179kb (savings)
```

## How to test

1. Go to https://deploy-preview-3028--oak-web-application.netlify.thenational.academy
2. Check image is of a decent size at all the breakpoints

## Screenshots
Before:
Screenshot 2024-11-26 at 17 53 36

After:
Screenshot 2024-11-26 at 17 53 55


sonalisin (approved):
Looks good to me!

rightsaidjames (approved):
Belatedly approving this because it does what it says. I think we should look into the mobile performance, as it's loading in an image that is a larger size that of the desktop or tablet equivalent, which is then not actually shown to the user 😄

orangemug:
Closing PR as outdated

weronika-szalas (approved):

PR #3029, opened 2024-11-26
feat: remove wobbly border from curriculum year jump buttons and add fade effect

assadk88:
## Description

Music year: 2010

- Remove the wobbly border from the year group buttons
- Added a fade effect to horizontal overflow of the button group
- Fixed scroll to year section scrolling ending prematurely bug

## Issue(s)

Fixes #[1038](https://www.notion.so/oaknationalacademy/Remove-wobbly-borders-from-year-jump-buttons-13b26cc4e1b1809fbca8e6bda2c6a1a5?pvs=4)

## How to test

Test using mobile and:

1. Go to https://deploy-preview-3029--oak-web-application.netlify.thenational.academy
2. Navigate to Science > KS3 & 4 > Select an exam board
3. Click through the years to see if the jump to section functionality works correctly
4. Scroll year button group horizontally to see if the fade effect works correctly

## Screenshots

Video:

https://github.com/user-attachments/assets/7427fc40-78d4-470a-9aab-fd56d669a053

🪧 In implementing the fade effect on horizontal overflow I hope that it doesn't matter that the effect also applies to the scroll bar. I tested this on my iOS device and the scroll bar does not appear when scrolling left or right on it, hence the effect does not apply to the scroll bar – meaning this is just a moot issue when we're emulating mobile on Chrome.

image

How it should now look:
image

## Checklist

- [ ] ~~Added or updated tests where appropriate~~
- [X] Design sign-off
- [ ] ~~Does this PR update a package with a breaking change~~


HelenHarlow (commented):
Only change from me is to reduce the spacing between the buttons to 8px. Otherwise looks good!

sonalisin (approved):

src/components/CurriculumComponents/CurriculumVisualiserFilters/CurriculumVisualiserFiltersMobile.tsx
```diff
@@ -105,43 +179,60 @@ function StickyBit({
...
+                          const targetElement = document.getElementById(
+                            `year-${yearOption}`,
+                          );
+                          if (targetElement) {
+                            targetElement.scrollIntoView({
+                              behavior: "smooth",
+                            });
+                            const yearHeading = document.getElementById(
+                              `year-${yearOption}`,
+                            ) as HTMLElement;
```
sonalisin, line 220: Casting this as an HTML element may not be the best approach, I think it'll probably throw an error if it expects that element may not exist and then the undefined will be cast as a HTMLElement. A conditional might be better, e.g. having a `if(yearHeading && instanceof yearHeading ===...`

orangemug (approved):

rightsaidjames:
Incredible work @assadk88! The only improvement I would suggest would be to add `role="group"` to the container div with the `aria-label="Select a year group"` property, although VoiceOver on macOS seems to work out that these buttons are part of a group even without this role.

- [x] Year group buttons have accessible names that match their visible labels (including Swimming all-years)
- [x] `aria-pressed` is set to true/false based on the currently active year
- [x] Buttons are reported as part of a group called 'Select a year group'
- [x] Selecting a year group scrolls the current year into view
- [x] Pressing tab after selecting a year group takes you to the first unit of that year
- [x] Works well with VoiceOver on macOS (Chrome) and iOS (Safari)

rightsaidjames (approved):

src/components/SharedComponents/ButtonGroup/ButtonGroup.tsx
```diff
@@ -16,7 +16,11 @@ const ButtonGroup: FC<ButtonGroupProps> = (props) => {
    */
   const { children, ...flexProps } = props;
 
-  return <Group {...flexProps}>{children}</Group>;
+  return (
+    <Group role="group" {...flexProps}>
+      {children}
+    </Group>
```
assadk88, line 22: As per @rightsaidjames suggestion I've added `role="group"` – to do this I've hardcoded `role="group"` into the `ButtonGroup` component, which should not be problematic as it's implied by the semantics inferred from its name

rightsaidjames:
@assadk88 Mostly good - the group role is being reported in DevTools. It's not being announced by iOS VoiceOver, but we can't (and shouldn't) control that, and it is likely controlled by a user preference or an iOS-specific heuristic.
Screenshot 2024-12-09 at 12 58 20

Since I last tested, the skip links on iPhone are jumping to just above the selected year, which means that you need to tap the year button twice to get it to jump to the right place:
Screenshot 2024-12-09 at 12 56 11
Screenshot 2024-12-09 at 12 56 20

rightsaidjames (approved):
Retested, fix confirmed, LGTM!

orangemug (approved):
LGTM

PR #3046, opened 2024-12-02
Feat/lesq 1185/unit downloads [LESQ-1185]

njblore:
## Description
This combines LESQ-1185 and LESQ-1196

- Add a UnitDownloadButton component with states for feature flag variants, signed in status and download status 
- Create helper fns for unit downloads similar to lesson download fns

The experiement for this can be seen [here](https://eu.posthog.com/project/124/experiments/28199) on posthog

## How to test

1. Go to https://deploy-preview-3046--oak-web-application.netlify.thenational.academy/teachers/programmes/geography-primary-ks1/units/local-area-where-do-we-go-to-school/lessons
2. Which button you see will depend on which cohort of the experiment you have been placed in, you need to add yourself to [this feature flag](https://eu.posthog.com/project/124/feature_flags/43245) in either group (without adding yourself you will be placed in the control group)
3. In the control group, or if the unit download does not exist, the button should not show
4. When signed out: clicking the unit download button should navigate you to the sign in journey, then redirect to this page
option a:
Screenshot 2024-12-02 at 12 14 15  

option b:
Screenshot 2024-12-02 at 14 14 54  

5. When signed in: clicking the button should trigger the download to start, the button will update with a loading spinner and a message will show
Screenshot 2024-12-02 at 12 14 32  
Screenshot 2024-12-02 at 12 14 41  

7. An error state should show when there is an error
Screenshot 2024-12-02 at 12 15 59


dkyyau (approved):

jack-skerrett-oak (approved):
I think we go with what we have for now. There are some fixes that will be done in future tickets, but getting unit downloads so we can test it behind the feature flag in prod is the most important thing for now.

PR #3096, opened 2025-01-08
fix: adjust logic for subject categories

assadk88:
## Description

Music year: 2020

- Adjusted the `enableGroupBySubjectCategory` logic so that so that it checks all units instead of just the first one and enables subject category grouping if ANY unit has the property
- Added unit test for cases where the the first unit does not have the `group_by_subjectcategory` enabled

## Issue(s)

Fixes #[1156](https://www.notion.so/oaknationalacademy/Fix-bad-logic-in-docx-subject-categories-16026cc4e1b1808dbe9ce18e36564eef)

## How to test

1. Check the snapshot output under the JSON representation for the `enables grouping when any unit has group_by_subjectcategory feature` and `disables grouping when no units have group_by_subjectcategory feature` test cases

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist


orangemug (approved):
LGTM!

PR #3102, opened 2025-01-09
fix: adjust z-indexes for pop-up survey

assadk88:
## Description

Music year: 2001

After accepting cookies the PostHog survey appears, and if the Subject or Phase Picker modals were clicked in the intervening time this would then be overlayed on them but would be un-interactive. It was decided that instead of engineering a solution that would mean that survey was interactive above the modal, that its z-index should be reduced so that it isn't visible above the modal to begin with.

## Issue(s)

Fixes #[1165](https://www.notion.so/oaknationalacademy/3bccfe060e4645fe97842977e365c2ad?v=d9869bab5f8a4e28a7a960d37f6ab31e&p=17426cc4e1b180d184b5c34608e88846&pm=s)

## How to test

1. Go to https://deploy-preview-3102--oak-web-application.netlify.thenational.academy/teachers/curriculum in `Incognito` mode
2. Once there accept the cookies and wait briefly for the survey to appear
3. Once the survey appears, switch to Mobile/Responsive viewport and click the Phase Picker which will cause it's associated modal to be displayed
4. The PostHog survey should **not** be visible
5. Close the modal to see that the survey is indeed still there

https://github.com/user-attachments/assets/2e34b93e-ab2c-4615-8606-fb4368b57e08

## Screenshots

image

How it should now look:
image

The survey now also appears behind the cookie banner:

image

## Checklist

- [X] Design sign-off


src/styles/posthog-hack.styles.ts
```diff
@@ -0,0 +1,13 @@
+import { css } from "styled-components";
+
+const posthogHackStyles = css`
+  /* 
+   * Styles for Posthog survey
+   * These styles help manage z-index conflicts with PostHog surveys
+   */
+  div[class^="PostHogSurvey"] {
+    z-index: 298 !important;
```
orangemug, line 9: Can we not set this to `z-index: 299;` and remove the other `z-index` changes in this PR?

src/styles/posthog-hack.styles.ts
```diff
@@ -0,0 +1,13 @@
+import { css } from "styled-components";
+
+const posthogHackStyles = css`
+  /* 
+   * Styles for Posthog survey
+   * These styles help manage z-index conflicts with PostHog surveys
```
orangemug, line 6: Add link to ticket describing issue here (notion page)

src/styles/posthog-hack.styles.ts
assadk88: image

We could remove the one for the cookie banner, however the one for the label would need due to issue in the above image.

src/components/SharedComponents/SubjectPhasePicker/SubjectPhasePicker.tsx
```diff
@@ -599,7 +599,7 @@ const SubjectPhasePicker: FC<SubjectPhasePickerData> = ({
         $color={"text-primary"}
         $font={"heading-7"}
         label={"Choose a curriculum"}
-        $zIndex={"banner"}
+        $zIndex={298}
```
orangemug, line 602: Is there a labeled `zIndex` we can use here? Tend to want to stay away from numbers in general.

orangemug:
@rightsaidjames (also @assadk88) the fix for this will potentially effect any survey presented to folk across the entire website, so I guess we'll need other squads to test for https://github.com/oaknational/Oak-Web-Application/pull/3102#discussion_r1909155601

src/components/SharedComponents/SubjectPhasePicker/SubjectPhasePicker.tsx
assadk88: There doesn't seem to be one for 298

src/styles/posthog-hack.styles.ts
orangemug: Discussed with @assadk88 this is fine

src/components/SharedComponents/SubjectPhasePicker/SubjectPhasePicker.tsx
orangemug: Discussed with @assadk88 this is fine

orangemug (approved):

rightsaidjames (approved):
Tested this fix in mobile view (and also on an iPhone) against the subject-phase picker, as well as all of the other modal/pop-up interfaces I'm aware of (hamburger nav, cookie consent dialogs, unit modal, sticky navs, hints on pupil quizzes). All good, LGTM!

PR #3105, opened 2025-01-13
[JT-8] Chore/update node support

jimCresswell:
## Description

- Updated Node to current LTS, 22.x .
- Also added some words from the codebase to the workspace settings to reduce the VSCode `problems` reported to 1 (which we should also fix).

Hello everyone 👋 I don't have any context for what has been happening, so ignore me if there is a reason we haven't updated to latest LTS yet.

I also may not check back in GitHub for a while, so if someone would like to champion this through I would really appreciate it.

I've upped the required npm version, because have mismatched major version there could cause all sorts of issues. Once this is merged everyone will need to update to latest Node LTS, e.g. `nvm install LTS`.

If building this PR in Netlify causes issues with other builds, it's because Netlify has cached some dependencies overly generally, and the differing major versions of npm means cache from one build breaks the other build. The "fix" is to clear the cache and rebuild, but really the way to stop it is to get everyone using the same version of Node/npm on all checkouts. Hopefully they fixed that issue in the last six months.

I'm looking forward to being properly back!

## Testing

CI/CD works... that's probably about it.


mantagen (approved):
Code looks good, and CI is passing.

I'm not aware of a decision to not update Node

jimCresswell:
@mantagen if there are build issues after this goes in please see the notes at the top for solutions

PR #3114, opened 2025-01-13
Remove old `/develop-your-curriculum` page that's no longer used

orangemug:
## Description
Removes the `/develop-your-curriculum` page which hasn't been used since the prior to cycle-1 release 

## Issue(s)
Fixes `CUR-1176`

## How to test
Double check that the redirect <https://thenational.academy/develop-your-curriculum> still works once deployed


rightsaidjames (approved):

src/components/ArchivedComponents/Home/HomeSiteCards.tsx
```diff
@@ -21,15 +21,15 @@ const HomeSiteCards = () => {
         />
       </OakGridArea>
       <OakGridArea $transform={["translateY(50%)"]} $colSpan={[12, 4]}>
-        <CardLinkIcon
+        {/* <CardLinkIcon
```
sonalisin, line 24: Is this supposed to be commented out or removed?

sonalisin (changes requested):
Looks good other than some commented code :)

src/components/ArchivedComponents/Home/HomeSiteCards.tsx
orangemug: Yeah these will be removed shortly as they are in `ArchivedComponents` so I didn't want to deal with the type issues. I think it's fine as it's not used anywhere, and will be removed shortly (this week I believe)

PR #3147, opened 2025-01-23
CTC-124 Handle pathways on subject listing page

weronika-szalas:
## Description

- Added pathways on subject listing page
- Created new component to be displayed when there are pathways available 

As discussed with @aakeshp  the programmes  with pathways should include:
- GCSE + all legacy units for gcse pathway
- Core + all legacy from Core pathway

The unit count on subject listing page should reflect that.

## Issue(s)

Includes work for [CTC-124](https://www.notion.so/oaknationalacademy/OWA-handle-pathways-on-subject-listing-page-17c26cc4e1b1802588adf9086bb1afe3)

and [CTC-123](https://www.notion.so/oaknationalacademy/OWA-SubjectListingCard-to-handle-pathways-17c26cc4e1b180b4b238f689211a8ee7)

## How to test

1. Go to https://deploy-preview-3147--oak-web-application.netlify.thenational.academy
2. Go to `/teachers/key-stages/ks4/subjects`
3. Citizenship should have two cards with pathways visible


messmike:
Thanks, @weronika-szalas. Just a few styling updates for the card component are needed:

- The height of the split card subhead background container is a bit tall (I have it using `all-spacing-7` in the [design kit component in Figma](https://www.figma.com/design/YcWQMMhHPVVmc47cHHEEAl/Oak-Design-Kit?m=auto&node-id=9563-17505):
Screenshot 2025-01-27 at 11 48 38 am
In the design kit, there are also some updates to the states of the components (single link and double link card) that were made, which would be good to get updated as part of this (or a subsequent ticket if needed).
- On mobile, the card headings are centre-aligned, can we left-align them as in the designs? 
Screenshot 2025-01-27 at 11 54 41 am

RBrownsett:
@messmike just looking at the designs, is the plan to show a pathway card even if there is only one pathway? Example below of RE from the designs
image

@weronika-szalas I believe KS4 PE is a pathway subject so if the answer to the above is yes then we would need to show the card? Would that be difficult to implement?

messmike:
> @messmike just looking at the designs, is the plan to show a pathway card even if there is only one pathway? Example below of RE from the designs

Yes, if there is a single pathway (like RE KS4), then it would be good to be able to notify the user with the subheading, but it would just be a single link.

RBrownsett:
@weronika-szalas is it possible to ensure that the pathway is readout via the screen reader by adding it to the aria-label? Currently when using the screen reader a visually imparied user would not know the difference between the two paths as they are read out the same. See example below, I have highlighted GCSE but this is not fed back to the user:
 
image

aria-label="Citizenship: 14 units, 86 lessons - new content"

RBrownsett:
@weronika-szalas will there be a separate ticket to cover the subject titles on the unit listing page?

Design:
image

Current PR (missing the text 'Core'):
image

weronika-szalas:
@messmike @RBrownsett I've implemented all PR feedback except states of the components which I will update in a separate ticket (https://www.notion.so/oaknationalacademy/OWA-Update-states-of-the-components-18a26cc4e1b1806e9892f297a6a975bd)

@RBrownsett for now Citizenship doesn't show legacy subjects and computing is split into two subjects (same as on live site). This will be fixed once we make update to the database. This ticket needs to go out first though.

kimon-satan (approved):

RBrownsett (approved):
Approved - changes made. This will re-tested once the DBT Tools PR has been merged to ensure the legacy lessons are pulled through and computing non GCSE are shown on a single subject card

PR #3155, opened 2025-01-29
Feat/pupil 1142/teacher share events

kimon-satan:
### Job Story

**When** someone saves and shares a teacher note,
**I want to send tracking information**
**So that I can understand how users are using the feature**.

### Implementation details

using the avo branch teacher-notes

## How to test

1. Go to https://deploy-preview-3155--oak-web-application.netlify.thenational.academy/teachers/lessons/transverse-waves

- open the teacher note dialogue and see that "Teacher Note Dialogue Opened" is sent
- write enough text that a save is triggered and see that "Teacher Note Saved" is sent
- note that src in the url is now 5 (signifying canonical lesson with teacher note)
- click share button and note that Share Activated is sent with note length as one of the properties
- note that this event is only sent once

### Acceptance criteria

- [x]  TeacherNoteDialogueOpened
- [x]  TeacherNoteSaved
- [x]  Share Activated on Teacher Share button pressed
- [x]  Data to identify that this is a teacher note  share
- [x]  Note length recorded 
- [x]  Prevent duplicate Activations


RBrownsett:
@kimon-satan works as detailed above. A couple of questions:

1. Share Activated only fires when the lesson is from the canonical url - is that correct?
2. If a user has already clicked Share but then edits the note do we want to fire the Share Activated event again? Currently it does not, and to be fair its most probably an edge case but I thought I would flag it if we are tracking note length (which would likely change in this scenario)

kimon-satan:
> @kimon-satan works as detailed above. A couple of questions:
> 
> 1. Share Activated only fires when the lesson is from the canonical url - is that correct?
whoops! fixed in the update I just pushed
> 2. If a user has already clicked Share but then edits the note do we want to fire the Share Activated event again? Currently it does not, and to be fair its most probably an edge case but I thought I would flag it if we are tracking note length (which would likely change in this scenario)
this is intentional and matches the previous tracking on simple share. Its true that the note length might not be readable here but we also get this in Teacher Note Saved . Having a single activation event makes things easier when working with posthog.

RBrownsett (approved):
Teacher share now works for non-canonical links. Tested with no issues

PR #3183, opened 2025-02-05
Restyle subject category filters

assadk88:
## Description

Music year: 1966

- Add HR elements between each filter section
  - @rightsaidjames Note that the custom HR component has been removed from the codebase in our attempt to standardise components and try to use components from the component library where possible. As such instead of the `Hr` component being used I had to use `OakHandDrawnHr`. Due to this the divider has a slightly different style and could only take specific margin-top and bottom spacing values – which therefore meant that it diverges very slightly in look from the Figma designs. I talked with @HelenHarlow about this and she was fine with the changes.

Note that there seems to be a slight bug with the filter at the moment, where subject categories and exam subjects aren't displayed at the same time when all is selected (they're only conditioanlly displayed if you click into their associated years):

https://github.com/user-attachments/assets/c8bf8e90-acc0-435c-b714-fd179eaef6f1

Ignore this for the mean time, as we'll get around to fixing this as part of work in another ticket.

## Issue(s)

Fixes #[1172](https://www.notion.so/oaknationalacademy/Initial-defaults-in-filtering-17d26cc4e1b180929f2ae7c4a1d7e890?pvs=4)

## How to test

1. Go to https://deploy-preview-3183--oak-web-application.netlify.thenational.academy/teachers/curriculum
2. Select a subject from the curriculum visualiser
3. You should see that filter sections are now visible on the left, with an appropriately spaced separator in between each filter section

## Screenshots

image

How it should now look:
image

## Checklist

- [x] Design sign-off

### DIFF
--- a/src/utils/getValidSubjectIconName.ts
+++ b/src/utils/getValidSubjectIconName.ts
@@ -1,11 +1,18 @@
 import { OakIconName, isValidIconName } from "@oaknational/oak-components";
 
+import { getValidSubjectCategoryIconById } from "./getValidSubjectCategoryIconById";
+
+import { SubjectCategory } from "@/utils/curriculum/types";
 import errorReporter from "@/common-lib/error-reporter";
 import OakError from "@/errors/OakError";
 
 export const getValidSubjectIconName = (
-  subject: string | null,
+  subject: string | null | number | SubjectCategory,
 ): OakIconName => {
+  if (subject && typeof subject === "object" && "id" in subject) {
+    return getValidSubjectCategoryIconById(subject.id) as OakIconName;
+  }
+
   const subjectIconName = `subject-${subject}`;
   if (!isValidIconName(subjectIconName)) {
     // Subject will be null when a canonical lesson appears in multiple subjects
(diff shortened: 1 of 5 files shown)


HelenHarlow (commented):
Styling looks great! I think we're just missing the icons on the primary english categories

src/utils/getValidSubjectIconName.ts
```diff
@@ -1,11 +1,18 @@
...
+import { SubjectCategory } from "@/utils/curriculum/types";
 import errorReporter from "@/common-lib/error-reporter";
 import OakError from "@/errors/OakError";
 
 export const getValidSubjectIconName = (
-  subject: string | null,
+  subject: string | null | number | SubjectCategory,
 ): OakIconName => {
+  if (subject && typeof subject === "object" && "id" in subject) {
+    return getValidSubjectCategoryIconById(subject.id) as OakIconName;
```
orangemug, line 13: Rather than pollute this function, I'd do it the other way around instead. So keep this "as-is" because it's a well defined function and we're adding a bit of a hack.

Make `getValidSubjectCategoryIconById` call `getValidSubjectIconName` instead.

orangemug (approved):

PR #3189, opened 2025-02-10
fix: hubspot email in localstorage [LESQ-1255]

njblore:
## Description

Music year: 1985

- always set signed in user's email in localstorage after onboarding
- always update user's email in localstorage when found in hubspot during resource download

## Issue(s)

Fixes #
users' emails not being sent with hubspot submissions if  they accepted email marketing during onboarding

## How to test

1. Go to https://deploy-preview-3189--oak-web-application.netlify.thenational.academy
2. Create a new account and do not subscribe to emails during onboarding (the email you use must not have been subscribed to emails before in hubspot, deleting in hubspot does not reliably remove all data)
3. After onboarding your email should be set in localstorage as `oak-downloads-email`
4. Go to download a lesson
5. after page load localstorage should still contain your email
6. On download the hubspot form submission should include your email in the fields array 
7. The above should all also be true if you subscribe to email during onboarding


dkyyau (approved):

left23 (approved):
Hi @njblore This works as described in the testing steps. I can see my email being set in localstorage and it is included in the hubspot form submission.
If a user is already subscribed to hubspot, do you anticipate different results that we should consider?

njblore:
@left23 good point if a user is already subscribed the results should be the same, the only difference would be during onboarding they won't see the checkbox to subscribe (in reality the hubspot workflow does not appear to be set up properly to subscribe people during onboarding on the dev account, so this may be tricky to test, but is not an area of concern right now)

PR #3195, opened 2025-02-11
Parallelize CI a little more

orangemug:
## Description
This should hopefully speed up CI a little

We're just moving these checks off into their own job and running checks(lint, check types) and tests(test,sonar) in parallel. This should save us ~1min on the test run 🤞

Screenshot 2025-02-11 at 16 43 56

### DIFF
--- a/.github/workflows/code_checks.yml
+++ b/.github/workflows/code_checks.yml
@@ -12,8 +12,8 @@ on:
   merge_group:
 
 jobs:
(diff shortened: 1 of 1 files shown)


BTitterington (approved):
Maybe get 2 reviews before merging. But it looks good to me.

kimon-satan (approved):

PR #3199, opened 2025-02-12
New larger github runner

orangemug:
## Description
Move over to new larger GitHub runners in CI to reduce time waiting

Total time ~**8m 20s** (4-core) down to

 - ~**4m 20s** (8-core)
 - ~**3m 50s** (16-core)

We can now adjust the larger runners based on the runner group in GitHub called `large-runners`

## How to test
Just check CI works as expected (it does!)

## Screenshots
4-core (current)
Screenshot 2025-02-12 at 11 21 07

8-core
Screenshot 2025-02-12 at 11 20 14

16-core
Screenshot 2025-02-12 at 11 32 38


stevegreenn:
@simonrose121 is the runner configured at organisation level or repository? If the former, is it OK to leave it generically named?

As a general point, I think it's worth documenting the existence of the group somewhere.

stevegreenn (approved):

simonrose121:
> @simonrose121 is the runner configured at organisation level or repository? If the former, is it OK to leave it generically named?
> 
> As a general point, I think it's worth documenting the existence of the group somewhere.

Organisation level but I've only given it access to this repo, we can also restrict the workflows it can run in.

PR #3207, opened 2025-02-13
CTC-142 Don't round video duration to zero

weronika-szalas:
## Description

Don't round video duration to zero - round to two coma spaces instead

## Issue(s)

Fixes [CTC-142](https://www.notion.so/oaknationalacademy/Media-Page-Bug-Clips-less-than-a-second-do-not-play-18a26cc4e1b18093900ae93894a689ee)

## How to test

1. Go to https://deploy-preview-3207--oak-web-application.netlify.thenational.academy/teachers/programmes/spanish-secondary-ks4-edexcel/units/media-and-technology-actores-y-peliculas/lessons/dos-actores-famosos-ar-3rd-person-singular-and-plural-preterite/media
2. The videos under 1 second should play

### DIFF
--- a/src/components/SharedComponents/VideoPlayer/getDuration.test.ts
+++ b/src/components/SharedComponents/VideoPlayer/getDuration.test.ts
@@ -12,9 +12,9 @@ const createRef = (duration: number | undefined) => {
 };
 
 describe("getDuration", () => {
-  test("gets duration rounded down", () => {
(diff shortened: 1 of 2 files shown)


weronika-szalas:
@RBrownsett the duration is set to null in database (for the other issue we were reading it from video object itself but we can't do it here) so we would need to overwrite it in exceptions table or amend this data... will open another ticket if we wanted to handle it in the future

PR #3208, opened 2025-02-14
Fix/pupil unit listing issue

kimon-satan:
## Issue(s)

When overrides and exceptions cause only legacy units to be available for a pupil unit listing page the page 404s. 

Fixes #

Instead get it to redirect to the legacy unit listing page. 

## How to test

1. Go to https://deploy-preview-3208--oak-web-application.netlify.thenational.academy/pupils/years/year-1/subjects
2. navigate to PE
3. you should arrive at the legacy listing


RBrownsett (approved):
Approved - checked that PE displays. Also checked other legacy / non legacy subjects show

PR #3217, opened 2025-02-19
Jest performance logging

orangemug:
## Description
See the [blog post](https://www.notion.so/oaknationalacademy/Jest-performance-logging-19f26cc4e1b180539985f4c9b1740a95) for a full rundown of the CLI + problems

## How to test
Try out the CLI using the instructions outlined in the [post](https://www.notion.so/oaknationalacademy/Jest-performance-logging-19f26cc4e1b180539985f4c9b1740a95) 

## Examples
Running the following to generate the data

```
./scripts/dev/test/automate.ts timings-test
```

Then inspecting it with the CLI
 
```
./scripts/dev/test/automate.ts timings-report --min 300 --path curric          
```

Which will output

Screenshot 2025-02-19 at 09 09 11

PR #3219, opened 2025-02-19
Speed up tests by adding missing key and minimising providers

orangemug:
## Description
On from #3217 I've fixed the performance of some slow tests

It turns out that two things were causing slow downs, firstly whenever an error gets logged it needs to calculate a stacktrace. Fixing this seems to speed up by ~30-50ms

Screenshot 2025-02-19 at 07 20 12

Secondly reducing the amount of providers included in tests with the following new function was a minor small bump.

```ts
const render = renderWithProvidersByName(["oakTheme", "theme"]);
```

Results are about 200ms saved on this test file alone (adds up quickly)

Screenshot 2025-02-19 at 13 49 48

## How to test
Just added a missing key to the SubjectPhasePicker

1. Go to https://deploy-preview-3219--oak-web-application.netlify.thenational.academy/teachers/curriculum
2. Check phase picker works as intended


assadk88 (approved):

rightsaidjames (approved):
Did a quick regression test on the subject-phase picker, all good.

PR #3222, opened 2025-02-20
Fix/testmonials image url

kimon-satan:
## Issue(s)

OWA throws invalid src prop , because testimonials component doesn't proxy sanity image src.

Fixes #

Refactor code from CMSImage so that Testimonials can also proxy sanity cdn url.

## How to test

1. Go to https://deploy-preview-3222--oak-web-application.netlify.thenational.academy
2. Click through the testimonials


sonalisin (commented):
Are there supposed to be images in the testimonials? I can just see the quotes / Oak icon

njblore (approved):
errors gone ✨ is there somewhere obvous we could put this as a helper fn so people know to use it with CMS images?

PR #3225, opened 2025-02-20
chore: add new component type to avo and link it to the sign posting …

tomwisecodes:
…to aila

Adds new component type and links up to the trigger


src/components/TeacherComponents/NoSearchResults/SignPostToAila.tsx
```diff
@@ -45,6 +48,13 @@ const SignPostToAila = ({
       buttonIconName="external"
       href={composeAilaLink({ keyStage, subject, unitTitle, searchExpression })}
       videoPlaybackID={videoPlaybackID}
+      onClick={() => {
+        track.lessonAssistantAccessed({
+          product: "ai lesson assistant",
+          isLoggedIn: false,
```
emmalanches, line 54: Maybe not one for now, but this one eventually be able to be made dynamic by looking at whether the user is already logged in? Just wondering if this would be possible/how complicated it would be

tomwisecodes: Yes exactly, once logins are in affect in owa it will by dynamic as it is in aila.

PR #3100, opened 2025-01-09
New curric visualiser filtering base branch

orangemug:
## Description
A base branch for vew curric visualiser filtering

## Issue(s)
Fixes `CUR-1169`

## How to test

1. Go to https://deploy-preview-3100--oak-web-application.netlify.thenational.academy
2. Click on \_\_\_\_\_\_
3. You should see \_\_\_\_\_\_

PR #3234, opened 2025-02-24
Organising fixtures

orangemug:
## Description
Reorganising fixtures, description in https://www.notion.so/oaknationalacademy/Fixtures-1a426cc4e1b18030bd60ce9633353211

Any feedback from squads please either add in this PR of the comments on the notion page.

## How to test
Only additional fixtures generators have been added. We'll start to actually use them in #3100 over the coming week.


njblore (approved):
looking good! thinking about how we can document & expose these - what about using a barrel file to namespace the exports? eg. `export * as Shared from ./shared; export * as Curriculum from ./curriculum`, then to use it you would use `Shared.createThread()` or `Curriculum.createThread()`?

BTitterington (approved):

orangemug:
> thinking about how we can document & expose these - what about using a barrel file...

@njblore I _think_ we may want to move away from barrel files more in the tests, as I believe it's causing performance issues by importing _all the things_ when we only need _one thing_. I need to confirm this though it's only a hunch at the moment.

sonalisin (approved):

PR #3244, opened 2025-02-27
Feat/lesq 1282/posthog alias update [LESQ-1282]

njblore:
## Description

Music year: 1962

- Update the id we send with `identify` events at sign up/in to be the user's clerkId instead of posthog distinctId (which is not consistent or unique to a user)
- Update the posthog analytics service with an `alias` function 
- Call `alias` when we track user sign up or in with the posthog distinctId as the alias value
- We need to do this manually becase Avo doesn't have the `alias` function as a supported action, otherwise we would want to include it as part of these events like the `identify` call

## Issue(s)

Fixes #
Users not being correctly reidentified when signing in again after signing up (particularly if from a different browser)

## How to test

1. Go to https://deploy-preview-3244--oak-web-application.netlify.thenational.academy 
2. Get your distinctId from the console
3. Go to [posthog people view ](https://eu.posthog.com/project/124/persons)and search your distinctId, you should see a single ID in the IDs list at the top and it should be your distinct ID  
Screenshot 2025-02-28 at 08 56 28   

4. Go back to OWA and sign up for a new account (you can do this on the Unit Listing page by clicking 'Download unit')
5. Go back to posthog and refresh your person view, you should now see another ID in the list beginning with `_user`, this is your clerk ID  
Screenshot 2025-02-28 at 08 59 16  

6. Now go to another browser and sign in to the account you just created
7. In posthog you should see a new 'Identify' action
8. Take some more actions on OWA and you should see them come through in posthog tied to your account


left23 (approved):
This looks good, works as described. Approving.

PR #3266, opened 2025-03-10
Added test logging report for FEC hack day

orangemug:
## Description
This is a report of logging created by our tests, so we can work through as a team and remove all the logging.

It works by capturing the logging from stdout/stderr and parsing it with some regexps.

## How to test
Just run the script locally and check the report output, so checkout this branch and run

```
./scripts/dev/test/automate.ts test-logging-report
```

Then open `./reports/jest-logging-issues.html` and you should see something like

Screenshot 2025-03-10 at 10 09 36

Note the tests will run a little slower while generating the report because we use `--runInBand` to run them serially.

You can open each `<detail/>` so see the logging. Any without tags are tests without logging (so are all good)


njblore (approved):
very nice!

PR #3282, opened 2025-03-12
Fixes incorrect unit count in hidden aria DOM node

orangemug:
## Description
Fixes for the unit count in hidden aria DOM node in the curriculum filters

Also small refactor to use `createFilter` helper

## Issue(s)

Fixes `CUR-1169`

## How to test

2. Check the accessible highlighted/unit count is correct when different combinations of filtering options are applied

## Screenshots
Screenshot 2025-03-12 at 16 17 21


rightsaidjames (approved):
Confirmed as fixed, thanks!

assadk88 (approved):

PR #3295, opened 2025-03-13
Store curric filter in the URLs querystring

orangemug:
## Description
Initial branch of storing the curriculum filter in the URLs querystring

**Note**: This doesn't "build time" / "server side", render which is inline with other squads, for example you'll note that https://www.thenational.academy/teachers/programmes/art-primary-ks1/units?year=year-2&learning-theme=draw gets a flash of "initial state" prior to JavaScript kicking in.

## Issue(s)

Fixes `CUR-1294`

## How to test

2. Check that filtering modifies the querystring in the correct way
3. Check that refreshing the browser, or opening the URL in a different browser restores the correct state


rightsaidjames (approved):
Tested locally, looks good. Risk is pretty low going forward, as we can just disable the const if we need to.

sonalisin (approved):

PR #3308, opened 2025-03-17
Typo in test when calling `.completeActivity(...)`

orangemug:
## Description
I believe this should always be calling

```diff
- result.current.completeActivity("intro")
+ result.current.completeActivity(section)
```

## How to test
@benprotheroe can you confirm the logic in the test


benprotheroe (commented):
looks good

PR #3311, opened 2025-03-18
Spike/Experiment: Update OWA to Next 15

jimCresswell:
This is an experiment to gauge how much work it would be to update OWA to Next 15. I've treated it like real work, so if we choose to, this PR will be merge-able. Alternatively we could adopt it piecemeal, but because of the interrelatedness of the dependencies that might be challenging.

It also updates Styled-Components to v6, which natively supports Typescript, that was a significant change, and I had to refactor the `responsive` style utility function.

Originally it also attempted to update to React 19, but that broke a lot of dependencies, and for now Next 15 still supports React 18. 

I have brought most of the tooling up to the latest versions. Excitingly that means both the Next config and the ESlint config can now be Typescript files and we can know what valid config looks like in the editor.

I found that the old ESLint rules for imports were not actually being applied. I've had to refactor the config considerably to match the new format, and I have temporarily disabled those rules, because fixing them touches _all of the files_. I have also added complexity linting rules, but set values that allow the current code. The ESLint rules should be reviewed, it's interesting which ones I have had to disable to get linting to pass without making this PR a lot larger than it is.

I also added a `.agent/` directory, for storing plans, prompts and summaries for LLM based tooling like Cursor or Claude Code.

**Current status**: type check passing, linting passing, tests passing, app builds. There are quite a few issues apparent in the console at run time, but the app is rendering.

I think the PR is ready for initial feedback, and because it touches most areas I think we need agreement from all of the squads before we merge. I fully expect there to be more work to do before everyone is happy.


jest.setup.mjs
kimon-satan, line 1: did you see the benchmarking work on switching to vitest. This upgrade might be a good opportunity to do this work too

src/styles/theme/default.theme.ts
kimon-satan, line 1: OWAs styled component implementation is still on a slow path to be superseded by the OakComponents. We therefore should implement the same changes there to keep things aligned.

src/styles/utils/responsive.ts
```diff
@@ -34,61 +63,90 @@ export const getMediaQuery = (device: Device) => {
...
-    ) => x,
-  ) =>
-  (props: Props): Interpolation<ThemedStyledProps<Props, DefaultTheme>> => {
-    const attrCss = (value: T | undefined | null) =>
-      typeof value === "undefined"
-        ? undefined
-        : css`
-            ${attr}: ${parse(value)};
-          `;
+// Return type that aligns with styled-components' expectations
```
kimon-satan, line 66: Yes this was the gnarly bit which stumped me last time I tried to update styled components.  We should do this in OakComponents as well

kimon-satan (approved):
A lot to take in here. I made a couple of comments around the things I understand better.

.agent/plans/import-order-fix-plan.md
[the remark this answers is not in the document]
jimCresswell: It's a huge change to bring it in line! But, once we do, there are rules that will only allow you to import from index files and type definitions, meaning you can use it to enforce architectural boundaries.

jest.setup.mjs
jimCresswell: I saw that Vitest was slower, which surprised me, and I wondered if that was because of the tangled dependency tree rather than the test engine

src/styles/utils/responsive.ts
jimCresswell: yeah, that took half a day, the Styled-Components docs were not massively useful on that front. The tests pass, so I think the functionality is the same, but we will need to go through the UI carefully

jimCresswell:
> A lot to take in here. I made a couple of comments around the things I understand better.

@kimon-satan we should definitely include some of these improvements in Oak Components, I think we might have to before this PR will build without errors.

jest.setup.mjs
kimon-satan: I can't quite remember the detail but a lot of the slowness is to do with logging and the way the tests are set up. Ultimately I believe it would be faster or the same speed. The payoff is really in the improved DX (especially mocking)

jimCresswell: Definitely, I use Vitest, it is faster, nicer, and handles TypeScript with zero config

kimon-satan: Jamie and Tash might remember what it was that makes our jest tests particularly slow.

next.config.ts
```diff
@@ -127,10 +186,10 @@ module.exports = async (phase) => {
         },
       );
       // Modify the file loader rule to ignore *.svg, since we have it handled now.
-      fileLoaderRule.exclude = /\.svg$/i;
+      (fileLoaderRule as RuleSetRule).exclude = /\.svg$/i;
 
       if (!dev && !isServer) {
-        config.plugins.push(
+        config.plugins?.push(
```
orangemug, line 192: If this is optional, should we be defining this above somewhere?

```
config.plugins = config.plugins ?? []
```

src/browser-lib/seo/Seo.tsx
```diff
@@ -45,6 +45,16 @@ const Seo: FC<SeoProps> = ({
     canonicalURL || `${getBrowserConfig("seoAppUrl")}${router.asPath}`
   )?.replace(/\/$/, ""); //?
 
+  // DEBUG
```
orangemug, line 48: Remove old code?

jest.setup.mjs
orangemug: The decision was made to not move to vitest because it was 100% slower with the experiments that Steve Green worked on.

With regards to slowness not 100% sure, as soon as you introduce our providers things slow up a lot. Most of the tests also include all of the providers.

There is also the issue with oak-components [not being tree shakable](https://github.com/oaknational/oak-components/pull/331)... which I assume is slowing us up a little in the tests (and definitely a big perf hit on the frontend)

orangemug:
I think we need to work out from @jimCresswell, what changes are in this PR prior to reviewing, there seems to be a bunch of non-nextjs update changes that are breaking the UI in a bunch of places. I've reverted some changes in https://github.com/oaknational/Oak-Web-Application/pull/3311/commits/614604e9b8b5907f92a6a6ed02663f85dfddbf38 but I'm not sure what the intention are with these sort of changes

Screenshot 2025-04-08 at 17 54 54

There are quite a lot of them and curric has lots of issues at the moment
 
Screenshot 2025-04-08 at 17 56 38


I would suggest we rollback the component changes, and just include what's required for nextjs updating. As we probably don't want to be testing lots of component changes in this PR as well.

jimCresswell:
Closing as superseded by #3351

PR #3321, opened 2025-03-27
fix: remove media/video object id fro media clips from schema 

weronika-szalas:
## Description

- Remove media/video object id fro media clips schema. We don't use it anywhere and it was causing errors as it wasn't on every object.
- Also made `duration` optional for `mediaObject` as it isn't always present - we don't reference it anywhere atm, we always use `videoObject` duration (even for audios)

## Issue(s)

Fixes [PUPIL-1256](https://www.notion.so/oaknationalacademy/Bug-unable-to-navigate-to-the-media-clip-page-for-a-number-of-lessons-1ae26cc4e1b180c697dad0c4bc07dae4)

## How to test

1. Go to lesson overview of one of the lessons with missing media clips.
2. They should now be present.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -204,6 +204,22 @@
         "npm": ">=10.9 <11"
       }
     },
+    ".yalc/@oaknational/oak-pupil-client": {
+      "version": "2.11.0",
+      "extraneous": true,
+      "license": "MIT",
+      "dependencies": {
+        "@oaknational/oak-curriculum-schema": "^1.22.0",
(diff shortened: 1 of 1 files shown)


RBrownsett (approved):
proved - tested the lessons in the ticket - media clips now show. Regression tested lessons that did not have any issues, audio and media clips still display and can be played
Subjects checed: pe, mfl, cooking, music

benprotheroe (approved):

PR #3328, opened 2025-03-31
Added new "all" icons for Science/RE subjects in curriculum filtering

orangemug:
## Description
Adds correct "all" icons for Science/RE in new filtering

## Issue(s)

Fixes `CUR-1333`

## How to test

1. Go to {owa_deployment_url}/teachers/curriculum/
2. Check subject category filters science/english are correct as per `CUR-1333`

## Screenshots

Screenshot 2025-03-31 at 16 00 32
Screenshot 2025-03-31 at 16 00 18

### DIFF
--- a/src/utils/getValidSubjectCategoryIconById.ts
+++ b/src/utils/getValidSubjectCategoryIconById.ts
@@ -20,7 +20,7 @@ const subjectCategoryIconMap: Record<number, string> = {
 const allIconMap: Record<string, OakIconName> = {
   "religious-education": "subject-religious-education",
   english: "books",
-  science: "rocket",
+  science: "subject-science",
 };
 
 export function getValidSubjectCategoryIconById(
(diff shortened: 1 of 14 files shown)


src/utils/getValidSubjectCategoryIconById.ts
```diff
@@ -17,8 +17,21 @@ const subjectCategoryIconMap: Record<number, string> = {
   17: "theology",
 };
 
-export function getValidSubjectCategoryIconById(id: number): OakIconName {
-  const subjectSlug = subjectCategoryIconMap[id];
+const allIconMap: Record<string, OakIconName> = {
+  "religious-education": "subject-religious-education",
+  english: "books",
+  science: "rocket",
```
rightsaidjames, line 23: @orangemug These should be the subject icons (e.g. `subject-science`) - this is the wrong rocket:
Screenshot 2025-03-31 at 16 31 53

rightsaidjames (changes requested):
See above

src/utils/getValidSubjectCategoryIconById.ts
rightsaidjames: @orangemug Need to do the same for English, right?

PR #3350, opened 2025-04-09
Feat/incomplete units again [LESQ-1344]

njblore:
## Description

Music year: 1985

- Incomplete units banner with hubspot form submission
- Updated lesson counts on unit listing cards (excluding optionality)
- Updated lesson counts on incomplete unit lesson listing pages
- Using a new MV version for browse queries

I've added some extra pages to be checked by percy to cover some more 'edge cases' in the data that would have flagged last time. 

## How to test

The whole journey should be checked thoroughly as MVs have been updated to latest versions throughout

1. Go to https://deploy-preview-3350--oak-web-application.netlify.thenational.academy
2. Go to an incomplete unit, eg. https://deploy-preview-3350--oak-web-application.netlify.thenational.academy/teachers/programmes/geography-secondary-ks4-aqa/units/tropical-rainforest/lessons
3. You should see an incomplete units banner
4. You should see an accurate lesson count
5. You should see the expected lessons compared to prod
6. You should be able to submit your email to hubspot
7. Go to other units that are no incomplete, they should appear the same as production with no banner

## Screenshots

How it should now look:
Screenshot 2025-04-09 at 16 10 22
Screenshot 2025-04-09 at 16 10 38

## Checklist

- [ ] Design sign-off


left23 (changes requested):
Generally, things look good. 
I can see the banner on incomplete units only, and the submission to hubspot works. The lesson counts are updated to reflect the incomplete units on the unit listing pages and on the lesson list pages. And all complete lessons behave as before.
The only issue seems to be with https://deploy-preview-3350--oak-web-application.netlify.thenational.academy/teachers/key-stages/ks4/subjects/german/programmes which I understand is being investigated.

Monika-Koziol (approved):
All looking and working fine for me, subscribed to email, saw banners etc.
Just one question about this pop-up – it won't be there on live site, right? 
Screenshot 2025-04-10 at 13 47 43

left23 (approved):
I understand after reading some comms regarding the issue with German KS4 that is was not directly related to this PR (I think?) and I see it is fixed now anyhow.

While testing this again, I came across another scenario which confused me...

See https://deploy-preview-3350--oak-web-application.netlify.thenational.academy/teachers/programmes/geography-secondary-ks4-aqa/units/natural-and-tectonic-hazards/lessons
On the frontend, the lessons are numbered 1-10, or 1-12 counting the missing lessons.
In the unit download, I see 1-11 (it skips 3) so there are still 10 lessons as we'd expect but the missing 3 looks like it is missing something, when it is not. This seems to be taking the ordering from the CAT, which has no 3.  

Screenshot of CAT (*note it is the staging CAT https://creator-staging.thenational.academy/app/units?=#phase_id=&programme_id=456&subject_id=2&unit_id=3284)
Selection_537
Screenshot of Unit download:
Selection_538

I am curious as to how this comes together when there are also placeholder txt files for incomplete units. It seems like in this case, the numbering will never match completely. We will see 12 lessons on the frontend, numbered 1-12 but the downloaded folders will be 1-13, without 3. I understand this is out of scope for this PR but I am curious if this is something that can be improved for clarity sake. If so, I will create a follow-up item.

Approving!

njblore:
@left23 I think what you're seeing is that the numbers are off in the cat (and currently unit downloads) because the lessons have an absolute 'order' value regardless of where they are used, so even when other lessons aren't included in a certain unitvariant the numbering for each other lesson doesn't adjust. We accomodate this in the front end by essentially ignoring the lesson's order value and we give them a number so it's sequential. We don't currently do that in unit downloads which is why it can be off, but we have made that fix for upcoming unit downloads [(see ticket)](https://www.notion.so/oaknationalacademy/Numbering-in-some-unit-downloads-is-confusing-wrong-17d26cc4e1b1801b85afcc12b38a269b?pvs=4) so the lessons will also be numbered sequentially even when there are missing lessons

PR #3373, opened 2025-04-22
ci: update semantic release config to update change log for all conventional commit types

simonrose121:
## Description

Our husky commitlint config currently doesn't match our semantic release config. Some commit types don't trigger a release at all (e.g. `style`) and some do trigger a release but then don't appear in the change log (e.g. `chore`). This PR updates our semantic release config to be explicit about what releases a version, as well as adding all commit types to our change log. Our open-source codebases should explicitly say what has changed in each version.

Music year: 1951

- Updated semantic release config commit analyser and release notes generator to be explicit for all commit types.


release.config.js
```diff
@@ -8,14 +8,45 @@ module.exports = {
...
+      {
+        preset: "angular",
+        parserOpts: {
+          noteKeywords: ["BREAKING CHANGE", "BREAKING CHANGES", "BREAKING"],
+        },
+        writerOpts: {
+          commitsSort: ["subject", "scope"],
+          types: [
+            { type: "feat", section: "Features" },
+            { type: "fix", section: "Bug Fixes" },
```
mantagen, line 37: Do we want to follow Oak's sentence case convention for these section headings?

mantagen (approved):
Looks great! One minor comment

release.config.js
simonrose121: A very good point, and I'm going to say why not!

simonrose121: Will update

charsville (approved):

PR #3375, opened 2025-04-23
Remove empty year from the visualiser for core/gcse

assadk88:
## Description

Music year: {owa_music_year}

- Remove empty year from the visualiser for core/gcs
- Amended tests and test fixtures to reflect new filtering

## Issue(s)

Fixes #

## How to test

1. Go to https://deploy-preview-3375--oak-web-application.netlify.thenational.academy/teachers/curriculum/computing-secondary-aqa/units?years=10&pathways=core
2. Click on Year 10 (Core)
3. You should see a non-empty year populated with units

## Screenshots

image

How it should now look:
image

## Checklist

- [X] Design sign-off

### DIFF
--- a/src/utils/curriculum/by-pathway.ts
+++ b/src/utils/curriculum/by-pathway.ts
@@ -18,10 +18,11 @@ export function applyFiltering(
     });
   }
 
-  const step1 = Object.values(unitsByYearSelector).filter((item) =>
-    filterIncludes("years", [item.year]),
-  );
-  const step2 = step1
+  const yearDataFilteredBySelectedYear = Object.values(
+    unitsByYearSelector,
+  ).filter((item) => filterIncludes("years", [item.year]));
+
+  return yearDataFilteredBySelectedYear
     .map((item) => {
       return {
         ...item,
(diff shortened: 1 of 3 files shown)


rightsaidjames (approved):

src/utils/curriculum/by-pathway.ts
```diff
@@ -21,18 +21,19 @@ export function applyFiltering(
   const step1 = Object.values(unitsByYearSelector).filter((item) =>
```
sonalisin, line 21: What does step 1 return? The naming of this feels a bit confusing

sonalisin (approved):
Just the one comment on naming - but otherwise looks good & tested in Safari and Chrome.

src/utils/curriculum/by-pathway.ts
sonalisin: I think the naming should reflect the output - I wouldn't be afraid of having longer names that feel overly verbose as it'll give a bit more insight into the filter. Maybe something along the lines of `yearDataFilteredBySelectedYear` and following that something like `unitDataFilteredBy(pathway/visibility etc.)`

orangemug: Yeah lets rename, my bad naming initially but let fix it now rather than down the road. How about

```ts
export function applyFiltering() {
  // Step 1: changed to yearDataFilteredBySelectedYear
  const yearDataFilteredBySelectedYear = // ...
  
  // Step 2: documented by the functions name `applyFiltering`
  return yearDataFilteredBySelectedYear.map((item) => {
  });
}
```

PR #3388, opened 2025-04-30
Fix/lesq 1314/search filter events [LESQ-1314]

njblore:
## Description

Music year: 1980

-  Update `searchRefined` event to remove filter type and value as these are now tracked in a separate event, also add search term and refactor active filters into an object
- Add new `searchFilterModified` event which tracks on filter clicks with type and value and search term
- Update search refined event to run when new filter or year filter is applied as well (instead of search initiated)

## Issue(s)

Fixes #
Filter type and value not being tracked granularly

## How to test

1. Go to https://deploy-preview-3388--oak-web-application.netlify.thenational.academy/search?term=macbeth
2. Click on a filter in the sidebar, you should see the Search Filter Modified event fire with the correct values and "Filter Modification Type": "add"
3. Deselect a filter and you should see the Search Modified event fire with the correct values and "Filter Modification Type": "remove"
4. Remove a filter from the active filters list at the top and the Search Modified event should fire with the same data as above
5. Every type of filter should fire the correct type and value, multiple values should be in a comma separated string list


left23 (approved):
Approving, works as described.

dkyyau (approved):

PR #3396, opened 2025-05-08
Feat/lesq 1323/save units [LESQ-1323]

njblore:
## Description

Music year: 1984

- add graphql queries to get saved units and save units
- add api routes for getSavedUnits and saveUnit
- add helper fns to get and post to the educator api
- add a useSaveUnits hook to handle saving and retrieving saved units in components 
- add a useOakToastContext to display OakToast on save success/error across pages

## How to test

1. Go to a unit listing page eg: https://deploy-preview-3396--oak-web-application.netlify.thenational.academy/teachers/programmes/art-primary-ks1/units/mark-making-using-drawing-tools-and-techniques
2. Save buttons should only appear when the feature flag is active for your posthog id
3. With the feature flag enabled you should see save buttons on each unit card
4. When you are signed out clicking the save button will do nothing yet
5. When you are signed in clicking save you should see a success toast and the button state should change to `Saved`
6. clicking the save button again will do nothing at this point
7. When you go that unit the button at the top of the page should be in a `Saved` state after loading
8. You should be able  to leave and come back and the unit will remain saved
9. You should be able to save a unit from the lesson listing page, and it should be reflected in the unit listing page

## Screenshots

How it should now look:
Screenshot 2025-05-08 at 15 27 46

Screenshot 2025-05-08 at 15 28 18


src/__tests__/pages/teachers/programmes/[programmeSlug]/units/[unitSlug]/lessonListing.test.tsx
```diff
@@ -164,14 +170,13 @@ describe("Lesson listing page", () => {
       ).rejects.toThrowError("no context.params");
     });
   });
-  describe("tracking", () => {
+  describe.skip("tracking", () => {
```
Comment (dkyyau) [inline src/__tests__/pages/teachers/programmes/[programmeSlug]/units/[unitSlug]/lessonListing.test.tsx:173]: Should this still be skipped?

dkyyau (approved):

src/__tests__/pages/teachers/programmes/[programmeSlug]/units/[unitSlug]/lessonListing.test.tsx
Comment (njblore) [reply src/__tests__/pages/teachers/programmes/[programmeSlug]/units/[unitSlug]/lessonListing.test.tsx]: fixed!

left23 (approved):
This looks great. I've tested and it seems everything works as described, including returning in a different session and seeing my previously saved items.

I did some screen reader and keyboard navigation testing too. The only issue I see is when I tab with the keyboard to a unit on the unit listing page and save the unit, the keyboard focus moves back to the top of the unit listing page which means I need to tab through the list again to get back to where I was. It would be more helpful if the focus remains on the current unit's save button or the unit's title link, if possible now or as a follow up. 
Approving!

src/node-lib/educator-api/helpers/saveUnits/useSaveUnits.tsx
```diff
@@ -0,0 +1,66 @@
...
+  );
+
+  const { setCurrentToastProps } = useOakToastContext();
+
+  const onSave = async (unitSlug: string) => {
+    setLocallySavedUnits((prev) => [...prev, unitSlug]);
+    setCurrentToastProps({
+      message: (
+        <OakP>
+          <b>Unit saved</b> to My Library
```
left23, line 28: Just making a note here that I think this should eventually display a more specific message eg. "Unit saved to English KS1". I'll add it to the AC if it's not already included.

PR #3400, opened 2025-05-08
feat: signed out unit saving modal [LESQ-1322]

dkyyau:
## Description

Music year: 1959

- Add `SavingSignedOutModal` component with different button for signed out and signed in but not onboarded users
- Add modal to unit listing and lesson listing save buttons

## How to test

Add posthog id to `teacher-save-units` in Posthog development environment to see save buttons

1. Go to https://deploy-preview-3400--oak-web-application.netlify.thenational.academy
2. Go to a [unit listing](https://deploy-preview-3400--oak-web-application.netlify.thenational.academy/teachers/programmes/english-secondary-ks4-aqa/units) or [lesson listing](https://deploy-preview-3400--oak-web-application.netlify.thenational.academy/teachers/programmes/english-secondary-ks4-aqa/units/macbeth-lady-macbeth-as-a-machiavellian-villain-4877/lessons?sid-1e8095=-oYU-WWOcn&sm=0&src=3) page
3. Click on `Save` button on either page
4. You should see the Saving is here modal with a `Sign up` button linking to signup page if signed out
5. You should see the Saving is here modal with a `Finish sign up` button linking to onboarding if signed in and not onboarded

## Screenshots

Signed out
image

Signed in, not onboarded
image

## Checklist

- [ ] Design sign-off


src/node-lib/educator-api/helpers/saveUnits/useSaveUnits.tsx
```diff
@@ -10,7 +10,12 @@ export const useSaveUnits = (
   programmeSlug: string,
 ) => {
   const [locallySavedUnits, setLocallySavedUnits] = useState<string[]>([]);
-  const { isSignedIn } = useUser();
+  const [openSavingSignedOutModal, setOpenSavingSignedOutModal] =
```
njblore, line 13: can we call this  `showSignIn` / `isSignInModalOpen` or similar, bit difficult to parse currently! plus probably better to keep it less specific as it doesn't have to be be used with that specific modal

njblore (approved):
looking good!

left23 (changes requested):
This looks good! I think there are a few more accessibility improvements that are worth doing now.
1. Add `aria-modal=true` on the same div that has the `role=alertdialog` - See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-modal (unless there is already as reason why this was not included)
2. On the unit listing page, when I exit the modal using keyboard navigation, the focus moves to the pagination (or the last element on the page before the modal markup) when it should be back to where I was when I clicked the button that opens the modal, so the "Save" button of the unit I was saving. On the lesson listing page, this works as expected. The focus returns to the Save button I clicked to open the modal.

left23:
@dkyyau One more accessibility improvement to add now that I've read a bit more.
See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alertdialog_role#associated_wai-aria_roles_states_and_properties - I think the modal wrapping div needs `aria-labelledby` and `aria-describedby` that are then linked to the heading and paragraph text by their IDs.
And I think it's probably more sensible to have the heading as an h2. Thanks!

dkyyau:
> This looks good! I think there are a few more accessibility improvements that are worth doing now.
> 
> 1. Add `aria-modal=true` on the same div that has the `role=alertdialog` - See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-modal (unless there is already as reason why this was not included)
> 2. On the unit listing page, when I exit the modal using keyboard navigation, the focus moves to the pagination (or the last element on the page before the modal markup) when it should be back to where I was when I clicked the button that opens the modal, so the "Save" button of the unit I was saving. On the lesson listing page, this works as expected. The focus returns to the Save button I clicked to open the modal.

The first point here and the ones in the other comment have been addressed. The focus issue seems to be a bigger issue so it might be better to put that into its own ticket to spend more time investigating it.

left23 (approved):
Thank you @dkyyau! Approving.

PR #3401, opened 2025-05-09
fix: removed feature flag check for teacher notes and use cookie conent for hide/enabled/disabled

charsville:
## Description

Music year: {owa_music_year}

- Removed feature flag check for teacher notes
- Hide teacher notes button when cookies have not been accepted/rejected
- Disabled teacher notes button shows if cookies are rejected
- Enabled teacher notes button shows if cookies are accepted

## Issue(s)

Fixes [https://www.notion.so/oaknationalacademy/Loading-performance-of-teacher-notes-button-lesson-overview-19f26cc4e1b180a29adee94d71511581?pvs=4](https://www.notion.so/oaknationalacademy/Loading-performance-of-teacher-notes-button-lesson-overview-19f26cc4e1b180a29adee94d71511581?pvs=4)

## How to test

1. Go to https://deploy-preview-3401--oak-web-application.netlify.thenational.academy/teachers/programmes/music-secondary-ks3/units/keyboard-fundamentals/lessons/playing-a-melody-in-the-c-position-on-the-keyboard
2. Open in an incognito window
3. Initially you should not see an Add/Edit teacher note button
4. If you accept cookies, the Add button should appear and be enabled
5. If you reject cookies the Add button should appear but be disabled

The feature flag check has also been removed, so we should check that there is no regressions from this. The teacher note should always appear and the legacy Share resources button should never appear.

## Checklist

- [ ] Design sign-off
- [ ] When cookies have not been accepted/rejected no button is visible
- [ ] When cookies are accepted the button appears and it works as expected (opens notes modal)
- [ ] When cookies are rejected the button appears but is disabled and clicking it performs no action

### DIFF
--- a/src/__tests__/pages/teachers/lessons/[lessonSlug].test.tsx
+++ b/src/__tests__/pages/teachers/lessons/[lessonSlug].test.tsx
@@ -1,11 +1,14 @@
 import { GetStaticPropsContext, PreviewData } from "next";
-import { useFeatureFlagEnabled } from "posthog-js/react";
+import { MockOakConsentClient } from "@oaknational/oak-consent-client";
+import { omit } from "lodash";
 
 import LessonOverviewCanonicalPage, {
   URLParams,
   getStaticProps,
 } from "@/pages/teachers/lessons/[lessonSlug]";
-import renderWithProviders from "@/__tests__/__helpers__/renderWithProviders";
+import renderWithProviders, {
+  allProviders,
+} from "@/__tests__/__helpers__/renderWithProviders";
 import lessonOverviewFixture from "@/node-lib/curriculum-api-2023/fixtures/lessonOverview.fixture";
 import curriculumApi2023 from "@/node-lib/curriculum-api-2023";
 import OakError from "@/errors/OakError";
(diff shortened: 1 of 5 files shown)


RBrownsett:
@charsville button load is much better. One thing we might need to decide on is the 'add teacher note' button displaying for users who view the shared note. Currently, when viewing a note, live shows the button 'Share resources with colleague', on click this copies the link with the notes. With this fix they see the 'Add teacher note button' which performs no action on click. @messmike what do think the flow should be? Should we keep the button that matches the designs?

Design
image

PR:
image

messmike:
> @charsville button load is much better. One thing we might need to decide on is the 'add teacher note' button displaying for users who view the shared note. Currently, when viewing a note, live shows the button 'Share resources with colleague', on click this copies the link with the notes. With this fix they see the 'Add teacher note button' which performs no action on click. @messmike what do think the flow should be? Should we keep the button that matches the designs?
> 
> Design image
> 
> PR: image

@RBrownsett, good catch. Yes, I think we should match the design. When opening the link with the note added, the user can't add another note or add to the existing note, but the button allows them to copy the link to the page  (if that is possible @charsville, if not, maybe we just hide the button?).

Also, just checking if the inline banner that should appear with the teacher note be part of another ticket? 
Screenshot 2025-05-12 at 9 06 39 am

charsville:
> > > @charsville button load is much better. One thing we might need to decide on is the 'add teacher note' button displaying for users who view the shared note. Currently, when viewing a note, live shows the button 'Share resources with colleague', on click this copies the link with the notes. With this fix they see the 'Add teacher note button' which performs no action on click. @messmike what do think the flow should be? Should we keep the button that matches the designs?
> > > Design image
> > > PR: image
> > 
> > 
> > @RBrownsett, good catch. Yes, I think we should match the design. When opening the link with the note added, the user can't add another note or add to the existing note, but the button allows them to copy the link to the page (if that is possible @charsville, if not, maybe we just hide the button?).
> > Also, just checking if the inline banner that should appear with the teacher note be part of another ticket? Screenshot 2025-05-12 at 9 06 39 am
> 
> Yep that sounds like the best approach. In terms of the banner, @charsville if a user enters data that is redacted does it get saved and displayed as the design above? e.g. My name is [Redacted] please call me on [Redacted]? Or does the api restrict it from being saved in db?

@RBrownsett @messmike yeah so if there is redacted content, then it doesnt even get stored. It blocks the user from adding the note by having the save button in the modal disabled while the PII error messages show. So it shouldnt be possible to see the redacted content there (which probably means we don't need to have the banner?)

And for the button when already viewing a shared note, I think it should be possible to just have the share button without teacher note. I will try this first and if not can hide it.

RBrownsett (approved):
Approved - share button now shows when viewing note. Tested against ACs with no issues. Button is disabled when cookies are not accepted

messmike (approved):

benprotheroe (approved):

PR #3402, opened 2025-05-12
Added `data-percy-hide` attribute to prevent race condition in video content

orangemug:
## Description
In the percy tests we end up in a state where sometimes videos are loaded and sometimes they aren't. Rather than adding hacks in to disable videos in the code, I've instead just added a `data-percy-hide="contents"` to allow the disabling anything we like in the percy tests

> Why not stop the video from loading entirely

Because we use the preview URL to do the testing against, we'd need a conditional in the querystring (or something like that) to disable loading entirely. This seems like a much simpler and robust solution.  Also `data-percy-hide="contents"` is very easy to identify during development.

## Issue(s)

Fixes `FEC-82`

## How to test
To test OWA

1. Go to https://deploy-preview-3402--oak-web-application.netlify.thenational.academy
2. Check videos still work as expects and still looks correct

To test percy: Rerun the tests a couple of time check that we don't get a video race condition

### DIFF
--- a/percy.config.js
+++ b/percy.config.js
@@ -27,6 +27,10 @@ module.exports = {
     minHeight: 1024,
     percyCSS: `
     #mtm-root-container { display: none!important; }
+    
+    *[data-percy-hide="contents"] > * {
+      display: none;
+    }
     `,
   },
   discovery: {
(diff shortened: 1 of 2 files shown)


charsville (approved):

rightsaidjames (approved):
Regression tested the various types of video embeds, all good

PR #3403, opened 2025-05-12
Feat/lesq 1326/unsave units [LESQ-1326]

njblore:
## Description

Music year: 2003

- Adds query and endpoint for unsaving units - deleteUserListContent
- Refactor getUserListContent to only get content currently in a list because we don't delete content from users_content when unsaving

## Issue(s)
~I noticed the OakToast duration does not extend when saving/unsaving multiple units in succession, leading to situations where the toast closes quickly after opening, I will open a separate bug ticket for this~ this should now be fixed and working correctly

## How to test

1. Go to https://deploy-preview-3403--oak-web-application.netlify.thenational.academy/teachers/programmes/art-primary-ks1/units
2. Click on save, there should be a 'unit saved' toast and button state updates to `Saved`
3. Click on saved
4. There should be a unit removed from my library toast matching designs and the button state should revert to `Save`
5. Refreshing the page the button state should stay in the unsaved state
6. Going to the unit the heading button should be in the unsaved state
7. You should still be able to save as normal
8. This should all only be accessible behind the feature flag

## Screenshots

How it should now look:
Screenshot 2025-05-12 at 10 21 02


dkyyau (approved):

left23 (approved):
This looks good, approving! 
I saw the error toast come up once or twice but I couldn't reproduce it intentionally. It was enough to signal to me (as the user) that I might need to click the save icon again, which I think is sufficient for now.
I also noticed that when saving the items in quick succession, the toast remains, without visibly updating to the most recent item (this is fine). 
It makes me wonder how this will work if/when the text in the toast changes to reflect the specifics of the item being saved, for example "Unit removed from [subject keystage]". I think the visual impact might be different (maybe more overwhelming?) due to the text changing quite quickly in that case. No action point right now, just making a note for later.

PR #3409, opened 2025-05-12
Only create error boundary once not every time

orangemug:
## Description
Only create error boundary once not every render

This prevents aggressive rei-rendering of the entire tree, and is the recommended usage, see <https://docs.bugsnag.com/platforms/javascript/react/capturing-render-errors/>

**Note**: Although this is true for lots of parts of the application, the testing steps below if for my particular usecase.

**Also note:** Previous attempt reverted in https://github.com/oaknational/Oak-Web-Application/pull/3408

## Issue(s)
hotfix

## How to test
Test the app locally (with `npm run dev`) as it doesn't appear to be an issue in production.
 
### Load the app in locally
1. Go to http://localhost:3000/teachers/curriculum/english-primary/overview
2. Select the `<h1/>` ("KS1 & KS2 English curriculum") in dev tools
Screenshot 2025-05-12 at 16 36 19

3. Navigate to the explainer
Screenshot 2025-05-12 at 16 36 26

4. The same DOM node should remain present, if the selection gets reset it means that the DOM node was recycled.

Also verify on the `main` branch that the above does indeed fail.

### DIFF
--- a/src/components/AppComponents/ErrorBoundary/ErrorBoundary.tsx
+++ b/src/components/AppComponents/ErrorBoundary/ErrorBoundary.tsx
@@ -60,12 +60,12 @@ export type ErrorBoundaryProps = {
  * and sending a report of the uncaught error to bugsnag.
  */
 const ErrorBoundary: FC<ErrorBoundaryProps> = (props) => {
-  const isBugsnagInited = bugsnagInitialised();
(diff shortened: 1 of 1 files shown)


rightsaidjames (approved):
Confirmed the fix again, and also checked for network and console errors!

njblore (approved):

PR #3412, opened 2025-05-14
fix: e.difference is not a function

njblore:
## Description

Music year: 1984

- An attempt to fix [this error](https://app.bugsnag.com/oak-national-academy/oak-web-application/errors/682318dbd383ca45787459bb?filters[event.since]=30d&filters[app.release_stage]=production&filters[error.status]=open&pivot_tab=event) we're seeing in bugsnag `e.difference is not a function` in useSaveUnits
- I haven't been able to reproduce the error so can't be sure this will fix it

## How to test

1. Go to https://deploy-preview-3412--oak-web-application.netlify.thenational.academy
2. Go to a unit listing page
3. There should be no errors
4. try again while signed in / out 
5. saving and unsaving should still work with the feature flag


left23 (approved):
I also cannot reproduce the error, approving following dev's advice :shipit:

PR #3413, opened 2025-05-14
Feat/lesq 1346/my library page [LESQ-1346]

njblore:
## Description

Music year: 2011

- create a new page at `/teachers/my-library`
- create new components in OWA storybook for header and no units

## How to test

1. While the feature flag is *not active*
    - Go to https://deploy-preview-3413--oak-web-application.netlify.thenational.academy/teachers/my-library
    - You should be redirected to the 404 not found page
2. While the feature flag is active and you are signed out
    - Go to the page again, you should be redirected to sign in
    - After signing up you should be redirected through onboarding
    - After onboarding you should be redirected back to the page
3. While the feature flag is active and you are signed in but not onboarded
    - You should be redirected to onboarding
4. While on the feature flag, signed in and onboarded
    - You should be able to go to the page
    - The design should match figma on desktop and mobile
    - The Start saving button should go to the KS1 subjects page 

## Screenshots

Screenshot 2025-05-14 at 09 33 35


left23 (changes requested):
All this looks good! Everything works as described, with the user flow.

Some comments:

1. When I land on `/my-library` page initially, before adding my ID to the feature flag, the page is blank, except for the cookies banner.  When I accept cookies, then the 404 page displays. Not sure if that can be improved, or if this scenario is too uncommon.

2. When I sign in (after signing up, then signing out), I get directed back to the homepage instead of landing back on my library, despite being redirected to sign in from there.  I think we were aware this might happen, just making a note of this.

3. The heading for the unit lists "No units yet" shouldn't be h1 as the "My library" heading is already h1. If, when there are saved units, the subject and key stage for the collection titles are h2, I think it's better also to have "No units yet" as h2. Or, remove the heading and use p and bold.

4. The mobile designs have the banner flush to the bottom of the header whereas right now there is some white space due to padding on the container (which should remain on the bottom but not the top)
Selection_620

dkyyau (approved):

PR #3414, opened 2025-05-14
fix: use arrays instead of sets to see if that resolves the bug

njblore:
## Description

Music year: 2005

- refactor useSaveUnits to use arrays instead of sets
- hoping this stops the bug we're seeing `e.difference is not a function` and doesn't raise another issue with it instead

## How to test

1. Go to https://deploy-preview-3414--oak-web-application.netlify.thenational.academy
2. Go to a unit listing page
3. There should be no errors
4. try again while signed in / out
5. saving and unsaving should still work with the feature flag


dkyyau (approved):

left23 (approved):
Approving. There are no errors in the console. I tried sign in and out and I can see saving still works.

PR #3416, opened 2025-05-14
Added subjectcategories slug into filter

orangemug:
## Description
Adds the recently added `subjectcategories.slug` into the URL filtering

## Issue(s)
Fixes `CUR-1455`

## How to test

2. Verify that selecting subject categories from the filter updates the URL with the `slug`, rather than the `id` 
3. Hard refreshing the page should return you to the same filter state 

## Checklist

- [ ] Design sign-off

### DIFF
--- a/src/fixtures/curriculum/subjectCategories/index.test.ts
+++ b/src/fixtures/curriculum/subjectCategories/index.test.ts
@@ -5,6 +5,7 @@ describe("createSubjectCategory", () => {
     const result = createSubjectCategory();
     expect(result).toEqual({
       id: 1,
+      slug: "foo",
       title: "Foo",
     });
   });
@@ -13,6 +14,7 @@ describe("createSubjectCategory", () => {
     const result = createSubjectCategory({ id: 999 });
     expect(result).toEqual({
       id: 999,
+      slug: "foo",
       title: "Foo",
     });
   });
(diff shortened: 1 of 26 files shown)


src/fixtures/curriculum/subjectCategories/index.test.ts
```diff
@@ -21,20 +23,21 @@ describe("createSubjectCategory", () => {
     const result = createSubjectCategory({ title: "Badger" });
     expect(result).toEqual({
       id: 1,
+      slug: "foo",
```
assadk88, line 26: Should slug be lowercase badger here?

src/utils/curriculum/filtering.test.ts
```diff
@@ -311,9 +315,13 @@ test("getFilterData", () => {
     subjectCategories: [
       {
         id: 1,
+        slug: "sub-cat-1",
+        title: "Foo",
```
assadk88, line 319: Is worth thinking about enforcing integrity of fixtures through some means, i.e. slug should always be lower-case variant of title? Not sure if we can do that through types though 🤔

assadk88 (approved):

src/fixtures/curriculum/subjectCategories/index.test.ts
orangemug: The rules are `slug` generates `title`, but `title` does not effect `slug`. I've pretty much followed that everywhere in the fixtures.

Essentially I don't want people using this in the tests

```ts
createSubjectCategory({ title: "Badger" })
```

And relying on `{slug: "badger"}`, because the tests will get messy as we should be using "slugs" almost everywhere.

src/utils/curriculum/filtering.test.ts
orangemug: I like that, although I think the underlying types don't enforce that so neither can we.

The database could have `slug` equal to `FOO`, so we'd need to make sure the database enforces that first. I think that sounds like a can of worms right now and lots of squad coordination. Maybe just one we keep in mind for future improvements.

rightsaidjames (changes requested):
@orangemug The currently active subject category (including 'all') is not being populated in the mobile header:
Screenshot 2025-05-22 at 09 59 24

Otherwise, looks good so far. Will report back if I find anything else!

rightsaidjames:
Another issue found @orangemug. In Secondary English, the Literature icon has changed from this (correct):
Screenshot 2025-05-22 at 14 58 18

To the fallback image (incorrect):
Screenshot 2025-05-22 at 14 58 08

PR #3420, opened 2025-05-15
Fix/lesq 1408/oak toast position [LESQ-1408]

njblore:
## Description

Music year: 2014

- Adds the router path as a dependency to the useEffect in oak toast context which controls the padding as this seemed to be happening when navigating to a page without a full refresh

## How to test

1. Go to https://deploy-preview-3420--oak-web-application.netlify.thenational.academy
2. Navigate to a unit page with the saving feature flag enabled
3. Clicking save should show the toast  82px from the top accounting for the nav bar
4. Scroll down the page and save, the toast should show 32px from the top
5. Without scrolling again click a unit to navigate to the page
6. Click save and the toast should still be fully visible 82px fom the top

## Screenshots

Screenshot 2025-05-15 at 13 25 00

How it should now look:
Screenshot 2025-05-15 at 13 25 50

### DIFF
--- a/CHANGE_LOG.md
+++ b/CHANGE_LOG.md
@@ -1,3 +1,5 @@
+## [1.747.1](https://github.com/oaknational/Oak-Web-Application/compare/v1.747.0...v1.747.1) (2025-05-15)
+
 # [1.747.0](https://github.com/oaknational/Oak-Web-Application/compare/v1.746.2...v1.747.0) (2025-05-14)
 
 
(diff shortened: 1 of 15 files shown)


left23 (approved):
Thanks @njblore for picking this up so quick. Works well now.

If it's not too annoying and easily included here, could you edit this line https://github.com/oaknational/Oak-Web-Application/blob/main/src/node-lib/educator-api/helpers/saveUnits/useSaveUnits.tsx#L17 so "Library" has a lower case first letter? Then it's consistent with sentence case and the remove unit message. Sorry, I could have caught that one sooner, many thanks!

dkyyau (approved):

PR #3421, opened 2025-05-16
[LESQ 1392] sign up btn in nav bar

Taha-Hassan-Git:
## Description

This PR adds a Sign Up button to the Nav bar

- Clerk Sign up button in the nav bar wrapping an OakSecondary Button, only visible to signed out teachers
- Redirects to Onboarding after sign up
- Redirects back to start point after Onboarding

Link to the designs - https://www.figma.com/design/YcWQMMhHPVVmc47cHHEEAl/Oak-Design-Kit?node-id=8866-10602&p=f&m=dev

Link to the deployment https://deploy-preview-3421--oak-web-application.netlify.thenational.academy/

Closes [LESQ 1392- I want a Sign Up button in the Nav Bar](https://www.notion.so/oaknationalacademy/I-want-a-sign-in-button-in-the-nav-bar-1eb26cc4e1b180aca2e1cff136c5cef7)  

## Issue(s)

- The designs call for a font size of 14px and weight of 600, the closest design token I could find to that is "body-3-bold" which has a weight of 700 
- There aren't tests for the redirects, I couldn't find any e2e tests at all actually.

## How to test

Checking the button renders as expected
1. Go to the home page with a signed out account
2. You should see the sign up button
3. Click the Pupil link
4. The sign up button should disappear
5. Click the teacher link and sign up
6. After completing sign up you should see the use avatar instead of the sign up button

Checking the redirects
1. Go to a page other than the home page in the Teacher user flow
2. Click the sign up button
7. Sign up with a new account
8. You should be redirected to onboarding after sign up is finished
9. Complete the onboarding
10. You should be redirected back to the page from step one after completing onboarding

## Screenshots

How it used to look:
Screenshot 2025-05-20 at 16 44 18

How it should now look:

Kapture 2025-05-20 at 16 51 45

## Checklist

- [ ] Design sign-off


src/components/AppComponents/AppHeader/AppHeader.tsx
```diff
@@ -92,6 +106,19 @@ const AppHeader: FC<HeaderProps> = () => {
                 }}
                 data-testid="clerk-user-button"
               />
+            ) : (
+              selectedArea == siteAreas.teachers && (
+                <SignUpButton forceRedirectUrl={onboardingRedirectUrl}>
+                  <OakSecondaryButton
+                    name="sign up"
```
left23, line 113: The name attribute here isn't necessary as there is no form element associated with this button and the span inside the button element has text which will be used for the screenreader.

left23 (changes requested):
This looks good, everything works as described.

I think this should be considered with the impact of this PR https://github.com/oaknational/oak-components/pull/422 which introduces the saving units counter (currently as a component only) https://deploy-preview-422--lively-meringue-8ebd43.netlify.app/?path=/docs/components-molecules-oaksavecount--docs

The designs show the "Sign up" button larger than the count icon, but as it is now, they will be the same height (32px)
Selection_634

src/components/AppComponents/AppHeader/AppHeader.tsx
```diff
@@ -92,6 +106,19 @@ const AppHeader: FC<HeaderProps> = () => {
                 }}
                 data-testid="clerk-user-button"
               />
+            ) : (
+              selectedArea == siteAreas.teachers && (
+                <SignUpButton forceRedirectUrl={onboardingRedirectUrl}>
+                  <OakSecondaryButton
```
njblore, line 112: It looks like the design is using [`OakSmallSecondaryButton` ](https://www.figma.com/design/YcWQMMhHPVVmc47cHHEEAl/Oak-Design-Kit?node-id=11615-2857&m=dev)which has the correct font size and padding already!

Taha-Hassan-Git: Oh my! @Monika-Koziol even pointed this out to me and I forgot to make the change. Will push it up now.

Taha-Hassan-Git:
> This looks good, everything works as described.
> 
> I think this should be considered with the impact of this PR [oaknational/oak-components#422](https://github.com/oaknational/oak-components/pull/422) which introduces the saving units counter (currently as a component only) https://deploy-preview-422--lively-meringue-8ebd43.netlify.app/?path=/docs/components-molecules-oaksavecount--docs
> 
> The designs show the "Sign up" button larger than the count icon, but as it is now, they will be the same height (32px) Selection_634


Is this a matter of making the other component smaller, or should mine be bigger?

src/components/AppComponents/AppHeader/AppHeader.test.tsx
```diff
@@ -102,4 +103,20 @@ describe("components/AppHeader", () => {
...
+
+    const signUpButton = screen.getByTestId("sign-up-button");
+    expect(signUpButton).toBeInTheDocument();
+  });
+
+  it("does not render a sign up button when a user is logged in", async () => {
+    setUseUserReturn(mockLoggedIn);
+    renderWithProviders()(<AppHeader />);
+
+    const signUpButton = screen.queryByTestId("sign-up-button");
```
njblore, line 119: 👀  did you refactor this from using `getByRole`? I think that would be the better approach here! was it causing issues?

src/components/AppComponents/AppHeader/AppHeader.tsx
```diff
@@ -1,6 +1,14 @@
...
+import {
+  OakBox,
+  oakColorTokens,
+  OakFlex,
+  OakSmallSecondaryButton,
+} from "@oaknational/oak-components";
+import { UserButton, useUser, SignUpButton } from "@clerk/nextjs";
+import { useRouter } from "next/router";
+
+import { resolveOakHref } from "../../../common-lib/urls";
```
njblore, line 11: we have a path alias set up for `@/` which maps to `src/`, so you can replace ugly relative imports with `@/common-lib/urls` 😁

njblore (approved):
Looking good! Just a question on your test 

Also not one for now necessarily but I think it would be good to extract the sign up button and user button logic to it's own component now it's getting a bit more complex, what do you think?

src/components/AppComponents/AppHeader/AppHeader.test.tsx
Taha-Hassan-Git: It was getByRole then with a selector for name, but I took the name attr out based on Lisa's comment above. I couldn't find a way to getByRole then getByText. Happy to put the name attr back in if you feel it's better, or perhaps there's another selector I could use instead of name.

Taha-Hassan-Git:
> Looking good! Just a question on your test
> 
> Also not one for now necessarily but I think it would be good to extract the sign up button and user button logic to it's own component now it's getting a bit more complex, what do you think?

Happy to make the change on this PR, definitely makes sense to do this.

src/components/AppComponents/AppHeader/AppHeader.test.tsx
njblore: Aha I see! with `getByRole` the name value should work using the text content of the button as well as that is the accessible name of the button, I think it's case sensitive though. If that's not working though it's possible it could be getting confused due to the nesting of buttons, in which case I'd probably prefer to use 'getByText' over `getByTestId`

Monika-Koziol (approved):
All good from my side. And really nice information in the PR. Thank you.

left23 (approved):
@Taha-Hassan-Git What I understood was that the "sign up" button was meant to be larger than the save icon. At the moment, it seems they will be the same size - difficult to see without the 2 elements being in place at once.
If @Monika-Koziol has approved this, I'm good with it as is. It can always be iterated upon. I also like @njblore 's idea to make this into a component at some point too. Approving!

Taha-Hassan-Git:
> Also not one for now necessarily but I think it would be good to extract the sign up button and user button logic to it's own component now it's getting a bit more complex, what do you think?

@njblore I decided to call the component ```TeacherAccountButton``` and put it in the TeacherComponents directory. Let me know if that's appropriate. I can see some components in this folder are re-exported via an ```index.ts``` file and some aren't, not sure about the logic there, but can re-export this component if needed.

src/components/AppComponents/AppHeader/AppHeader.tsx
```diff
@@ -66,31 +74,10 @@ const AppHeader: FC<HeaderProps> = () => {
...
-                      [`@media (max-width: ${getBreakpoint("small")}px)`]: {
-                        width: "100%",
-                        maxWidth: "100%",
-                        marginLeft: "0",
-                      },
-                    },
-                  },
-                }}
-                data-testid="clerk-user-button"
+            {selectedArea === "TEACHERS" && (
```
njblore, line 77: this has slightly changed the behaviour as previously you could sign out via the `UserButton` in the pupils area but now you won't be able to - I'd suggest passing `selectedArea` through to `TeacherAccountButton` and have the conditional rendering on just the `SignUpButton` in there

Taha-Hassan-Git: Wow! Good catch!

PR #3426, opened 2025-05-21
Fixing more test log issues

orangemug:
## Description
Fixes to reduce log output in tests (only 17 files are logging now)

## Issue(s)
Partial fix for `FEC-72`

## How to test
Only dev review required.

### DIFF
--- a/src/components/TeacherViews/Onboarding/HowCanOakSupport/HowCanOakSupport.view.tsx
+++ b/src/components/TeacherViews/Onboarding/HowCanOakSupport/HowCanOakSupport.view.tsx
@@ -107,18 +107,21 @@ const HowCanOakSupport = () => {
               control={control}
               name={key as OakSupportKey}
               key={key}
-              render={({ field }) => (
-                <OakCheckBox
-                  id={key}
-                  key={key}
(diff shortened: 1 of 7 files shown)


src/components/TeacherViews/Onboarding/HowCanOakSupport/HowCanOakSupport.view.tsx
```diff
@@ -107,18 +107,21 @@ const HowCanOakSupport = () => {
...
-                  {...field}
-                  value={value}
-                  onChange={(event) => {
-                    clearErrors();
-                    field.onChange(event);
-                  }}
-                />
-              )}
+              render={({ field }) => {
+                const { ref, ...fieldProps } = field;
```
orangemug, line 111: @njblore turns out we were passing `ref` here but [`<OakCheckBox/>`](https://github.com/oaknational/oak-components/blob/main/src/components/molecules/OakCheckBox/OakCheckBox.tsx) doesn't use [`forwardRef(...)`](https://react.dev/reference/react/forwardRef)

njblore (approved):

PR #3432, opened 2025-05-22
Feat/lesq 1324/saved unit counts [LESQ-1324]

njblore:
## Description

Music year: 1956

- Create a new query and endpoint getSAvedUnitCount
- Create a new component SaveCount which fetches the save coutn for a user and renders it in an OakSaveCount component when the feature flag is on
- Add SaveCount to AppHeader at tablet & desktop sizes
- Add my-library as a page type in Avo, [branch here ](https://www.avo.app/schemas/5PhajbVijwhXVKIJtGMT/branches/jDHHtOWO-NLkNu3SwIMv5/diff)
- Created a new context SaveCount to keep save count in sync in the app header when it's updated in unit or lesson listing pages. If we start to need more global context like this as we develop this feature we might want to look into more of a state management solution.

## How to test

1. Go to https://deploy-preview-3432--oak-web-application.netlify.thenational.academy
2. Without the feature flag 
    - you should not see the save count button in the nav bar
4. With the feature flag while signed out
    - you should see an icon with 0 saves
    - clicking the icon should redirect to the sign up journey
    - after sign up you should be redirected to onboarding
5. While signed in
    - you should see a count matching the number of saves you have
    - when you save a unit the count should update
    - when you unsave a unit the count should update

## Screenshots

Screenshot 2025-05-22 at 14 35 45


left23 (changes requested):
This works well! Sign up/in and directing to onboarding is all good. I can see an accurate saved items count on my existing user when I sign in, and the count increases and decreases as I save/unsave units.
The only thing that stands out is on mobile the counter is hidden accidentally. When I forced it to display with the web dev tools, I can see too that the gap will need to be reduced so everything fits on narrow viewports
I have a 360px wide mobile btw. It's pretty small by today's standards but maybe this is a good minimum size to provide for?
Selection_638

njblore:
@left23 thanks! I can add another breakpoint for large mobiles but I'd need to set it to around 460px to accomodate the sign up button as well, I'm not sure how many actual mobiles this would capture? 

460px:  
Screenshot 2025-05-23 at 08 13 02 
  
360px:  
Screenshot 2025-05-23 at 08 17 21

Taha-Hassan-Git (approved):
Looks good to me!

avo.json
```diff
@@ -7,17 +7,17 @@
     "name": "Oak's Tracking Plan"
   },
   "branch": {
-    "id": "master",
```
Taha-Hassan-Git, line 10: Just for my own learning, would be interested to know what's going on here. Can do this via other channels.

njblore: Basically we use a tool called Avo to manage our events for us, when we need to add or update events you make a new branch in Avo and make your changes to the event structure in there, then pull the branch into OWA and the event types get updated in `Avo.ts` so they can be used throughout the app. It's mostly a glorified type generator for our use case but keeps our events tidy and consistent. 
See the docs [here](https://www.notion.so/oaknationalacademy/Avo-5ff21cbd33994c84907c83c6af5a25ee) and you can see this branch [here](https://www.avo.app/schemas/5PhajbVijwhXVKIJtGMT/branches/jDHHtOWO-NLkNu3SwIMv5/diff)

left23: I'm also curious! If you haven't already, maybe you can elaborate here to spread the knowledge.

njblore:
@left23 Ah I see I misunderstood sorry, you mean to adjust the whole nav bar layout so it works on smaller mobiles not larger mobiles? That is a limitation of the design system that we only have mobile-tablet-desktop breakpoints set up, and these are baked into all of our spacing value so to apply different padding/sizing on smaller mobiles would require a new breakpoint to be added and propagated throughout the system

src/components/TeacherComponents/SaveCount/SaveCount.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const { savedUnitsCount, setSavedUnitsCount, loading } =
+    useSaveCountContext();
+
+  useEffect(() => {
+    if (unitsCount !== undefined) {
+      setSavedUnitsCount(unitsCount);
+    }
+  }, [unitsCount, setSavedUnitsCount]);
+
+  return isSaveEnabled ? (
```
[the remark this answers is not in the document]
njblore: Oh interesting why is that? I tend to prefer a ternary in these sitatuations as the intent is more readable to me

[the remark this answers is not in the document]
Taha-Hassan-Git: Same reason but the opposite! I just find the other way more readable.

left23:
@njblore I just noticed @Monika-Koziol  and your conversation about this in Notion. I think the proposal is to hide only the count numbers, not the icon itself. Should I expect this to happen on this PR or in the component? Let me know what I should do next please. Am I waiting for adjustments to be made or do you need to merge this to proceed?

PR #3434, opened 2025-05-22
Updated gleap to `v14.6.6`

orangemug:
## Description
There are gleap errors being thrown in the frontend for various users in the gleap dependency.

## Issue(s)
hotfix

## How to test
Smoke test gleap still works correctly


rightsaidjames (approved):
Have done a regression test of Gleap (including some test submissions), all looks good.

PR #3460, opened 2025-06-03
fix: reset saved units count on sign out [LESQ-1431]

njblore:
## Description

Music year: {owa_music_year}

- listen for sign out in SaveCount component and set count to 0 on sign out
- call `mutate` on sign in to update unit count

## Issue(s)

Fixes #
https://www.notion.so/oaknationalacademy/Save-icon-and-count-does-not-return-to-zero-when-user-signs-out-20326cc4e1b180e2a174e0248a0807b5?pvs=4

## How to test

1. Go to {owa_deployment_url} and sign in
2. Go to a unit listing page and save some units so your save count is above 0
3. Sign out
4. You should see your save count reset to 0
5. Sign in again
6. You should see a loading state (grey save count) and then your count should update to the correct number

### DIFF
--- a/src/components/TeacherComponents/SaveCount/SaveCount.tsx
+++ b/src/components/TeacherComponents/SaveCount/SaveCount.tsx
@@ -1,6 +1,7 @@
 import { OakSaveCount } from "@oaknational/oak-components";
 import { useFeatureFlagEnabled } from "posthog-js/react";
 import { useEffect } from "react";
+import { useUser } from "@clerk/nextjs";
 
 import { resolveOakHref } from "@/common-lib/urls";
 import useSaveCountContext from "@/context/SaveCount/useSaveCountContext";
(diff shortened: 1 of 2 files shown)


src/components/TeacherComponents/SaveCount/SaveCount.tsx
```diff
@@ -1,17 +1,21 @@
...
+  const { isSignedIn } = useUser();
 
-  const { data: unitsCount, isLoading } = useGetEducatorData<number>(
-    "/api/educator-api/getSavedUnitCount",
-  );
+  const {
+    data: unitsCount,
+    isLoading,
+    mutate,
+  } = useGetEducatorData<number>("/api/educator-api/getSavedUnitCount");
```
Taha-Hassan-Git, line 18: Looks like calling mutate with no arguments updates server state with local values? Is this just for the saved units data or everything?

PR #3468, opened 2025-06-06
Fix/lesq 1405/focus issues [LESQ-1405]

njblore:
## Description

Music year: {owa_music_year}

- update SavingSignedOutModal with manual `returnFocus` method, ignoring the default suggested node if an element Id has been set (for a save button)
- update `saveButtonId` on save click
- refactor `UnitList`, nested components within the main component were causing focus to be lost when clicking save and triggering rerenders

## Issue(s)

Fixes #
- focus goes to the bottom of the page after closing the saving sign up modal
- focus returns to the top of the units list after clicking save on a unit

## How to test

1. Go to https://deploy-preview-3468--oak-web-application.netlify.thenational.academy/teachers/programmes/history-primary-ks2/units
2. While signed out use keyboard nav to click the save button and close the modal
3. Closing the modal either with keyboard or escape should return focus back to the save button you just clicked
4. After signing in, clicking a save button with keyboard should keep focus on the button


left23 (approved):
Works as described with the modal when signed out, and general saving and navigating the unit list. Thanks for the efforts, it's a much smoother and more accessible UX.

dkyyau (approved):

PR #3475, opened 2025-06-10
Feat/lesq 1454/curriculum download [LESQ-1454]

njblore:
## Description

Music year: 1974

- create a component `CurriculumDownloadsBanner`
- create a hook `useCurriculumDownloads` to manage download button state
- extract some shared functions for creating the download button and update in curriculum journey
    - importing other helpers directly from curriculum tab as I believe this will be a short-lived banner
    - if it ends up sticking around we may want to extract to a shared place
- add tracking 
- disable curriculum download for RSHE*

*I noticed this had been overlooked when adding the banner, there is no curriculum download for RSHE

## How to test

1. Go to https://deploy-preview-3475--oak-web-application.netlify.thenational.academy
2. Go to a cycle 2 unit listing page
3. You should see the download banner
4. When you have school picker data in your localstorage
    - the download button should immediately start the curriculum download
6. When you have no school picker data 
    - the button should link to the curriculum download page

To add school picker data you can go via the curriculum download or lesson download journeys, or sign in.
To remove the data you can delete it from localstorage, it is under the key `oak-downloads-school`

## ACs
- [ ] you should be able to download curriculum from the unit listing page while having saved school picker details
- [ ] you should not be able to download from the unit listing page without having saved school picker details
- [ ] you should not be able to download without having accepted terms and conditions
- [ ] you should not see the download banner on RSHE pages
- [ ] you should see a tracking event fired on download: `curriculumResourcesDownloaded`

### DIFF
--- a/src/__tests__/pages/teachers/programmes/[programmeSlug]/units/unitListing.test.tsx
+++ b/src/__tests__/pages/teachers/programmes/[programmeSlug]/units/unitListing.test.tsx
@@ -134,6 +134,24 @@ describe("pages/programmes/[programmeSlug]/units", () => {
     });
     expect(curriculumDownloadButton).not.toBeInTheDocument();
   });
+  it("does not render curriculum download button on rshe", () => {
+    render(
+      <UnitListingPage
+        curriculumData={{
+          ...unitListingFixture(),
+          programmeSlug: "rshe-secondary-ks4",
+          subjectSlug: "rshe-pshe",
+          hasCycle2Content: true,
+          subjectTitle: "RSHE (PSHE)",
+        }}
+        curriculumRefreshTime={0}
+      />,
+    );
+    const curriculumDownloadButton = screen.queryByRole("button", {
+      name: "Download curriculum plan",
+    });
+    expect(curriculumDownloadButton).not.toBeInTheDocument();
+  });
   it("renders curriculum download button on cycle 2 programmes", () => {
     render(
       <UnitListingPage


src/components/TeacherComponents/CurriculumDownloadBanner/useCurriculumDownloads.tsx
```diff
@@ -0,0 +1,128 @@
...
+      schoolId: data.schoolId!,
+      schoolName: data.schoolName!,
+      email: data.email!,
+      termsAndConditions: data.termsAndConditions!,
+      schoolNotListed: data.schoolNotListed!,
+    };
+    saveDownloadsDataToLocalStorage(schoolData);
+
+    try {
+      await downloadFileFromUrl(downloadPath);
```
Taha-Hassan-Git, line 90: What happens if downloads fail?

Taha-Hassan-Git (approved):
Looks good! Just a couple of questions.

src/pages/teachers/programmes/[programmeSlug]/units.tsx
```diff
@@ -221,6 +223,12 @@ const UnitListingPage: NextPage<UnitListingPageProps> = ({
     );
   };
 
+  const subjectParentSlug = subjectParent
```
Comment (Taha-Hassan-Git) [inline src/pages/teachers/programmes/[programmeSlug]/units.tsx:226]: Not sure what's happening here. Perhaps worth a comment?

src/pages/teachers/programmes/[programmeSlug]/units.tsx
```diff
@@ -221,6 +223,12 @@ const UnitListingPage: NextPage<UnitListingPageProps> = ({
...
+    ? convertSubjectToSlug(subjectParent)
+    : null;
+  const showCurriculumDownloadBanner =
```
Comment (Taha-Hassan-Git) [inline src/pages/teachers/programmes/[programmeSlug]/units.tsx:229]: Worth adding a test for this behaviour?

src/pages/teachers/programmes/[programmeSlug]/units.tsx
Comment (njblore) [reply src/pages/teachers/programmes/[programmeSlug]/units.tsx:226]: some subject have parent-child relationships, eg. `science` is the parent of `biology`, `physics` and `chemistry`. We _generally_ don't really care about parent subjects in the browse journey and just show child subjects on their own, but in the curriculum journey they use the parent so we need to get the slug to pass through to the curriculum download endpoint

src/components/TeacherComponents/CurriculumDownloadBanner/useCurriculumDownloads.tsx
njblore: very good question!  perhaps we should show an error toast, I'll check with Monika

PR #3483, opened 2025-06-11
feat: lesson listing page SEO body text [LESQ-1453]

dkyyau:
## Description

Music year: {owa_music_year}

- Adds LessonListSEOHelper to lesson list page for non-legacy units

## Issue(s)

Fixes #

## How to test

1. Go to {owa_deployment_url}
3. You should see body text accordion at the bottom of the page

## Screenshots

How it should now look:
image

## Checklist

- [ ] Design sign-off

### DIFF
--- a/src/components/TeacherComponents/LessonList/LessonList.test.tsx
+++ b/src/components/TeacherComponents/LessonList/LessonList.test.tsx
@@ -98,4 +98,46 @@ describe("components/ Lesson List", () => {
 
     expect(onClick).toHaveBeenCalledTimes(1);
   });
+  test("it renders the SEO accordion for non-legacy units", async () => {
+    const { getByTestId } = render(
+      <LessonList
+        paginationProps={mockPaginationProps}
+        subjectSlug={"computing"}
+        keyStageSlug={"ks2"}
+        headingTag={"h2"}
+        currentPageItems={lessonsWithUnitData}
+        unitTitle={"Unit title"}
+        lessonCount={4}
+        onClick={onClick}
+        lessonCountHeader={`Lessons (${lessons.length})`}
+        programmeSlug="computing-primary-ks2"
+        yearTitle="Year 6"
+        subjectTitle="Computing"
+      />,
+    );
+    const seoAccordion = getByTestId("lesson-list-seo-accordion");
+
+    expect(seoAccordion).toBeInTheDocument();
+  });
+  test("it does not render the SEO accordion for legacy units", async () => {
+    const { queryByTestId } = render(
+      <LessonList
+        paginationProps={mockPaginationProps}
+        subjectSlug={"computing"}
+        keyStageSlug={"ks2"}
+        headingTag={"h2"}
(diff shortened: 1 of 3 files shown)


src/components/TeacherComponents/LessonList/LessonList.test.tsx
```diff
@@ -98,4 +98,46 @@ describe("components/ Lesson List", () => {
...
+        unitTitle={"Unit title"}
+        lessonCount={4}
+        onClick={onClick}
+        lessonCountHeader={`Lessons (${lessons.length})`}
+        programmeSlug="computing-primary-ks2-l"
+        yearTitle="Year 6"
+        subjectTitle="Computing"
+      />,
+    );
+    const seoAccordion = queryByTestId("lesson-list-seo-accordion");
```
Taha-Hassan-Git, line 139: Might it be better to query by text here, since that's what the user will see?

Taha-Hassan-Git (approved):
Looks good to me! Just one suggestion.

src/components/TeacherComponents/LessonList/LessonList.test.tsx
dkyyau: The text is broken up into multiple elements when it renders so it's difficult to query by a specific enough string that finds the element

left23 (changes requested):
This looks good!
I noticed there is a vertical scroll on the body of the accordion, maybe due to the `overflow: auto` on that element. In browserstack, I don't see this on Mac Safari or Chrome, but I see it in Firefox, and locally on my Ubuntu Chrome and Firefox.
Selection_057

Also I find the text a bit hard to understand when it refers to the unit title without any punctuation. 
`
slide decks, worksheet PDFs, quizzes and lesson overviews. You can select individual lessons from the Step into the unknown: fiction reading and creative writing unit and download the resources you need, or download the entire unit now. See every unit ...
`
We had a similar issue with the lesson SEO accordion and currently there is a colon before the lesson title so "To help you plan your year 7 english lesson on: [lesson title], download all teaching resources..." which is better although still has its issues (for example, if the lesson title also has a colon or single quotes)
I think something is needed here to distinguish the unit title from the paragraph of text.
Maybe either with a colon or quotes around it:
`You can select individual lessons from: [unit title] ...` or `You can select individual lessons from "[unit title]"...`

Or maybe the unit title in bold or italics would prevent the issue of having multiple colons in the sentence. This isn't a blocker and depending on the purpose of this (for SEO or also humans?) it is fine as is.

left23:
@dkyyau I think there needs to be an exception for RSHE - currently it doesn't appear in the curriculum so the link leads to a 404. For the lesson SEO section, the link to the curric is omitted in the case of RSHE.

left23 (approved):
Looks great! Thanks for making those changes to the lesson SEO section too. I think since we know these sections don't get much human attention, unlinking the text is sufficient for now despite the sentence still mentioning the curriculum.
I can confirm the scrollbar is gone too!

PR #3486, opened 2025-06-12
Fix toasts rendering behind the header

Taha-Hassan-Git:
## Description

This PR fixes an issue where toasts intermittently appeared behind the header. I think this was due to a race condition where the IntersectionObserver evaluated header visibility before DOM was fully rendered. It wasn't just happening in the unit listing page, but on other pages also.

### Solution

- Added 100ms timeout before observing header element (this feels a bit brittle. from reading around I think [a mutation observer](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) might be better. I tried to use one but I couldn't get it to work)
- Drive-by fix : Set intersection threshold to 0.5 (header must be 50% visible)
- Drive-by refactor: added a named variable for header visibility to make the code read a bit more clearly

#### threshold behaviour

Old behaviour: header is barely visible, toasts are offset
Kapture 2025-06-12 at 11 43 08

New behaviour: toasts are not offset until more than 50% is visible
Kapture 2025-06-12 at 11 46 38

## Issue(s)

Fixes https://www.notion.so/oaknationalacademy/Toast-still-appears-in-wrong-position-sometimes-20326cc4e1b1801ebfc8ff3dd26ee5e3

## How to test

This bug didn't appear every time the steps below were followed, but often enough that a few repetitions will ensure it's no longer a problem

1. Go to a key stage (subjet listing page?)
2. Click on a subject
3. Without scrolling, click save on a unit
4. The toast should render below the header

## Screenshots

screenshot

How it should now look:
As above

## Checklist

- [ ] Design sign-off

### DIFF
--- a/src/components/GenericPagesComponents/CurriculumTab/__snapshots__/CurriculumTab.test.tsx.snap
+++ b/src/components/GenericPagesComponents/CurriculumTab/__snapshots__/CurriculumTab.test.tsx.snap
@@ -8,7 +8,7 @@ exports[`CurriculumTab renders 1`] = `
     >
       <div
         aria-live="polite"
-        class="sc-fqkvVR sc-gsFSXq sc-442d8add-0 hkKUlc cQyxGE ehPczd"
+        class="sc-fqkvVR sc-gsFSXq sc-d65e4df9-0 hkKUlc cQyxGE jTGGef"
       />
       <div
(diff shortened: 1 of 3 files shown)


njblore (approved):

left23 (approved):
Nice, all works well. Thanks, approving!

PR #3503, opened 2025-06-23
[ENG-1241] Set up storybook in vercel

KofoworolaOgunleye:
## Description

- deploys storybook build of OWA

## Issue(s)

[ENG-1241](https://www.notion.so/oaknationalacademy/Set-up-Storybook-in-Vercel-with-Terraform-21626cc4e1b180e6a6a1e451a5a1438c)

## How to test

1. check that everything works as expected in vercel

## Checklist

- [ ] Something that needs doing

### DIFF
--- a/infrastructure/project/builds.tf
+++ b/infrastructure/project/builds.tf
@@ -4,6 +4,7 @@ locals {
       description     = "Oak Web Application Website"
       domains         = ["owa-vercel.thenational.academy"]
       build_type      = "website"
+      deployment_type = "standard_protection"
       framework       = "nextjs"
       ignore_command  = "if [ \"$VERCEL_GIT_COMMIT_REF\" = \"feat/eng-1145-vercel-migration\" ]; then exit 1; else exit 0; fi"
       skew_protection = "7 days"
@@ -13,6 +14,7 @@ locals {
       domains          = ["storybook-vercel.thenational.academy"]
       build_command    = "npm run storybook:build"
       build_type       = "storybook"
+      deployment_type  = "all_deployments"
       framework        = "storybook"
       output_directory = "storybook-static"
       skew_protection  = "1 day"
--- a/infrastructure/project/builds.tf
+++ b/infrastructure/project/builds.tf
@@ -1,0 +1,21 @@
+locals {
+  builds = {
+    website = {
+      description     = "Oak Web Application Website"
+      domains         = ["owa-vercel.thenational.academy"]
+      build_type      = "website"
+      framework       = "nextjs"
+      ignore_command  = "if [ \"$VERCEL_GIT_COMMIT_REF\" = \"feat/eng-1145-vercel-migration\" ]; then exit 1; else exit 0; fi"
+      skew_protection = "7 days"
+    },
+    storybook = {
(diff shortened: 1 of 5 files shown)


infrastructure/project/builds.tf
```diff
@@ -0,0 +1,20 @@
...
+      framework       = "nextjs"
+      ignore_command  = "if [ \"$VERCEL_GIT_COMMIT_REF\" = \"feat/eng-1145-vercel-migration\" ]; then exit 1; else exit 0; fi"
+      skew_protection = "7 days"
+    },
+    storybook = {
+      description     = "Oak Web Application Storybook"
+      domains         = ["storybook-vercel.thenational.academy"]
+      build_type      = "storybook"
+      framework       = "storybook"
+      ignore_command  = "if [ \"$VERCEL_GIT_COMMIT_REF\" = \"feat/eng-1145-vercel-migration\" ]; then exit 1; else exit 0; fi"
```
tweakster, line 16: If we are going to go live we cannot have this line. We need to build for all branches.

tweakster (changes requested):

infrastructure/project/variables.tf
```diff
@@ -5,32 +5,33 @@ variable "cloudflare_zone_domain" {
...
+      NEXT_PUBLIC_CLERK_SIGN_IN_URL = optional(string)
+      NEXT_PUBLIC_CLERK_SIGN_UP_URL = optional(string)
+    }))
+    prod = optional(object({
+      OAK_CONFIG_LOCATION = optional(string)
+    }))
+    preview = optional(object({
+      OAK_CONFIG_LOCATION             = optional(string)
+      NEXT_PUBLIC_CLIENT_APP_BASE_URL = optional(string)
+    }))
```
tweakster, line 18: I assume the optional is because all variables aren't needed in all builds.

I'm wondering can we put a check in to make sure based on build type that the correct variables are set?

We want the Terraform to be documentation (in code) of how to build the apps. As it stands if we needed to rebuild either of these without any reference back to Vercel we wouldn't know which env vars are needed for each. Some code that checks if you have the env vars (and only the env vars) you need would enable us to recover with confidence the `env_var` variable has been set correctly.

infrastructure/project/variables.tf
```diff
@@ -17,6 +17,18 @@ variable "env_vars" {
...
   })
+
+  validation {
+    condition = alltrue([
+      for group in ["shared", "prod", "preview"] : alltrue([
+        for required_key in local.current_env_requirements[group] :
+        contains(keys(var.env_vars[group]), required_key) && 
+        var.env_vars[group][required_key] != null && 
+        var.env_vars[group][required_key] != ""
+      ])
```
tweakster, line 28: Does this make for simpler code (just comparing lists of keys):
```hcl
locals {
  env_map = {
    HELLO = "Hello",
    BYE   = "Bye",
    NULL  = null,
  }

  req_env = [ "HELLO", "BYE" ]
}

output "test" {
  value = tolist(keys({for k, v in local.env_map : k => v if v != null})) == sort(local.req_env)
}
```

tweakster (approved):

PR #3510, opened 2025-06-30
Feat/lesq 1387/media clips attributions [LESQ-1387]

njblore:
## Description

Music year: 1986

- add a new component `LessonMediaAttributions`
- conditionally rendered behind feature flag
- pulls in schema changes to get attributions from media clip objects

## How to test

*note: using beta pages to test means that linking between pages may be broken as these lessons are not published

1. Go to https://deploy-preview-3510--oak-web-application.netlify.thenational.academy/teachers/beta/lessons/exploring-characterisation-in-the-final-year/media 
2. With the feature flag disabled you should see no attributions 
3. With the feature flag enabled (https://eu.posthog.com/project/124/feature_flags/78774) you should see attributions at the bottom of the page
4. You can use [this doc](https://www.notion.so/oaknationalacademy/Test-Data-for-Copyright-Content-21526cc4e1b180ca9684f16b3d30653c) to find other lessons with media clip attributions to use for testing

## Screenshots

How it should now look:

Screenshot 2025-06-30 at 16 03 02

## Checklist

- [x]  When attributions are present and required for a media clip they are present on the page
- [x]  Behind the feature flag


src/components/TeacherViews/LessonMedia/LessonMedia.view.tsx
```diff
@@ -478,6 +479,20 @@ export const LessonMedia = (props: LessonMediaProps) => {
           <OakBox $display={["block", "block", "none"]}>
             {helpArticleLink}
           </OakBox>
+          <LessonMediaAttributions
+            mediaClipsWithAttributions={
```
Taha-Hassan-Git, line 483: Nit: I would put this logic in the child component

Taha-Hassan-Git (changes requested):
Behaviour is as-expected and tests look good! Just a small issue with the font weights, the designs have a bigger contrast: 600 for the bold text and 300 for the lighter text.

The other thing is the spacing around the attributions component isn't consistent with the designs. But I think you've intentionally left that because it would mean changing the whole layout, and then you'd have to put all those layout changes behind the feature flag also. Not worth it! But does it make sense to create a ticket to fix this when we retire the feature flag?

src/components/TeacherViews/LessonMedia/LessonMedia.view.tsx
njblore: yeah I could see that, I went with keeping the child dumb as it doesn't need to know that it's looking at media clip object metadata and what the logic is for showing attributions, it just needs to render what it's given. This way the component could be easily reused elsewhere with different data structures, and it keeps the places we're expecting logic and data manipulation to be to a minimum (in theory making the code easier to work with)

dkyyau (approved):

PR #3516, opened 2025-07-02
Fix/lesq 1439/save spam error toast [LESQ-1439]

njblore:
## Description

Music year: 2015

- Update the `save` button state to become disabled while the save action completes
- This should prevent people being able to spam the button

## Issue(s)

Fixes #
Error toast and out of sync save counts

## How to test

1. Go to https://deploy-preview-3516--oak-web-application.netlify.thenational.academy
2. Go to a unit or lesson listing page
3. Try to recreate the error by spam clicking a save button 
4. You should not see an error and the save count should stay consistent
5. Go to /my-library and do the same as above

### DIFF
--- a/CHANGE_LOG.md
+++ b/CHANGE_LOG.md
@@ -1,3 +1,5 @@
+## [1.789.3](https://github.com/oaknational/Oak-Web-Application/compare/v1.789.2...v1.789.3) (2025-07-04)
+
 ## [1.789.2](https://github.com/oaknational/Oak-Web-Application/compare/v1.789.1...v1.789.2) (2025-07-03)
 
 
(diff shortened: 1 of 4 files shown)


dkyyau (commented):
I don't think I'm able to recreate the spamming error but for some units in My Library, the save button isn't showing as `Saved`. It works for some other units though. This could be a separate issue.
image

Taha-Hassan-Git:
> thanks @Taha-Hassan-Git, I felt the risk of people spamming on my library was low 😅 but maybe better to update anyway just for consistency? 

Up to you, I don't have a strong opinion really.

jack-skerrett-oak (changes requested):
When using keyboard controls to toggle the button, focus is lost. On chrome, it isn't too bad because focus remains on the un-focusable element and if you tab again focus is visible however on safari it seems to move the user back to the top of the page.

jack-skerrett-oak (approved):

PR #3531, opened 2025-07-08
[ENG 1240] Owa Custom env - Staging

KofoworolaOgunleye:
## Description
- Rather than creating a separate project, we chose custom environment
for the OWA-staging build that is pointing to the staging Hasura instance

## Issue(s)

[ENG-1240](https://www.notion.so/oaknationalacademy/Create-owa-staging-on-Vercel-21626cc4e1b180dbabc8cfe39c6d71df)

## How to test
1. Cant test till migration is over and `ignored build step` is removed because deployments are skipped.

## Checklist

- [ ] Test after migration


infrastructure/project/variables.tf
```diff
@@ -3,6 +3,24 @@ variable "cloudflare_zone_domain" {
...
 }
 
+variable "custom_environments" {
+  description = "Name and domain for each custom Vercel environment"
+  type = list(object({
+    name   = string
+    domain = string
+  }))
+  default = []
+}
```
tweakster, line 13: Do we need this? Is this not in the build config?

infrastructure/project/locals.tf
```diff
@@ -56,5 +56,34 @@ locals {
...
+    [
+      for key, value in var.env_vars.shared : {
+        key                     = key
+        value                   = value
+        custom_environment_name = env
+      }
+    ]
+  ])
+
+  all_custom_env_vars = concat(local.custom_env_vars, local.custom_env_vars_shared)
```
tweakster, line 86: All of this feels very complicated. Do we need it all?

Can you not just transform `var.custom_env_vars` into a suitable object and use that? Am I missing something?

infrastructure/project/locals.tf
```diff
@@ -56,5 +56,30 @@ locals {
     ]
   ])
 
+  custom_env_names = [for ce in local.build_config.custom_environments: ce.name]
+
+  custom_env_vars=flatten([
+    for env_name, env_map in var.custom_env_vars : [
```
tweakster, line 62: It won't be a problem for this but it's possible that not all entries in `var.custom_env_vars` will be relevant. Only if they are also in `local.custom_env_names`.

infrastructure/project/main.tf
```diff
@@ -17,7 +17,7 @@ resource "terraform_data" "workspace_validation" {
 }
 
 module "vercel" {
-  source                 = "github.com/oaknational/oak-terraform-modules//modules/vercel_project?ref=v1.2.3"
+  source                 = "github.com/oaknational/oak-terraform-modules//modules/vercel_project?ref=ENG-1240-owa-staging-custom-env-"
```
tweakster, line 20: I assume this will be a version number before merging.

infrastructure/project/locals.tf
```diff
@@ -56,5 +56,30 @@ locals {
...
+      for key, value in env_map : {
+        key                     = key
+        value                   = value
+        custom_environment_name = env_name
+      }
+    ]
+  ])
+
+  custom_env_vars_shared = flatten([
```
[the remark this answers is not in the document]
KofoworolaOgunleye: it should, just needed to run terraform apply

tweakster (approved):

PR #3536, opened 2025-07-10
Fix/lesq 1414/unit filters count [LeSQ-1414]

njblore:
## Description

Music year: {owa_music_year}

- List of changes

## Issue(s)

Fixes #

## How to test

1. Go to {owa_deployment_url}
2. Click on \_\_\_\_\_\_

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist

- [ ] Design sign-off


src/components/TeacherComponents/UnitsLearningThemeFilters/UnitsLearningThemeFilters.test.tsx
```diff
@@ -41,23 +41,17 @@ describe("UnitsLearningThemeFilters", () => {
...
     );
     const themes = await screen.findAllByRole("radio");
     expect(themes[0]).toBe(screen.getByLabelText("All"));
     expect(themes[1]).toBe(screen.getByLabelText("Algebra"));
     expect(themes[2]).toBe(screen.getByLabelText("No theme 1"));
   });
-  test("should call callback method with correct value", async () => {
-    const onChangeCallback = jest.fn();
+  test("should call setTheme method with correct value", async () => {
+    const setTheme = jest.fn();
```
dkyyau, line 54: This is very minor but would it make sense to have this at the top of the file as it's used in all of the tests?

Taha-Hassan-Git (changes requested):
Works well, much prefer the shared logic between the two components.

I found an edge case where after a refresh, the url state of the filters isn't making its way into the new submitted query.

Kapture 2025-07-14 at 11 18 23

Not something for this PR, but it took me a while to realise that the unit list only updating on submit was the expected behaviour. I'm much more used to filters updating dynamically, so it seemed like a bug.

src/components/TeacherComponents/YearGroupFilters/YearGroupFilters.test.tsx
```diff
@@ -27,111 +35,62 @@ describe("YearGroupFilters", () => {
...
+          setYear={jest.fn()}
         />
       </OakThemeProvider>,
     );
     expect(getByText("Year")).toBeInTheDocument();
   });
 
-  it("updates the router query and selected year on click", () => {
+  it("calls set year year on click", async () => {
+    const mockSetYear = jest.fn();
```
dkyyau, line 55: Same with this one and the other tests in this file

dkyyau (approved):
Looks good and functionality working as expected

njblore:
thanks @Taha-Hassan-Git well spotted! should be fixed now 🤞 and yeah I can see why the drawer behaviour could be confusing now that you mention it, I think it makes a bit more sense on mobile as you couldn't see the updates anyway so may as well batch them 🤷‍♀️

PR #3564, opened 2025-07-23
Fix/lesq 1510/lesson level restrictions [LESQ-1510]

njblore:
## Description

Music year: {owa_music_year}

- update the beta lessons query so that unit data is not required for previewing unpublished lessons
- update lesson query to use loginRequired and geoRestriction from features rather than the MV

## Issue(s)

Fixes #
Mismatch in restrictions between features and lesson MV

## How to test

1. Go to https://oak-web-application-website-eztylo8gg.vercel-preview.thenational.academy/teachers/beta/lessons/performing-there-are-things-that-lurk-in-the-library-by-joseph-coelho
2. While signed out: you should see a banner that you will need to sign in and not be able to access lesson resources
3. While signed in: if outside the uk you should see a georestricted banner
4. Other lessons that are / aren't currently restricted should remain the same

## Screenshots

Screenshot 2025-07-23 at 10 40 43

How it should now look:
Screenshot 2025-07-23 at 10 40 59
Screenshot 2025-07-23 at 10 41 30


njblore:
@Taha-Hassan-Git yeah to explain why it was removed if someone wants to add it back & add some context if something is breaking due to invalid unit data

jack-skerrett-oak (approved):
I can't 100% confirm this is better than it was because the content I wanted to see is only available on this PR (this new unit state thing you did means we can now see some of the restricted media clips content on staging) but every lesson that I check that should be restricted is labelled correctly. 

I will look at writing a ticket for potential clean up of the lesson mv

PR #3567, opened 2025-07-23
Fix/lesq 1512/keyboard nav sign up [LESQ-1512]

njblore:
## Description

Music year: {owa_music_year}

- Refactor the sign up link in the banner to use a button element internally, as an `a` tag without a href is not tabbable and we only need it for the styling

Sorry I also changed a few unrelated things that I noticed while in the area 🫣 
- I noticed an issue with non-onboarded users in the banner as well, they'd be served the `geoblocked` version of the banner despite not having onboarded, so I fixed that
- I also refactored the onboarding page link config to accept a returnTo query param to unify how we were applying this in a few places

## Issue(s)

Fixes #
Sign up link not reachable by keyboard

## How to test

1. Go to https://oak-web-application-website-9t01dcp6b.vercel-preview.thenational.academy/teachers/programmes/english-secondary-ks3/units/the-twisted-tree-fiction-reading/lessons/the-twisted-tree-writing-an-article-about-mormors-death?sid-a7d6d9=Xvp28KrPnT&sm=0&src=4#lesson-details
2. While signed out, you should be able to tab to the sign in link
3. Signing in should redirect you through to onboarding and then back to this page again still
4. After signing in and **not onboarding** you should see the same banner on this page, but the sign in link goes to onboarding

## Screenshots

How it should now look:
{screenshots}
Screenshot 2025-07-23 at 15 18 43


Taha-Hassan-Git (approved):
🚀👍

PR #3578, opened 2025-07-28
fix: only use auth token when required by lesson restrictions [LESQ-1530]

njblore:
## Description

Music year: {owa_music_year}

- change prop to download fn to authRequired and use lesson level restrictions as well as the feature flag to determine if auth is required
- the downloads api already uses lesson level restrictions to determine whether auth is required

## Issue(s)

Fixes #
Signed out users unable to download when the copyright flag is enabled

## How to test

1. Go to {owa_deployment_url}
2. While signed out you should be able to download lessons without restrictions
3. While signed in you should be able to download units and lessons with restrictions


dkyyau (approved):
Tested signed out and signed in and looks to be working as expected

Taha-Hassan-Git (approved):
😳

PR #3592, opened 2025-07-30
Fix: onboarding request failed

Taha-Hassan-Git:
## Description

I couldn't get to the bottom of what was causing the [401 bugs](https://app.bugsnag.com/oak-national-academy/oak-web-application/errors/6787b5f3aefec07c64d3d139?filters[error.status]=open&filters[event.since]=30d&pivot_tab=browser.version), but I think it's probably not a bad idea to check for auth on the onboarding route anyway. If somehow a user ends up here without being signed in, they will be redirected to `sign-up`. 

## Issue(s)

Sort of Fixes [#LESQ-1474](https://www.notion.so/oaknationalacademy/Bugsnag-onboarding-request-failed-21526cc4e1b18011856dd18803fe205d)

## How to test

1. Go to https://oak-web-application-website-ifoxma6ox.vercel-preview.thenational.academy
2. When signed out, navigate to `/onboarding`
3. You should be redirected to sign in

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist

- [ ] Design sign-off


njblore (commented):
this approach is exactly right but we have a wrapper component that does the work for us already! sorry that's my bad for not going through this / adding it to the ticket 🙈 it's called `WithPageAuthRequired` and you can see an example of how it's used in `my-library.tsx`

njblore (approved):
Nice, we should probably wrap the sub routes for onboarding as well just to be safe

PR #3596, opened 2025-08-02
[ENG 1244] Refactor sensitive env vars

KofoworolaOgunleye:
## Description

-  Remove ignore build and git branch since migration to vercel is complete, `feat/eng-1145-vercel-migration` has been merged into main
- Refactor sensitive env vars to individual variables for easier management. Currently, we need to know all the values of sensitive env var object when we add/update a var because TFC hides sensitive values. This ticket breaks down the monolithic senstive env vars object into individual variables, making it easier to update single sensitive values without knowing all others

## Issue(s)

[ENG-1244](https://www.notion.so/oaknationalacademy/7a067876b8d94be0b4100d6c831d5a1a?v=450b96370cb34a7dacd6994b6a8540c4&p=21826cc4e1b1804897f6ea2237d0fab0&pm=c)

## How to test

1.Terraform plan should show no changes

### DIFF
--- a/infrastructure/project/builds.tf
+++ b/infrastructure/project/builds.tf
@@ -22,8 +22,6 @@ locals {
       build_type       = "storybook"
       deployment_type  = "all_deployments"
       framework        = "storybook"
-      git_branch       = "feat/eng-1145-vercel-migration"
-      ignore_command   = "if [ \"$VERCEL_GIT_COMMIT_REF\" = \"feat/eng-1145-vercel-migration\" ]; then exit 1; else exit 0; fi"
       output_directory = "storybook-static"
       skew_protection  = "1 day"
     }
--- a/infrastructure/project/locals.tf
+++ b/infrastructure/project/locals.tf
@@ -45,14 +45,28 @@ locals {
     ]
   ])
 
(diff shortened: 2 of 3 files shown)


infrastructure/project/main.tf
```diff
@@ -16,6 +16,20 @@ resource "terraform_data" "workspace_validation" {
...
+    precondition {
+      condition = alltrue([
+        for gk, gv in local.required_current_sensitive_env : alltrue([
+          toset(keys({ for k, v in local.sensitive_env_vars[gk] : k => v if v != null })) == toset(gv)
+        ])
+      ])
+      error_message = "Sensitive environment variables don't match requirements for '${local.build_type}' build. Required: ${jsonencode(local.required_current_sensitive_env)}"
+    }
+  }
+}
```
tweakster, line 31: Did you try doing these per variable? i.e. add a conditional to each `variable` and check
* If it's required and null error
* If it's not required and not null error
I think you will get better error messages this way. The above won't be able to specify which variable is missing/not required but a per `variable` statement will.

tweakster (approved):

PR #3605, opened 2025-08-05
Fix/get country code from vercel header

njblore:
## Description

Music year: 1955

- update the request header used to get country code for geolocating users in onboarding
- the header key is different on vercel so we are currently setting all new users to `isRegionAuthorised: false` 

## Issue(s)

Fixes #
Failing to geolocate all new users

## How to test

1. Go to https://oak-web-application-website-lqlnrmnnb.vercel-preview.thenational.academy
2. Sign up with a new user and onboard
3. You should be region authorised


jack-skerrett-oak (approved):
Works via google, microsoft and email.
Works if onboarding interrupted
Works cross browser

PR #3622, opened 2025-08-13
build: specify vercel regions preferences

jimCresswell:
Changing the region that we prefer to run server-side logic in Vercel

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -1,6 +1,7 @@
 {
+  "$schema": "https://openapi.vercel.sh/vercel.json",
+  "regions": ["lhr1", "eu-west-1", "cdg1"],
   "github": {
-    "enabled": true,
-    "silent": false
+    "enabled": true
   }
 }


vercel.json
```diff
@@ -1,6 +1,7 @@
 {
+  "$schema": "https://openapi.vercel.sh/vercel.json",
+  "regions": ["lhr1", "dub1", "cdg1"],
   "github": {
-    "enabled": true,
-    "silent": false
+    "enabled": true
```
jack-skerrett-oak, line 5: I was just looking at what these github configs did because I wasn't aware, docs say they have been deprecated: https://vercel.com/docs/project-configuration/git-configuration#github.enabled

jimCresswell: Nice catch! I think we can delete the github key entirely, it was a very old attempt to control the verbosity on feedback in the PRs... please make that change @Jack! Get your coding on

jack-skerrett-oak: I am not comfortable making changes when I do not know the full impact, nor when it is not my primary focus.

jimCresswell: Fair enough! I have pushed the change

vercel.json
```diff
@@ -1,6 +1,4 @@
 {
-  "github": {
-    "enabled": true,
-    "silent": false
-  }
+  "$schema": "https://openapi.vercel.sh/vercel.json",
+  "regions": ["lhr1", "dub1", "cdg1"]
```
orangemug, line 3: Out of interest why these three? Why not also `fra1` for instance?

KofoworolaOgunleye (approved):

PR #3627, opened 2025-08-14
Fix/lesq 1462/onboarding form roles [LESQ-1462]

njblore:
## Description

Music year: 2022

- use fieldset and legend elements for onboarding form
- there's some kind of bug with how testing library is interpreting the dom when the fieldset and legend are added (see https://github.com/testing-library/user-event/issues/662) , so the continue button is seen as having 'pointer-events: none' causing tests to fail. I got around this by telling it to not check for pointer events before clicking using `pointerEventsCheck` param in click events ([see docs](https://testing-library.com/docs/user-event/options/#pointereventscheck))

## Issue(s)

Fixes #
Invalid roles on div elements.
Adds sign up and sign in pages to pa11y, but onboarding pages are now behind a login so we can't add those. We will look into alternative ways to get visibility on changes in these pages.

## How to test

1. Go to https://oak-web-application-website-xrkld6c2f.vercel-preview.thenational.academy/onboarding
3. You should be able to go through the onboarding journey using mouse or keyboard
4. pa11y checks should pass


RBrownsett (approved):
Approved - ran axe core Dev Tools - fieldset role has been removed and native fieldset has been used, error has now gone. Checked with voice over and navigation with keyboard. No issues.

PR #3635, opened 2025-08-15
Fix/lesq 1461/search result a11y [LESQ-1461]

njblore:
## Description

Music year: 1972

- refactor dropdown button so the aria label is applied to the card, which is the button, instea of the p tag
- add tests for extracted fn
- add the search page back into pa11y urls

## Issue(s)

Fixes #
Invalid aria label on p tag

## How to test

1. Go to https://oak-web-application-website-7tswqfe1n.vercel-preview.thenational.academy/teachers/search
2. Search for something with variants, eg `macbeth`
3. You should see the result cards with dropdowns contain an aria label
4. The dropdown text does not contain an aria label
5. The labels are accurate

## Screenshots

Screenshot 2025-08-15 at 13 49 12

Screenshot 2025-08-15 at 13 49 39

### DIFF
--- a/CHANGE_LOG.md
+++ b/CHANGE_LOG.md
@@ -1,3 +1,16 @@
+# [1.834.0](https://github.com/oaknational/Oak-Web-Application/compare/v1.833.0...v1.834.0) (2025-08-18)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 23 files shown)


dkyyau (approved):
Tested with different searches and aria-label is on the button and labels are correct

PR #3651, opened 2025-08-27
Feat/lesq 1508/promote copyright feature flag [LESQ-1508]

njblore:
## Description

Music year: 2011

- remove authRequired param from download requests
- remove isBehindFeatureFlag param from useCopyright hook and buttons that use the hook

## How to test

1. Go to https://oak-web-application-website-git-feat-lesq-1508promote-co-b14c81.vercel-preview.thenational.academy/
2. Go to a unit while signed out
3. You should see the unit download button (if that button also exists on prod, ie. the unit has been zipped)
4. The unit download button should direct you through sign in
5. You should be able to download the unit once signed in
6. Lesson download functionality should be unchanged
7. *Further testing needs to be done against copyrighted lessons but none are live on the platform right now so this will have to wait

### DIFF
--- a/CHANGE_LOG.md
+++ b/CHANGE_LOG.md
@@ -1,3 +1,11 @@
+## [1.847.2](https://github.com/oaknational/Oak-Web-Application/compare/v1.847.1...v1.847.2) (2025-09-02)
+
+
+### Bug Fixes
+
+* added role and labelledby to multi answer checkboxes questions in quizzes ([3408de7](https://github.com/oaknational/Oak-Web-Application/commit/3408de78d6ee32f1b51e5a53a652b95d0257f582))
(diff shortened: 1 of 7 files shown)


dkyyau (approved):
Looks good, tested unit and lesson downloads for signed in and out 👍

Taha-Hassan-Git (approved):

RBrownsett:
> thanks @RBrownsett, yes it could be the feature flag loading as it's still active, but it looks like it's still loading your account in the screen shot as the sign in button is still there, so that could be why also?

Ah yes, it looks like the account loads when going to certain pages (the same happens in live) - I could raise a bug ticket for us to investigate in the future if you think its worth time looking into 


https://github.com/user-attachments/assets/ecb74d93-af89-4947-acda-7ca7ba2e262f

njblore:
@RBrownsett yeah let's raise a ticket, not necessarily a bug but it would be good to get a better loading state

PR #3688, opened 2025-09-17
`useFeatureFlag()` now just returns value and does type assertion on returned value

orangemug:
## Description
`useFeatureFlag()` now just returns value and does type assertion on returned value

Safe usage is now

```ts
const flagBooleanValue = await useFeatureFlag("some-flag", "boolean");
const flagStringValue = await useFeatureFlag("some-flag", "string");
```

## Issue(s)
n/a

## How to test
Test https://oak-web-application-website-fr0s92pqx.vercel.thenational.academy/timetabling/new is only shown when feature flag is enabled.

## Checklist

- [ ] Design sign-off

### DIFF
--- a/src/app/(beta)/timetabling/name/page.test.tsx
+++ b/src/app/(beta)/timetabling/name/page.test.tsx
@@ -8,13 +8,13 @@ jest.mock("src/utils/featureFlags");
 
 describe("/timetabling/units", () => {
   test("when enabled", async () => {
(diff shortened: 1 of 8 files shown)


rightsaidjames (approved):
Regression tested - works exactly as before! As discussed, I can't directly test the string-based feature flag scenario, but I'm confident that the unit tests have our backs covered on that front.

PR #3714, opened 2025-09-25
Feat/ai 1636/fuzzy search

njblore:
## Description

Music year: 1973

This turned into non-fuzzy term matching :) 

- update curriculum data to add `reception` and `all-years` to `OAK_YEARS` and `eyfs` to `OAK_KEYSTAGES`
- adds a new function `findPfMatches` which will get any matching subject, keystage, examboard or year programme factor from a given search term
- add a function `getMatch` which handles the actual string matching of the term against a given library of curriculum data
- add a function `rankMatches` which will do a basic ranking of matches when multiple are found (this is mostly courtesy of Claude so lmk if anything looks funky!)
- add a function `getIsAmbiguousMatch` which will instead return no matches if the overall search term is not sufficiently covered by pf matches 
- adds a boat load of tests, possibly overkill but uncovered and resolved some edge cases by doing so


src/context/Search/suggestions/match/getIsAmbiguousMatch.test.ts
```diff
@@ -0,0 +1,28 @@
...
+      subject: "history",
+    });
+    expect(result).toBe(false);
+  });
+  it("returns false when there is 1 extra word", () => {
+    const result = getIsAmbiguousMatch("golf history", {
+      subject: "history",
+    });
+    expect(result).toBe(false);
+  });
```
codeincontext, line 21: I think this is a really useful test file as it highlights the intended behaviour well

When we had the subject card as a feature, we had planned to only show it if there's a direct intent. The example string here don't capture that very well as I don't think that "history of holf is a direct subject query". Can we think of some genuine inputs that would be unambiguous for 1-2 extra words?

Maybe something like "maths curriculum"

src/context/Search/suggestions/match/getMatch.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export const getMatch = (
+  query: string,
+  data: CurriculumData[],
+): MatchResult | undefined => {
+  const matches: Array<MatchResult> = [];
+
+  data.forEach((datum) => {
+    const slugRegex = new RegExp(datum.slug, "i");
+    const matchesSlug = query.match(slugRegex);
```
codeincontext, line 15: In case you hadn't considered it, another way to do this is `slugRegex.toLowerCase === datum.slug.toLowerCase`

I don't think that either approach is better or worse so long as we trust the input data. Building regexes from user data could potentially allow a ReDoS which I know SonarCloud is keen to flag up!

src/context/Search/suggestions/match/rankMatches.test.ts
```diff
@@ -0,0 +1,85 @@
...
+      "maths",
+    ],
+    [
+      "year 10",
+      [
+        { slug: "year-10", matched: "year 10" },
+        { slug: "year-11", matched: "year 11" },
+      ],
+      "year-10",
+    ],
```
codeincontext, line 20: I personally find this quite hard to read. I need to jump from the test to the data to understand what each string represents. Given that the test code is so small, have you considered just having two tests? I don't even think it would be fewer lines

If this is a standard practice at Oak then it's totally fine and something I can adjust to

src/context/Search/suggestions/match/getMatch.ts
codeincontext: I've just realised that you're matching for substrings so in that case it would obviously be `query.toLowerCase().includes(datum.slug.toLowerCase())`

src/context/Search/suggestions/match/rankMatches.test.ts
codeincontext: Please ignore this for now. I can see how it makes context in the rest of the tests where it's a less clunky example

src/context/Search/suggestions/findPfMatch.test.ts
```diff
@@ -0,0 +1,249 @@
...
+  });
+
+  it("should return null for queries that don't match any subject", () => {
+    expect(findPfMatch("xyzzyx")).toBeNull();
+  });
+
+  it.each([
+    "the history of mathematics in ancient greece",
+    "the geography of the nile river",
+    "history world war 2",
```
codeincontext, line 24: I'm assuming that `history ww2` would match?

src/context/Search/suggestions/findPfMatch.ts
```diff
@@ -0,0 +1,63 @@
...
+  if (
+    !parsedSubject.data &&
+    !parsedKeystage.data &&
+    !parsedYear.data &&
+    !parsedExamboard.data
+  ) {
+    return null;
+  }
+
+  const isAmbiguousMatch = getIsAmbiguousMatch(query, {
```
codeincontext, line 48: It took me a moment to understand what this function is doing. To me, ambiguous would imply that there are multiple results that could be correct - such as matching two subjects

I wonder if we could name it more literally like "hasNonMatchingWords", "hasExtraWords", etc?

codeincontext (commented):
Reviewed and left comments. I wonder if the ranker is a hint that substring matching is overcomplicated. I would like to try word-by-word matching with the same test suite. If using a regex, that would be using `\b`; or you could do direct text comparisons using `split("\n")`

**UPDATE: I didn't think about multi-word terms like "key stage 2". So there's complexity in both options. I'm happy to stick with this approach**

src/context/Search/suggestions/match/getMatch.ts
```diff
@@ -0,0 +1,40 @@
...
+export type MatchResult = { slug: string; matched: string };
+
+export const getMatch = (
+  query: string,
+  data: CurriculumData[],
+): MatchResult | undefined => {
+  const matches: Array<MatchResult> = [];
+
+  data.forEach((datum) => {
+    const slugRegex = new RegExp(datum.slug, "i");
```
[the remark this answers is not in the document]
njblore: I think we decided against this for now so will resolve, reopen if I'm mistaken!

src/context/Search/suggestions/match/getIsAmbiguousMatch.test.ts
codeincontext: I'm going to be busy this morning, but Joe might be able to point us at the logic/rubric. I think that we would probably want to allow either 1 or 0 extra words to be considered a direct match. Maybe we could inform it with the CSV of popular search terms

njblore: cc @JBR90

JBR90 (approved):

PR #3716, opened 2025-09-25
OpenAI subject suggestions

Taha-Hassan-Git:
## Description

Replaces the Dummy AI data with an api call to open ai using their sdk.

- Model returns an array of subject slugs with a confidence score
- output is parsed and validated with a zod schema using hard-coded curriculum data
- response is then sorted by confidence score, transformed to the correct format and returned
- tests for the above behaviour

## Issue(s)

[Fixes #AI-1601](https://www.notion.so/oaknationalacademy/AI-generate-subject-suggestions-with-OpenAI-27026cc4e1b180b1bddcda4b4d268ff5)

## How to test

1. make a `curl` or Postman request to https://oak-web-application-website-git-feat-ai-1601openai-subje-ef6d26.vercel.thenational.academy/api/search/intent/
2. include a `searchTerm` query param with something like "magnets" or "frogs"
3. You should get a response with this structure:

```json
{
    "directMatch": null,
    "suggestedFilters": [
        {
            "type": "subject",
            "slug": "physics"
        },
        {
            "type": "subject",
            "slug": "science"
        }
    ]
}
```

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist

- [ ] Design sign-off


src/pages/api/search/intent.ts
```diff
@@ -68,4 +71,71 @@ const handler: NextApiHandler = async (req, res) => {
...
+Examples:
+- "maths" → {"subjects": [{"name": "maths", "confidence": 5}]}
+- "shakespeare" → {"subjects": [{"name": "english", "confidence": 5}, {"name": "drama", "confidence": 3}]}
+- "ww1" → {"subjects": [{"name": "history", "confidence": 5}]}
+- "photosynthesis" → {"subjects": [{"name": "biology", "confidence": 5}, {"name": "science", "confidence": 4}]}
+
+Analyze the user's search term:
+`;
+
+  const subjectsEnum = z.enum(subjects as [string, ...string[]]);
```
JBR90, line 102: I think if you make subjects 'as const' it makes it a string literal and you don't need to use as.

src/pages/api/search/intent.ts
```diff
@@ -68,4 +71,71 @@ const handler: NextApiHandler = async (req, res) => {
...
+    model: "gpt-5-nano",
+    input: [
+      {
+        role: "system",
+        content: systemContent,
+      },
+      {
+        role: "user",
+        content: `
+        <retrieved-data>
```
JBR90, line 114: Nice!

src/pages/api/search/intent.ts
codeincontext: I don't think that you could use `as const` as it's imported from OAK_SUBJECTS

But I agree that this would make the type mismatch the schema which isn't ideal. Presumably we have a type which describes the available subject slugs?

If it needs a readonly type then maybe the type of `subjects` should be `Readonly<OAK_SUBJECTS["slug"]>`?

Taha-Hassan-Git: Sorry getting quite tangled up here. Maybe I'm missing something but I can't figure out a way of inferring the type for the subject slugs without writing the slugs out again in a new array. I could do that and then narrow the type for "slug" on the CurriculumData object? 
There is also [this type in oak-curriculum schema](https://github.com/oaknational/oak-curriculum-schema/blob/b51360ba77eac14c675e98b9b91ffd9ea7cd2cbb/src/schema/base.schema.ts#L176) but it has about 20 extra slugs in it.

codeincontext: If it ends up too complicated, I would use `z.array` rather than enum, then we can make sure it's in our list of correct subjects in the following code

Am I right that we have that fixture now with the hardcoded subjects? Can we extract a type from that?

src/pages/api/search/intent.ts
```diff
@@ -68,4 +83,38 @@ const handler: NextApiHandler = async (req, res) => {
...
+              slug: z.string(),
+              confidence: z.number().min(1).max(5),
+            }),
+          ),
+        }),
+        "subjects",
+      ),
+    },
+  });
+  const parsedResponse = response.output_parsed;
```
njblore, line 107: is the expectation that this will be null/undefined if the response doesn't match the schema passed in above?

src/node-lib/getServerConfig.ts
```diff
@@ -246,6 +246,20 @@ const envVars = satisfies<Record<string, EnvVar>>()({
...
+    value: process.env.OPENAI_API_KEY,
+    envName: "OPENAI_API_KEY",
+    required: true,
+    availableInBrowser: false,
+    default: null,
+  },
+  searchEnabledInProd: {
+    value: process.env.SEARCH_ENABLED_IN_PROD,
+    envName: "SEARCH_ENABLED_IN_PROD",
+    required: true,
```
codeincontext, line 259: Is this is required in all environments, then it might make sense for it just to be `AI_SEARCH_ENABLED` and then it can be set to true,, true, and false in dev, staging, and prod environments. I've pinged you a message on slack with more detail

I don't think it's that important and doesn't have to block this PR

src/pages/api/search/intent.ts
```diff
@@ -53,7 +59,16 @@ const handler: NextApiHandler = async (req, res) => {
       return res.status(200).json(payload);
     }
 
-    const payload = searchIntentSchema.parse(DUMMY_AI_RESPONSE);
+    const modelResponse = await callModel(searchTerm);
```
codeincontext, line 62: The name `modelResponse` implies to me that its a structured response of some kind. Looking at the  `modelResponse.map` below, it seems to just be a list of subjects, in which case we could call it something like `subjectsFromModel`

src/pages/api/search/intent.ts
codeincontext: Response has an `error` key,. I believe that `output_parsed` could be null when the model fails to generate a response. I think that in those cases we should report the error. We can decide whether to throw or ignore it for the user, but I think we should make it visible to us

If we find that it happens in the general course of using the LLM, we could decide to put a retry here first - but thats a long way down the line

scripts/build/write_env_file/index.ts
```diff
@@ -193,6 +193,12 @@ async function main() {
     DISABLE_ISR: process.env.DISABLE_ISR,
     NEXT_PUBLIC_SENTRY_ENABLED: process.env.NEXT_PUBLIC_SENTRY_ENABLED,
 
+    // Search API
```
njblore, line 196: I'd call this section `AI features` or something similar because there are other search api env vars above and it might be confusing

njblore (approved):
left a few comments but looks good overall! 🚀

src/pages/api/search/intent.ts
```diff
@@ -70,4 +74,51 @@ const handler: NextApiHandler = async (req, res) => {
...
+    });
+    reportError(error);
+  }
+
+  const validSubjects =
+    parsedResponse?.subjects?.filter((subject) =>
+      subjects.includes(subject.slug),
+    ) ?? [];
+
+  const sortedResponse = validSubjects.sort(
```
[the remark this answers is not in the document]
Taha-Hassan-Git: Good tip! Thanks.

Taha-Hassan-Git:
Merged main, and fixed an issue that came up. 

I got to try out both back and frontend systems working together. 

The problem with the below is twofold:
1. "Poetry" should obviously be returning the response "english", the model seems unable to match it for some reason 
2. When the model returns an empty array, we shouldn't show the suggested filters view at all.

I'm conscious that Tash has a [WIP over here](https://github.com/oaknational/Oak-Web-Application/pull/3736) that is already diverging significantly, So these make sense as follow up fixes. We can bundle fix#1 with the AI gateway issue. For fix #2  worth adding a ticket, or adding to a ticket @JBR90?

Screenshot 2025-10-03 at 09 35 06

PR #3736, opened 2025-10-02
Feat/ai 1645/get user intent

njblore:
## Description

Music year: {owa_music_year}

- List of changes

## Issue(s)

Fixes #

## How to test

2. Click on \_\_\_\_\_\_

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist

- [ ] Design sign-off


RBrownsett:
@njblore looking good, the direct intent filters seem to be pulling through as filters. Some observations that might be useful for the future:
The term **year 10 online safety** pulls out the year and suggests RHSE as the subject filter. Although the first few results returned are Computing
image

Some words are there own do not produce suggested filters e.g. online - in those cases we might want to have the old filters already expanded?

src/pages/api/search/intent/index.ts
```diff
@@ -0,0 +1,79 @@
...
+        directMatch: pfMatch,
+        suggestedFilters: getSuggestedFiltersFromSubject(
+          pfMatch.subject.slug,
+          pfMatch,
+        ),
+      };
+      return res.status(200).json(payload);
+    } else if (aiSearchEnabled) {
+      const subjectsFromModel = await callModel(searchTerm);
+      const bestSubjectMatch = subjectsFromModel[0]?.slug;
```
Taha-Hassan-Git, line 43: If we're only going to use the best subject match, then we should probably update the prompt to only request one subject match. I suspect that'll make the task easier and perhaps save some money on reasoning tokens.

Taha-Hassan-Git (approved):
Looks good to me! Just one comment and not relevant to this PR, just for future work.

codeincontext (approved):
Looking great 👍

src/pages/api/search/intent/index.ts
njblore: this is just to get appropriate keystage and examboard filters for the best matched subject, we'll still return a second subject match if it's found

Taha-Hassan-Git: Oh gotcha! Thanks!

PR #3737, opened 2025-10-02
Feat/posthog cookieless tracking

njblore:
## Description

The goal of the change is to allow feature flags to load before consent has been accepted, or when consent has been rejected

- initialise posthog regardless of cookie consent status
- add `cookieless_mode` flag to init, so posthog wont create a cookie until we explicitly opt in to tracking
- opt in when cookies have been accepted
- opt out when cookies rejected

## Issue(s)

Fixes #
Feature flags don't load for users who have opted out of / not accepted analytics cookies

## How to test

2. As a fresh user / not signed in / no cookies
3. You should retrieve a feature flag value (example flag to be given)
4. Click reject cookies
5. You should see no posthog cookie added
6. You should still see feature flagged content
7. Analytics event should contain no distinct id

## Screenshots

Test flag value enabled while cookies not accepted:
Screenshot 2025-10-02 at 15 53 27

No distinct id:
Screenshot 2025-10-02 at 15 54 12

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -56,7 +56,7 @@
         "node-fetch": "^2.6.7",
         "path-browserify": "^1.0.1",
         "path-to-regexp": "^6.2.1n",
-        "posthog-js": "^1.214.0",
+        "posthog-js": "^1.269.0",
         "posthog-node": "^4.10.1",
         "querystring-es3": "^0.2.1",
         "react": "^18.3.1",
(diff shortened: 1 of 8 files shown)


RBrownsett:
@njblore I get a permission error when trying to view your screenshots - the one I'm interested in `No distinct id:` where would I find that value? Is it in the avo page event or the local storage. This is an example of pageview event I'm seeing
image

PR #3744, opened 2025-10-08
feat: searchintent rate limit [AI-1631]

dkyyau:
## Description

- Adds rate limiting using Upstash
- The rate limit is just for requests to AI suggestions
  - If the response is cached (after #3746), it won't hit the rate limit
  - If the response is a direct text match for a subject, no AI is needed and it won't hit the rate limit
- We use an env var, AI_SEARCH_RATE_LIMIT_PER_24H to set the limit per day
  - It's set to 5 per 24 hours for this branch in the Vercel branch config

Note: The Upstash redis db needs to be upgraded from the free tier but can be currently tested with the env vars found in the `owa-search-api-experiment` db in Vercel Storage

## Issue(s)

Fixes AI-1631

## How to test

1. Go to the search page
2. Make a search that would trigger API. This is a search term that doesn't directly match the name of a subject
3. Submit variants of that term. eg: "shakespeare 1", "shakespeare 2", "shakespeare 3"
4. After 5 requests, you should be degraded to the old filter UI without suggested filters

## Checklist

- [ ] Design sign-off


codeincontext:
I've pushed some changes to this PR. We need to verify that x-forwarded-for is available to us in Vercel, as it's [stripped by default](https://vercel.com/docs/headers/request-headers#x-forwarded-for). One alternative is to use cloudflare ratelimiting, like we have on the downloads endpoint. That should "Just Work", and would save us from needing this custom code. I know that's a bit of a wildcard suggestion given the work we've already put into this approach

New changes:
- Require an IP to be detected, and throw if no IP was extracted
  - This will prevent false positives where it looks like it's working but it actually broken
- Make the rate limiter util accept custom limit objects
  - This makes the separation of concerns really clear. The code calling the rate limiter can specify the prefix and limits it needs
- Make rate limits configurable through an env var
  - This lets us change the values without a code change. It also lets us set values per environment or even per branch for testing
- Specify a redis key prefix
  - We will be using the same redis instance for other functionality such as caching. It makes sense to explicitly set a relevant name
- Wait for background tasks to complete on Vercel
  - The upstash docs recommend using [waitUntil](https://upstash.com/docs/redis/sdks/ratelimit-ts/gettingstarted#serverless-environments) to ensure that it's been able to tidy up its work before the process ends. We don't actually need this unless we use multi-region config or analytics. But it's a safe bet to add
- Explicit env var use rather than `Redis.fromEnv()`. It's inevitable that someone will search the project for the env vars, so it's useful to name them explicitly

PR #4158, opened 2026-04-13
Switch from `npm` to `pnpm`

orangemug:
## Description
Switch our package manager from `npm` to `pnpm`

Changes

 - Change from `actions/setup-node` to `pnpm/action-setup`
 - All the scripts moved from `npm` to `pnpm`
 - Added "unlisted dependencies", deps we directly import but didn't have in our `package.json` (pnpm requires this)
 - Added `minimumReleaseAge`/`minimumReleaseAgeExclude`
 - Added `allowBuilds`, currently all of the modules that were building previously (maybe we can reduce this) 

## Todo

 - ~~It seems to be taking a long time installing chrome in CI `.../node_modules/puppeteer postinstall: Chrome (121.0.6167.85) downloaded to /home/runner/.npm/chromium/chrome/linux-121.0.6167.85`~~ — no longer an issue
 - ~~Percy still seems to be failing~~ — fixed

## Issue(s)
Partially implements `FEC-150`

## How to test
Code review + would appreciate if anyone can see scripts/old refs to `npm` I might have missed.

QA please smoke test in the same way you would dependagames updates.

## Checklist

- [ ] Design sign-off


.github/workflows/code_checks.yml
```diff
@@ -22,22 +22,22 @@ jobs:
         with:
           fetch-depth: 0
       - name: Use Node.js
-        uses: actions/setup-node@v4
+        uses: pnpm/action-setup@08c4be7e2e672a47d11bd04269e27e5f3e8529cb #v6
         with:
-          node-version: 22.x
-          cache: "npm"
+          version: 10
+          cache: true
```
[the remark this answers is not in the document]
orangemug: This doesn't look correct to me, can someone else double check the docs for me. I think AI is having a wobble

package.json
```diff
@@ -282,7 +306,8 @@
   },
   "engines": {
     "node": "22",
-    "npm": ">=10.9 <11"
+    "npm": ">=10.9 <11",
```
jimCresswell: it might be better to remove this, to avoid giving the impression that either is fine?

infrastructure/project/builds.tf
```diff
@@ -21,7 +21,7 @@ locals {
     storybook = {
       description        = "Oak Web Application Storybook"
       domains            = ["storybook.thenational.academy"]
-      build_command      = "npm run storybook:build"
+      build_command      = "pnpm run storybook:build"
```
rjkward, line 24: just to highlight that this will require a `terraform apply` otherwise Vercel will continue to build the storybook with npm

rjkward: diff looks fine but please reach out if you have any questions

src/components/TeacherComponents/LessonShareCardGroup/LessonShareCardGroup.test.tsx
```diff
@@ -55,34 +54,6 @@ describe("lesson share card group", () => {
     const resourceCardLabel = screen.getByText("Video");
     expect(resourceCardLabel).toBeInTheDocument();
   });
-  it("should toggle the checkbox when clicked", async () => {
```
Taha-Hassan-Git, line 58: Any idea why this test has been removed?

package.json
```diff
@@ -90,6 +90,7 @@
     "@hubspot/api-client": "^11.2.0",
     "@mdx-js/loader": "^3.1.0",
     "@mux/mux-node": "^9.0.1",
+    "@mux/mux-player": "^3.10.2",
```
Taha-Hassan-Git, line 93: Why does this file have a bunch of extra packages?

Taha-Hassan-Git (commented):
Tooling isn't my forte, but it broadly looks ok to me. Just a couple of things I thought were odd:

package.json
Taha-Hassan-Git: Oh, just found the answer to this in your PR description!

src/components/TeacherComponents/LessonShareCardGroup/LessonShareCardGroup.test.tsx
orangemug: Yeah so the commit is https://github.com/oaknational/Oak-Web-Application/commit/8c168a7b7380c43ccfa8dd9e86b5a02c4521bb2c it seems to have been incorrectly passing before... not entirely sure why, but that's what me (well Tash) managed to work out

package.json
orangemug: Yeah, that surprised me also 😄 

`pnpm` to the rescue!

jack-skerrett-oak (approved):
QA hasn't had a huge opportunity to look but all major journeys seem fine. What's the worst that can happen.

.github/workflows/code_checks.yml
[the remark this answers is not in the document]
CatMillerCM: I'm not sure, but I think we may still need `actions/setup-node`. According to the [pnpm/action-setup docs](pnpm/action-setup), there's a note at the end that states "This action does not setup Node.js for you, use [actions/setup-node](https://github.com/actions/setup-node) yourself."
I think it'll still pass on branch because apparently "GitHub runners already have Node on them", but if we don't explicitly keep this setup then we aren't controlling node and locally might be flakey?

However, I've now found [this thread](https://github.com/orgs/pnpm/discussions/9044?utm_source=chatgpt.com#discussioncomment-16815734) and one of the maintainers has said "The standalone scripts are the best way to install pnpm because in that case you can install it on a system that doesn't have node.js installed and then you can use pnpm to install node." so now I'm confused 🤷‍♀️

rightsaidjames (approved):
Checked the curriculum and about us journeys - all working as before.

.github/workflows/code_checks.yml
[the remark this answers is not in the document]
orangemug: It looks like we do the same elsewhere (https://github.com/oaknational/oak-ai-lesson-assistant/blob/4d42584e0c9b6f503ed1209689db01129ae3af02/.github/workflows/lint.yml#L22) nice catch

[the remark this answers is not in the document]
orangemug: Hopefully fixed in https://github.com/oaknational/Oak-Web-Application/pull/4158/commits/abdeec5a1598cb56426916e07648f003c906a760

CatMillerCM (approved):
Pulled down locally and pnpm scripts and local dev seem to be working. Not my forte here but all looks good to me!

.github/workflows/code_checks.yml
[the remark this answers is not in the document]
orangemug: Fixed

PR #4171, opened 2026-04-21
PUPIL-1719 fix: added bugsnag init inside classroom error reporter

charsville:
## Description
It seems like Google Classroom errors thrown from the app/api layer were not being sent to Bugsnag. 

Looking through Vercel logs it appears that the Bugsnag.start has not been called initialising Bugsnag before we trigger errors.

Music year: 1951

- Added initialisation of Bugsnag when GC error reporter is created. 

## Issue(s)
[Notion Ticket](https://www.notion.so/oaknationalacademy/Bugsnag-not-receiving-error-reports-34926cc4e1b180d19944c76bee293356?source=copy_link)

## How to test

1. Trigger an error in Google Classroom
    1. In a browser window, goto: https://oak-web-application-website-git-fix-pupil-1719-gc-bugsnag-init.vercel.thenational.academy/api/classroom/auth/callback
    2. This should trigger an error - in the browser you should see `{"error":"Authentication failed"...`
2. You should then see in Bugsnag Development env has a GC error (may not happen due to sampling)
3. You should also see in Vercel logs that there is no error line after the Google Classroom error that says:
 `Bugsnag.notify() was called before Bugsnag.start()`

## Checklist

- [ ] Design sign-off


benprotheroe (approved):

paulsweeting (approved):
Looks good, error was raised in Bugsnag. 

Image

I also tested it with a user denying access, which triggers a 'Warning', and this was logged but not raised in Bugsnag, which is good as we don't want to be spammed each time someone denies access.

PR #4198, opened 2026-04-30
Feat/lesq 1951/unit download success [LESQ-1951]

njblore:
## Description

Music year: 1966

- rename `UnitView` -> `UnitOverviewContent`
- create new component `UnitView`, this needed to be moved out of the page component so the download button state can be shared between the header and content
- show oak toast when unit download starts
- add an inline banner which renders when unit download starts

## How to test

1. Go to https://oak-web-application-website-git-feat-lesq-1951unit-downl-e090f0.vercel.thenational.academy/programmes/music-secondary-ks3/units/atmospheric-music/lessons
2. Click on download unit
3. You should see the banner appear when you click
4. You should see a success toast when the download starts

## Screenshots

Screenshot 2026-04-30 at 13 06 52

Screenshot 2026-04-30 at 13 11 22


RBrownsett:
@njblore I know we discussed it during pointing, but thinking maybe we should have the toast auto dismiss after a period of time e.g. 10 seconds? Just makes it easier for keyboard users rather than trying to get the toast in focus. Plus the inline banner also tells the user the download could take a few minutes.

njblore:
@RBrownsett yeah I can try that, it does feel like a bit much with both appearing and needing attention

RBrownsett (approved):
Approved, think it works better with the auto dismiss. Banner and toast show as expected

messmike (approved):

Taha-Hassan-Git (approved):
🚀

PR #4203, opened 2026-05-01
Fix/remove legacy curriculum downloads

CatMillerCM:
## Description
Removes the legacy curriculum downloads page and all remaining access points to it now that legacy resources no longer need to be available from OWA.

## List of changes

- removed /teachers/curriculum/previous-downloads page and associated tests
- removed the Previously released plans link from SubjectPhasePicker
- removed from deployment urls and urls
- removed redundant types
- moved shared DownloadCategory typing/constants into a better-named download categories fixture file and removed previousDownloads fixture
- removed `HeaderListingCurriculumDownloadButton` as no longer needed
- removed `hasCurriculumDownload` bool from `HeaderListing` component and use cases, as it was only needed to render previous-downloads link (in specialist) or now redundant `HeaderListingCurriculumDownloadButton`
- removed legacy/specialist unit-list links to the deleted page
- now passing `null` to curriculumHref of existing units - potentially they could be removed if we don't want this?

There are still a few "legacy" bools and funcs knocking around, but I wasn't sure it was a good idea to go poking around those areas and instead scoped just to the curriculum page. But let me know if a wider OWA clean-up is needed!

## Issue(s)

Fixes ADOPT-1965

## How to test

2. Click on \_\_\_\_\_\_

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist

- [ ] Design sign-off


src/components/TeacherComponents/UnitList/UnitList.tsx
CatMillerCM, line 1: @rightsaidjames I have passed `null` to the curriculumHref now as wasn't sure if we straight up wanted to remove these units, or only the association to the previous-downloads page? Din't want to drill down too far 😬

src/common-lib/urls/urls.ts
```diff
@@ -1108,12 +1100,6 @@ export const OAK_PAGES: {
     configType: "internal",
     pageType: "curriculum-downloads",
   }),
-  "curriculum-previous-downloads": createOakPageConfig({
-    pathPattern: "/teachers/curriculum/previous-downloads",
-    analyticsPageName: "Curriculum Previous Downloads",
-    configType: "internal",
-    pageType: "curriculum-previous-downloads",
-  }),
   onboarding: createOakPageConfig({
```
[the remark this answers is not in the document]
CatMillerCM: Redirects PR here: https://github.com/oaknational/Leaf/pull/145

src/components/TeacherComponents/UnitList/UnitList.tsx
rightsaidjames: I don't know too much about how the legacy curric downloads logic works on Teacher listing pages, but passing null here rather than digging too much into the logic seems like a sensible approach. I also imagine it's mostly moot anyway, as we're not displaying legacy units anywhere on the site except for the [redesigned EYFS pages](https://thenational.academy/teachers/eyfs/maths) (which don't link to legacy curric downloads).

Perhaps @njblore or @Taha-Hassan-Git can advise here?

charsville (approved):
LGTM code-wise

src/components/TeacherComponents/UnitList/UnitList.tsx
njblore: Yep that looks good to me! The link in the swimming section looks like it should have been `null` to begin with 🙈 and the legacy sections aren't being rendered now so it shouldn't have any impact there (we will be removing and cleaning these up after the integrated journey is launched... 🔜 )

njblore (approved):
All looks sensible to me 🚀

rightsaidjames (approved):
LGTM!

PR #4214, opened 2026-05-08
Feat/lesq 1946/success toast

couplm:
## Description

Music year: [1985](https://www.youtube.com/watch?v=K38xNqZvBJI&list=RDK38xNqZvBJI)

- Update oak components version
- Fix some sonar issues
- Show toast on download (success and error)

## How to test

1. Go to [Downloads page](https://oak-web-application-website-d9jjk5dnz.vercel.thenational.academy/programmes/science-secondary-ks3/units/forces/lessons/what-forces-do/downloads?preselected=all)
Requires feature flag
2. Click on the "Download .zip" button
3. You should see the success toast in the top right if successful
4. You should see the error toast in the top right if unsuccessful

## Checklist

I spent what felt like an entire year looking at tests but then i put this draft in and the code coverage was fine 🤡 
Looks like there aren't tests for the download flow in this area...? They kicked my butt 😭 
- [ ] Design sign-off


src/components/TeacherViews/LessonDownloads.view.tsx
```diff
@@ -359,7 +361,8 @@ export function LessonDownloads(props: LessonDownloadsProps) {
               unitSlug: unitSlug!,
             })}
           />
-        ) : isDownloadSuccessful ? (
+        )}
+        {!showGeoBlocked && isDownloadSuccessful ? (
```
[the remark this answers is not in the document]
couplm: @copilot which is better, the ternary operation or separately declaring each condition? i.e. showGeoBlocked, then !showGeoBlocked && isDownloadSuccessful, then !showGeoBlocked && !isDownloadSuccessful?

I am trying to address sonar issue typescript:S3358

messmike (approved):

RBrownsett:
Nice @couplm - works with no issues in chrome. In Safari the message is dismissed straight away and the dismiss button is missing - not sure if there is an easy fix? If not I can raise a bug for the future. Example:

https://github.com/user-attachments/assets/74bc4e38-8a15-41ca-8ebb-082e4a71bf67

couplm:
> Nice @couplm - works with no issues in chrome. In Safari the message is dismissed straight away and the dismiss button is missing - not sure if there is an easy fix? If not I can raise a bug for the future. Example:
> Screen.Recording.2026-05-08.at.14.20.45.mov

Thank you Richard, I will try and resolve that on Monday. I think that navigating away from the page forces the message to disappear. I will try and find a way to make it stay and check out the close button issue. I will get back to you when this is ready 🤠

src/components/TeacherViews/LessonDownloads.view.tsx
```diff
@@ -286,9 +296,13 @@ export function LessonDownloads(props: LessonDownloadsProps) {
     } catch {
       setIsAttemptingDownload(false);
       setIsDownloadSuccessful(false);
-      setApiError(
```
njblore, line 288: can we remove this `apiError` state now or is it used for something else? If we're moving to a toast for the download error I think we should bring the curriculum download page in line at some point as well, and we can remove that error field from the download form entirely

src/app/(core)/programmes/[subjectPhaseSlug]/units/[unitSlug]/lessons/[lessonSlug]/downloads/success/Components/DownloadSuccessView.tsx
```diff
@@ -76,6 +80,17 @@ export function DownloadSuccessView({
...
+  useEffect(() => {
+    setCurrentToastProps({
+      message: "Download started. This may take a few minutes",
+      variant: "success",
+      autoDismiss: true,
+      showClose: true,
+      showIcon: true,
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
Comment (couplm) [inline src/app/(core)/programmes/[subjectPhaseSlug]/units/[unitSlug]/lessons/[lessonSlug]/downloads/success/Components/DownloadSuccessView.tsx:89]: Didn't couldn't see a better way from researching, but happy for suggestions. It felt better to put the success message on the thanks for downloading page since you can't await for a URL replace. Having it on the lesson downloads view was causing it to disappear too quickly when Richard was testing

src/components/TeacherViews/LessonDownloads.view.tsx
couplm: This is a good point, I am not entirely certain so added it back for safety for now. It was just a quick decision made on Figma to show an error toast

couplm:
Also was just wondering, what is (/is there) a correct way to mock downloads and hubspot? How might we test things related to this flow?
I spent an embarrassing amount of time looking at adding some code coverage in some areas and testing that the download banners were showing correctly etc but couldn't get this to work. I sacked it off in favour of doing some other stuff on this PR and code coverage turned out to be fine, but just realised I never got to the bottom of it 😅

njblore:
@couplm generally it will be checking which functions are being called in the process and adding a `jest.mock()` with a response if needed for each function ([example mocking fetchHubspotContactDetails with response](https://github.com/oaknational/Oak-Web-Application/blob/main/src/app/(core)/programmes/%5BsubjectPhaseSlug%5D/%5Btab%5D/Components/ProgrammeDownloads/ProgrammeDownloads.test.tsx#L75)). But it's a bit tricker in this specific case because we navigate the page on success and we can't do that in a test 🫠  
What we could do is mock `router.replace` and check that it was called with the expected url after the mocked download succeeded ([example](https://github.com/oaknational/Oak-Web-Application/blob/main/src/app/(core)/programmes/%5BsubjectPhaseSlug%5D/%5Btab%5D/Components/ProgrammeOverview/ProgrammeOverview.test.tsx#L12)), but I'd say the value to effort ratio for all that mocking is pretty high!

njblore (approved):
looks good! we might want a follow up ticket to unify the download error messaging across lessons, curriculum and units - not sure whether having both a form error and a toast is overkill or we should pick one for all journeys?

messmike:
> looks good! we might want a follow up ticket to unify the download error messaging across lessons, curriculum and units - not sure whether having both a form error and a toast is overkill or we should pick one for all journeys?

I agree that we should be consistent for the lesson, unit and curriculum download errors. We want to keep the form error on the page in line with the form so it doesn't disappear until a user resubmits the form. The toast is more for letting a user know they have initiated something and for providing them with feedback in the moment. Hope that makes sense! @RBrownsett fyi - not sure if we need a ticket for this as Tash mentioned for post-launch?

RBrownsett (approved):
Nice, now works the same on safari. Yep @messmike I will create a ticket in the tidy up epic to have consistent message/action when there is an issue with the download

PR #4235, opened 2026-05-19
fix: reject requests to the curriculum page with too many segments

carlmw:
## Description

Music year: 1982

the catch all route meant that a request with too many segments would still be served with a 200 which would be cached through ISR creating a lot of entries in the cache.

E.g. `/teachers/curriculum/science-secondary-edexcel/units/iamevil/HTTP/1.0%25252525252525250D%25252525252525250Ax:/`

would be served a 200 with only a cosmetic 404 in the unit modal

## How to test

1. Go to https://oak-web-application-website-git-fix-isrpocalypse.vercel.thenational.academy/teachers/curriculum/computing-secondary-aqa/units/iamevil/this/is/an/injection/attack
2. You should see a 404
3. Go to https://oak-web-application-website-git-fix-isrpocalypse.vercel.thenational.academy/teachers/curriculum/computing-secondary-aqa/units
4. All proper links should work as before

## Checklist

- [ ] Design sign-off

### DIFF
--- a/src/pages/teachers/curriculum/[subjectPhaseSlug]/[...slugs].tsx
+++ b/src/pages/teachers/curriculum/[subjectPhaseSlug]/[...slugs].tsx
@@ -235,6 +235,18 @@ export type URLParams = {
   subjectPhaseSlug: string;
 };
 
+/** Rejects catch all paths with too many segments */
+function hasAllowedCurriculumPathShape(
+  tab: string,
+  pathSegmentsAfterTab: string[],
+): boolean {
+  if (tab === "units") {
+    return pathSegmentsAfterTab.length <= 1;
+  }
+
+  return pathSegmentsAfterTab.length === 0;
+}
+
 export const getStaticPaths = async () => {
   if (shouldSkipInitialBuild) {
     return getFallbackBlockingConfig();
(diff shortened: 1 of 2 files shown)


orangemug (approved):
Looks good to me! Also thank you for sorting this, great sleuthing

rightsaidjames (approved):

src/pages/teachers/curriculum/[subjectPhaseSlug]/[...slugs].tsx
```diff
@@ -258,13 +270,18 @@ export const getStaticProps: GetStaticProps<
...
       }
-      const [tab] = context.params.slugs;
+      const [tab, ...restSegments] = context.params.slugs;
       const isPreviewMode = context.preview === true;
       if (!tab || !VALID_TABS.includes(tab as CurriculumTab)) {
         throw new OakError({
           code: "curriculum-api/not-found",
         });
       }
+      if (!hasAllowedCurriculumPathShape(tab, restSegments)) {
```
Comment (mantagen) [inline src/pages/teachers/curriculum/[subjectPhaseSlug]/[...slugs].tsx:280]: Wondered if we'd be better off using the data fetch result for conditional 404ing rather than making assumptions about the shape in the client

Comment (carlmw) [reply src/pages/teachers/curriculum/[subjectPhaseSlug]/[...slugs].tsx:280]: Yeah we already do that for the `subjectPhaseSlug`, it's only the unit slug that doesn't need to be valid currently. Leaving it to after the data is fetched wouldn't help us ultimately - we'd still need to 404 if we received anything but the first extra segment for the unit slug.

In any case we could still be spammed with valid subject phase and unit slugs (which we are) plus any arbitrary garbage (which is what is happening).

PR #4246, opened 2026-05-22
Feat/lesq 1999/topnav exam board

couplm:
## Description

Music year: [1990](https://www.youtube.com/watch?v=etviGf1uWlg)

- Added exam board panel to top nav dropdown and hamburger menu
- User can tab between sections
- Selector only shows for KS4
- Subject is highlighted with the selected state
- Tabbing out closes the exam board panel
- Exam boards are sorted by board name (then foundation and higher)

## How to test

1. Go to [OWA Preview URL from Vercel bot comment](https://oak-web-application-website-git-feat-lesq-1999topnav-exam-board.vercel.thenational.academy/)
2. Click on the dropdown/hamburger
3. Click on KS4
4. Click on a subject that has exam boards, e.g. Physics
5. You should see the exam board panel appear

## Screenshots

image

How it should now look:
image
image

## Checklist

- [ ] Design sign-off


src/app/(core)/teachers/programmes/[slug]/[tab]/Components/ProgrammeDownloads/ProgrammeDownloads.tsx
Comment (couplm) [inline src/app/(core)/teachers/programmes/[slug]/[tab]/Components/ProgrammeDownloads/ProgrammeDownloads.tsx:1]: I just made these changes due to something breaking from oak-components, not related to my work

src/components/AppComponents/TopNav/TeachersTopNavHamburger/HamburgerSubMenu.tsx
```diff
@@ -133,6 +151,42 @@ export function SubmenuContent(
       );
       if (!keystage) return null;
       const subjects = keystage.children;
+
+      if (selectedExamBoardSubject?.examBoards?.length) {
+        return (
+          <SubmenuContainer
+            title={
+              `${keystage.title}, ${selectedExamBoardSubject.title}` as SubmenuState
+            }
```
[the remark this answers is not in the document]
couplm: Valid, but unsure about this as it has knock on effects by changing the type to string and potentially makes other areas less type safe/jankier if they are made strings

couplm:
@copilot please make a list of what to test, including accessibility paths, for manual QA

src/components/AppComponents/TopNav/TopNavDropdown/TopNavSubjectButtons.tsx
njblore: I think we want to make the subject buttons that open the examboard menu actual `button`s with an undefined href rather than links for a11y

src/components/AppComponents/TopNav/ExamBoardPanel/ExamBoardPanel.tsx
```diff
@@ -0,0 +1,308 @@
...
+              if (selectedSubject.slug === "maths")
+                title =
+                  examBoard.title.charAt(0).toUpperCase() +
+                  examBoard.title.slice(1).toLowerCase();
+              const key = examBoard.tierSlug
+                ? `exam-board-${examBoard.slug}-${examBoard.tierSlug}`
+                : `exam-board-${examBoard.slug}`;
+
+              return (
+                <OakRadioAsButton
```
njblore: I think these should just be `Link` elements with a href, I'm not sure the radio group makes sense here semantically - it's really just a list of links. I'd suggest using oakSecondaryButton as a link though it doesn't match the design exactly, maybe oakPrimaryInvertedButton if it's too harsh? (I imagine the design using the radio button is what lead to using the radio group?)

src/node-lib/curriculum-api-2023/queries/topNav/getTeachersNavData.ts
```diff
@@ -2,6 +2,100 @@ import { TopNavResponse, TeachersBrowse } from "./topNav.schema";
...
+    return (
+      programme.programme_fields.dataset !== "legacy" &&
+      !isSlugLegacy(programme.programme_slug)
+    );
+  });
+
+  const examBoards = programmesForKs
+    .map((p) => {
+      const examBoard = parseExamBoardFromProgrammeSlug(
+        p.programme_fields.examboard_slug ?? p.programme_slug,
```
njblore: I'm a bit unclear on why we need to parse the examboard slug from the programme slug? it should always be in the programme fields, is there an edge case we're capturing?

src/components/AppComponents/TopNav/TopNavDropdown/TopNavSubjectButtons.tsx
```diff
@@ -12,53 +12,126 @@ import { getValidSubjectIconName } from "@/utils/getValidSubjectIconName";
...
+        })
+      : resolveOakHref({
+          page: "teacher-programme",
+          subjectPhaseSlug:
+            programmeSlug?.replace(/-ks\d+(?=-|$)/, "") ?? // remove keystage but keep other options
+            `${subjectSlug}-${phaseSlug}`,
+          tab: "units",
+        });
+
+  return href + (queryParams ? `?${queryParams}` : "");
```
njblore: we should update the url generator to include a `query` param rather than amending the string here, `teacher-programme` already has it set up for keystages and years, I think that's all we need for the subject buttons?

src/components/AppComponents/TopNav/ExamBoardPanel/ExamBoardPanel.tsx
```diff
@@ -0,0 +1,308 @@
...
+          }}
+          $gap="spacing-12"
+        >
+          {examBoards
+            .toSorted((a, b) => a.title.localeCompare(b.title))
+            .map((examBoard) => {
+              let title = examBoard.tierSlug
+                ? `${examBoard.title} (${examBoard.tierSlug.charAt(0).toUpperCase() + examBoard.tierSlug.slice(1)})`
+                : examBoard.title;
+              if (selectedSubject.slug === "maths")
```
njblore: I don't love this subject specific logic, is there a way to handle this more generically for subjects with tiers and no examboards?

src/components/AppComponents/TopNav/ExamBoardPanel/ExamBoardPanel.tsx
```diff
@@ -0,0 +1,308 @@
...
+                title =
+                  examBoard.title.charAt(0).toUpperCase() +
+                  examBoard.title.slice(1).toLowerCase();
```
njblore: is this logic that could be done when constructing `examBoards`? (sidenote `examboards` as a name here is a bit misleading as it's tiers and/or examboards, confused me initially!)

src/components/AppComponents/TopNav/ExamBoardPanel/ExamBoardPanel.tsx
```diff
@@ -0,0 +1,308 @@
...
+    ).toString();
+
+  if (!examBoards || examBoards.length === 0) {
+    return null;
+  }
+
+  const navigateToSubject = ({ examBoard }: { examBoard: string }) => {
+    if (examBoard === null) return;
+
+    let board = examBoard.split("(")[0]?.toLowerCase().trim(); // in case value is in format "examBoard-tier"
```
njblore: this might be handled with a refactor to use proper links with hrefs but we should keep examboard anf tier as separate props rather than sticking them together and then splitting them again here to reduce number of places it can go wrong

src/node-lib/curriculum-api-2023/queries/topNav/getTeachersNavData.ts
couplm: Yeah this ticket has been really annoying with edge cases. This one is that most of the time it takes the exam board, but maths has just higher and foundation with a null exam board. There may be a better way but I am not super familiar with the data. I'll take another look

src/components/AppComponents/TopNav/ExamBoardPanel/ExamBoardPanel.tsx
couplm: I agree completely with the radio group thing. The decision was made based on a design decision and the button style not existing anywhere else, which is why we chose this instead of the buttons you suggested. However it doesn't really make sense like you say.

I'll try with some links and see how close I can get it.

src/components/AppComponents/TopNav/ExamBoardPanel/ExamBoardPanel.tsx
couplm: Me neither 🥴 I'll try and figure out a better way to detect it. If I find a good way to find an alternate to exam board as discussed below it could help here too.

src/components/AppComponents/TopNav/ExamBoardPanel/ExamBoardPanel.tsx
couplm: Sure, it is because it did start out as exam boards and I forgot to change it when it gave birth to multiple yaks 🥴 Will clarify

src/node-lib/curriculum-api-2023/queries/topNav/getTeachersNavData.ts
njblore: Ah yeah I see, in that case i'd just do a check whether tier or examboard exists here, and pass the relevant data with examboard as null when only tier exists

src/node-lib/curriculum-api-2023/queries/topNav/getTeachersNavData.ts
```diff
@@ -2,6 +2,100 @@ import { TopNavResponse, TeachersBrowse } from "./topNav.schema";
...
+  // Handles cases where exam board is only in programme slug, e.g. maths higher and foundation levels
+  if (examboardSlug.includes("-")) {
+    const parts = examboardSlug.split("-");
+    const slug = parts.at(-1);
+    if (!slug) return null;
+    const title = slug.toUpperCase();
+    return { slug, title };
+  }
+  // Convert slug to title (e.g., "aqa" => "AQA", "ocr" => "OCR")
+  const title = examboardSlug.toUpperCase();
```
njblore: I notice some of the casing is a little off, eg `EDEXCEL` instead of `Edexcel` - we should be able to get proper titles from the data rather than converting slugs, eg. `tier_description` should get you `Foundation` and for examboard it will get `AQA` or `Edexcel` as appropriate. These should be in the `programme_fields` object in the data

src/components/AppComponents/TopNav/ExamBoardPanel/ExamBoardPanel.tsx
njblore: makes sense! seems to be a a proliferation of yaks lately 😭

src/components/AppComponents/TopNav/TopNavDropdown/TopNavDropdown.tsx
```diff
@@ -54,8 +56,17 @@ const TeachersPhaseSection = ({
       TeachersSubNavData["primary" | "secondary"]["children"][number]["slug"]
     >(defaultKeystage);
 
+  const [selectedSubject, setSelectedSubject] =
+    useState<SubjectsNavItem | null>(null);
+
+  const [examBoardPanelMode, setExamBoardPanelMode] = useState<
```
njblore: with only 2 modes I'd prefer to use a boolean state here for simplicity & consistency, eg `examboardPanelOpen, setExamboardPanelOpen` = true/false

couplm: Good spot thank you, it did used to have 3 states 😅

src/components/AppComponents/TopNav/ExamBoardPanel/ExamBoardPanel.tsx
couplm: Thank you, I didn't know how to resolve this as the radios were preventing me passing in the correct data. Felt constrained by those but once that was addressed it made it a lot easier.
Your suggestions have really helped me to work through things that felt messed up but really difficult to fix before. Breaking them up in the ways you identified allowed me to think about them more clearly and unlocked each problem like a puzzle 🙇‍♀️ I am very grateful 😁

src/components/AppComponents/TopNav/ExamBoardPanel/ExamBoardPanel.tsx
```diff
@@ -0,0 +1,308 @@
...
+      tier,
+    );
+
+    router.push(queryParams ? `${href}?${queryParams}` : href);
+
+    onClick(selectedSubject.slug, "ks4");
+    onClose();
+  };
+
+  const focusNextExamBoard = (
```
njblore, line 102: not for now, but I think we have essentially duplicated a lot of the functionality of `dropdownFocusManager` here so for a future improvement it would be good if the examboard panel is added as another "layer" to the dropdown menu, so examboard options would be added as children of the subject buttons, like the subjects buttons are children of the keystage buttons. That should make the focus flow automatically, without us having to move it around manually here (though we'll likely need to do some work to ensure the focus manager works with an optional third layer)

couplm: Agreed, without adding too many yaks to this one, it'd be good to have it all in one place for future readability and any more dropdowns :)

messmike:
Thanks @couplm, as mentioned in the Figma comment I left, I think 'OakSecondaryButton' with the solid black border looks too visually strong for the key stage & tier buttons. Is there a way to achieve the same visual styling as 'ButtonAsRadio' for these buttons?
Screenshot 2026-05-27 at 08 09 09

src/components/AppComponents/TopNav/TopNavDropdown/TopNavSubjectButtons.tsx
```diff
@@ -12,61 +12,121 @@ import { getValidSubjectIconName } from "@/utils/getValidSubjectIconName";
...
-        keyStageSlug,
-      })
-    : // If there is only one programme, link to the unit listing page for that programme
-      resolveOakHref({
-        page: "unit-index",
-        programmeSlug: programmeSlug!,
-      });
+  if (keyStageSlug === "early-years-foundation-stage") {
+    return resolveOakHref({
+      page: "unit-index",
```
njblore, line 30: can we move the custom eyfs href resolution logic from `unit-index` to `teacher-programme` so we don't have to do the extra checks here?

src/components/AppComponents/TopNav/TopNavDropdown/TopNavSubjectButtons.tsx
```diff
@@ -12,61 +12,121 @@ import { getValidSubjectIconName } from "@/utils/getValidSubjectIconName";
...
+      programmeSlug: programmeSlug!,
+    });
+  }
+
+  if (programmeCount > 1 && keyStageSlug) {
+    return resolveOakHref({
+      page: "programme-index",
```
njblore: the subjects with more than one programme will now open the examboardPanel so I think this is moot

src/components/AppComponents/TopNav/TopNavDropdown/TopNavSubjectButtons.tsx
```diff
@@ -12,61 +12,121 @@ import { getValidSubjectIconName } from "@/utils/getValidSubjectIconName";
...
+      subjectSlug,
+      keyStageSlug,
+    });
+  }
+
+  return resolveOakHref({
+    page: "teacher-programme",
+    subjectPhaseSlug:
+      programmeSlug?.replace(/-ks\d+(?=-|$)/, "") ??
```
njblore: I believe we should have a helper function for this, yes it's `getSubjectPhaseSlug` - generally we want to make sure we're using shared functions for this kind of logic to make sure we're handling it consistently throughout the app

src/components/AppComponents/TopNav/ExamBoardPanel/ExamBoardPanel.tsx
```diff
@@ -0,0 +1,268 @@
...
+    <OakFlex $flexDirection={"column"}>
+      <OakBox $width={"max-content"} $position={"relative"}>
+        <OakHeading
+          $font={"heading-7"}
+          tag="h6"
+          $mt={"spacing-0"}
+          $mb={"spacing-16"}
+        >
+          Choose exam board for KS4{" "}
+          {convertUnitSlugToTitle(selectedSubject.slug, false)}
```
njblore: does selected subject not have a `title` property?

src/components/AppComponents/TopNav/ExamBoardPanel/ExamBoardPanel.tsx
```diff
@@ -0,0 +1,268 @@
...
+              },
+            });
+
+            return (
+              <OakSecondaryButton
+                element="a"
+                href={href}
+                key={key}
+                data-testid={key}
+                onClick={(e) => navigateToSubject(e, href)}
```
njblore: I think we can remove the `navigateToSubject` fn and just call the onClick callback here? the link will handle the navigation itself

couplm:
> Thanks @couplm, as mentioned in the Figma comment I left, I think 'OakSecondaryButton' with the solid black border looks too visually strong for the key stage & tier buttons. Is there a way to achieve the same visual styling as 'ButtonAsRadio' for these buttons?

No problem, I had a play around and I think this is close enough? The buttons don't allow us to change the style very easily, but it's possible to override. Could be one for the Oak Components work :)

Screenshot 2026-05-27 at 08 39 48

messmike:
> > Thanks @couplm, as mentioned in the Figma comment I left, I think 'OakSecondaryButton' with the solid black border looks too visually strong for the key stage & tier buttons. Is there a way to achieve the same visual styling as 'ButtonAsRadio' for these buttons?
> 
> No problem, I had a play around and I think this is close enough? The buttons don't allow us to change the style very easily, but it's possible to override. Could be one for the Oak Components work :)
> 
> Screenshot 2026-05-27 at 08 39 48

Perfect, thank you! On the override point - for sure, yes. It might be good to add to this Notion page as I think this is a really good example: https://www.notion.so/oaknationalacademy/Trying-out-oak-components-standards-35326cc4e1b18041a88adc2fc5cd29a7?source=copy_link#35326cc4e1b180aeafa9ce5634f92502

src/components/AppComponents/TopNav/TopNavDropdown/TopNavSubjectButtons.tsx
```diff
@@ -76,18 +136,30 @@ const TopNavSubjectButtons = ({
             <OakLI key={subject.title}>
               <OakSubjectIconButton
                 variant={"horizontal"}
-                element={Link}
```
njblore, line 79: sorry we do still want the actual links to be links with an href, only the subject buttons that open the examboard panel should be buttons 🙈

messmike (approved):

src/components/AppComponents/TopNav/TopNavDropdown/TopNavSubjectButtons.tsx
```diff
@@ -76,18 +136,30 @@ const TopNavSubjectButtons = ({
...
-                  keyStageSlug,
-                })}
-                onClick={() => handleClick(slug, keyStageSlug)}
-                onKeyDown={(e) =>
-                  buttonId && focusManager?.handleKeyDown(e, buttonId)
+                selected={selectedSubject?.title === subject.title}
+                onClick={(e: React.MouseEvent<HTMLButtonElement>) =>
+                  handleSubjectClick(e, subject)
                 }
+                onMouseLeave={() => onSubjectLeave?.(subject)}
```
njblore: what do we do on mouse leave?

src/node-lib/curriculum-api-2023/queries/topNav/getTeachersNavData.ts
```diff
@@ -1,7 +1,80 @@
...
+      return [
+        {
+          slug: tier_slug,
+          title: tier_description,
+          programmeSlug: p.programme_slug,
+          tierSlug: tier_slug,
+        },
+      ];
+    }
+
```
njblore, line 79: I think the `ExamBoard` type has got away from us a bit here as it's sometimes referring to an examboard with an optional tier, but sometimes just a tier which I think could get confusing down the line - I'd suggest reworking it a bit to clarify, something like this for example:
```
type ProgrammeFactorButton = {
   buttonTitle : string;
   programmeSlug: string;
   programmeFactors: {
     tier?: {slug, title};
     examboard?: {slug, title};
   }
}
```

src/components/AppComponents/TopNav/TopNavDropdown/TopNavSubjectButtons.tsx
couplm: Hero, this is so much easier, thank you 😭

src/components/AppComponents/TopNav/TopNavDropdown/TopNavSubjectButtons.tsx
couplm: Sorry that was me misunderstanding! To clarify, we want a conditional such as if there's exam boards, make it a button, otherwise make it a link?

src/components/AppComponents/TopNav/TopNavDropdown/TopNavSubjectButtons.tsx
couplm: It should set the selectedsubject on hover and then when leaving, it should deselect it. I can't remember why but doing it this way made sense at the time. I'll try and rework this now that the components being used are different :)

src/components/AppComponents/TopNav/TopNavDropdown/TopNavSubjectButtons.tsx
njblore: yes exactly we want the actual element to match the intent semantically, so if it's a straight link to a page it should be a link element, but if it's a button that opens another panel then it's a button

src/components/AppComponents/TopNav/ExamBoardPanel/ExamBoardPanel.tsx
njblore: I made a tech debt ticket for this and added it to the integrated journey epic :)

src/components/AppComponents/TopNav/TopNavDropdown/TopNavSubjectButtons.tsx
couplm: I tried to do this but it ended up in lots of mismatches, so I left it as is. I am definitely open to a better solution but found it difficult to separate out stuff like the queries, so the path kept being incorrect

src/components/TeacherComponents/helpers/getSubjectPhaseSlug.ts
```diff
@@ -9,3 +9,7 @@ export const getSubjectPhaseSlug = ({
 }) => {
   return `${subject}-${phaseSlug}${examBoardSlug ? `-${examBoardSlug}` : ""}`;
 };
+
+export const getSubjectPhaseFromProgrammeSlug = (programmeSlug: string) => {
+  return programmeSlug.replace(/-ks\d+(?=-|$)/, "");
+};
```
couplm, line 14: I didn't want to have the regex and looked at doing this by passing the correct values in, but they didn't appear to exist in the data I was looking at in the nav data. The programme slug otherwise had everything in. It is possible I missed the correct data.

src/components/TeacherComponents/helpers/getSubjectPhaseFromProgrammeSlug.ts
```diff
@@ -0,0 +1,3 @@
+export const getSubjectPhaseFromProgrammeSlug = (programmeSlug: string) => {
+  return programmeSlug.replace(/-ks\d+(?=-|$)/, "");
+};
```
couplm: I didn't want to have the regex and looked at doing this by passing the correct values in, but they didn't appear to exist in the data I was looking at in the nav data. The programme slug otherwise had everything in. It is possible I missed the correct data.

Taha-Hassan-Git: Hey, just taking a peek and it looks like you have the `subject` and `phaseSlug` as arguments in the `getSubjectLinkHref` function, so you can call the `getSubjectPhaseSlug` function (src/components/TeacherComponents/helpers/) with those arguments. I don't think you'll need an exam board, because subjects with exam boards won't be links so they don't need an href.

couplm: Ah sorry I didn't clarify properly! I tried to do it with that but then I had issues with the ones that needed core in the url etc. Is there something similar?

njblore: @couplm core/gcse should be the `pathwaySlug` in programme fields data

src/components/AppComponents/TopNav/TopNavDropdown/TopNavSubjectButtons.tsx
njblore: ok no probs, I'll add it to tech debt we can clean it up later 🧹

src/components/TeacherComponents/helpers/getSubjectPhaseFromProgrammeSlug.ts
couplm: Thank you, I'll get this sorted

njblore (approved):
🏌️‍♀️

RBrownsett (approved):
Nice work, tested edge cases with no issues. Further testing will be carried out in the IJ branch

PR #4255, opened 2026-05-28
Feat/lesq 2033/topnav seo [LESQ-2033]

njblore:
## Description

Music year: {owa_music_year}

- Refactor the topnav drop down menus so that they render all possible links on first page load, hidden with css only
- When the page loads with javascript these links will then be removed from the page entirely until requested, to prevent issues with focus and a11y (the top nav is unusable without javascript enabled)

## How to test

**To test you will need to disable javascript in chrome** - on mac you can do this by pressing `cmd + shift + p` and typing javascript in the dialog -> click disable javascript and refresh the page

1. Go to [OWA Preview URL from Vercel bot comment]
2. You should not see anything out of the ordinary
3. Open the dev tools and search the html for `visually-hidden-`
4. You should see matches for all the visually hidden containers, these contain all of the links for the topnav ; there should be containers for each phase, each keystage, each examboard subject at ks4, about us and guidance
5. You should be able to expand the html to see all of the links
6. You should see links for every section of the topnav
7. When javascript is enabled, these links are removed with javascript so you won't see them in the html

Additionally I had to make some changes to the html elements & roles as Pa11y caught some issues now that the topnav sections are all in the page on first load, so it could do with a sanity check of the accessible navigation of the top nav in general

## Screenshots

Screenshot 2026-05-28 at 10 08 48


src/components/AppComponents/TopNav/TopNavDropdown/TopNavDropdown.tsx
```diff
@@ -146,107 +188,91 @@ const TeachersPhaseSection = ({
   };
 
   const isKeystageOpen = (slug: string) => selectedKeystage === slug;
-
   return (
     <OakFlex $gap={"spacing-40"}>
-      <OakUL
+      <OakFlex
```
njblore, line 193: pa11y ci was flagging this, element with role `tablist` shouldn't be a `ul` so it was complaining about the children being `li` instead of `tab`

src/components/AppComponents/TopNav/TopNav.tsx
```diff
@@ -32,6 +33,37 @@ export type TopNavProps = {
...
+  }, []);
+
+  if (shouldDisplay) {
+    return children;
+  } else {
+    return afterInitialRender ? null : (
+      <VisuallyHidden id={`visually-hidden-${hiddenElementId}`}>
+        {children}
+      </VisuallyHidden>
+    );
```
carlmw, line 63: Could this be done with a combination of `VisuallyHidden` and the `inert` attribute instead of removing it from the dom entirely?

carlmw (approved):
Looks good to me, just a thought about using `inert` instead of removing from the DOM after hydration.

src/components/AppComponents/TopNav/TopNav.tsx
njblore: Yeah I initially tried doing it with `inert` but could not get it to work as I expected (could be user error), I didn't want to spend too long digging into it so went with this instead as it just worked - could be something to look into when we have a bit more time!

RBrownsett (approved):
Approved - visually hidden containers displayed as expected with links. Tested top nav with keyboard and voiceOver no regressions

PR #4263, opened 2026-06-01
Cache Curriculum+Sanity data plus enable ISR for the IJ LESQ 2060

carlmw:
## Description

Music year: 1961

Ensures that IJ pages are fully cached where possible and/or API requests are fully cached 

- Programme pages cannot be ISRd as they use `searchParams`, instead we cache every API call with Vercel so that we don't hit Sanity or the Curriculum API on every request
- Cache is for 2 hours to match the ISR cache value
- Other pages with dynamic segments that weren't previously caching for ISR have `export const dynamic = "force-static";` set which will force them to be cached

## How to test

1. Go to https://oak-web-application-website-git-fix-lesq-2060cache-curri-6172e2.vercel.thenational.academy/teachers/programmes/citizenship-secondary-gcse/units
2. Soft reload
3. The page should load very fast
4. Click on a unit
5. Open dev tools
6. Soft reload
7. You should see a `HIT` for the `x-vercel-cache` header which indicates the page was cached
8. Same for lesson and EYFS pages

## Screenshots

What you're looking for in the Vercel logs if you want to delve in. https://vercel.com/oak-national-academy/oak-web-application-website/logs?search=environment%3Apreview+host%3Aoak-web-application-website-git-fix-lesq-2060cache-curri-6172e2.vercel.thenational.academy

** Fully cached page log **

Screenshot 2026-06-01 at 17 34 18

** Cached API calls **

Screenshot 2026-06-01 at 17 35 59

## Checklist

- [ ] Design sign-off

### DIFF
--- a/src/app/(core)/layout.tsx
+++ b/src/app/(core)/layout.tsx
@@ -4,10 +4,13 @@ import LayoutSiteFooter from "@/components/AppComponents/LayoutSiteFooter";
 import TopNav from "@/components/AppComponents/TopNav/TopNav";
 import OakError from "@/errors/OakError";
 import curriculumApi2023 from "@/node-lib/curriculum-api-2023";
+import { cacheData } from "@/node-lib/cache";
 
 // TD: [integrated journey] get revalidate from env somehow
 // revalidate in layout controls revalidation of child pages in route
 export const revalidate = 7200;
+
+const getTopNav = cacheData(() => curriculumApi2023.topNav(), ["top-nav"]);
 
 export default async function CoreLayout({
   children,
@@ -15,7 +18,7 @@ export default async function CoreLayout({
   children: React.ReactNode;
 }>) {
   try {
-    const topNavProps = await curriculumApi2023.topNav();
+    const topNavProps = await getTopNav();
 
     return (
       <>
--- a/src/app/(core)/teachers/eyfs/[subjectSlug]/page.tsx
+++ b/src/app/(core)/teachers/eyfs/[subjectSlug]/page.tsx
@@ -1,9 +1,9 @@
 import { Metadata } from "next";
 
 import { EyfsUnitSection } from "./components/EyfsUnits/EyfsUnits";
+import { getCachedEyfsPageData } from "./getCachedEyfsPageData";
 
 import withPageErrorHandling from "@/hocs/withPageErrorHandling";
-import curriculumApi2023 from "@/node-lib/curriculum-api-2023";
 
 export const metadata: Metadata = {
   title: "Early years foundation stage lesson resources",
@@ -24,7 +24,7 @@ const InnerEyfsPage = async ({
 }>) => {
   const { subjectSlug } = await params;
 
-  const eyfsPageData = await curriculumApi2023.eyfsPage({ subjectSlug });
+  const eyfsPageData = await getCachedEyfsPageData(subjectSlug);
   const units = Object.values(eyfsPageData.units);
 
   return <EyfsUnitSection units={units} />;
--- a/src/app/(core)/teachers/eyfs/[subjectSlug]/page.tsx
+++ b/src/app/(core)/teachers/eyfs/[subjectSlug]/page.tsx
@@ -17,6 +17,8 @@ export const metadata: Metadata = {
   },
 };
 
+export const dynamic = "force-static";
+
 const InnerEyfsPage = async ({
   params,
 }: Readonly<{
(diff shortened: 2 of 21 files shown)


src/node-lib/cache/index.ts
```diff
@@ -0,0 +1,21 @@
...
+export const DEFAULT_PAGE_DATA_REVALIDATE_SECONDS = 7200;
+
+type CacheDataOptions = NonNullable<Parameters<typeof unstable_cache>[2]>;
+
+/**
+ * `unstable_cache` with a default `revalidate`
+ *
+ * Note: Once we upgrade to Next 16, we can switch to a `use cache` directive
+ */
+export function cacheData<Args extends unknown[], Result>(
```
carlmw, line 12: Just a wrapper around `unstable_cache` which sets a 2 hour revalidate, the default is to cache for 1 year which is probably a bit too high for us without invalidation setup.

src/app/(core)/teachers/eyfs/[subjectSlug]/page.tsx
```diff
@@ -17,14 +17,16 @@ export const metadata: Metadata = {
   },
 };
 
+export const dynamic = "force-static";
```
Comment (carlmw) [inline src/app/(core)/teachers/eyfs/[subjectSlug]/page.tsx:20]: The magic bullet to force a dynamic page to be cached for ISR.

The `revalidate` period will be inherited from the layout which is currently set to 2 hours.

src/app/(core)/layout.tsx
```diff
@@ -4,18 +4,21 @@ import LayoutSiteFooter from "@/components/AppComponents/LayoutSiteFooter";
 import TopNav from "@/components/AppComponents/TopNav/TopNav";
 import OakError from "@/errors/OakError";
 import curriculumApi2023 from "@/node-lib/curriculum-api-2023";
+import { cacheData } from "@/node-lib/cache";
 
 // TD: [integrated journey] get revalidate from env somehow
 // revalidate in layout controls revalidation of child pages in route
 export const revalidate = 7200;
 
+const getTopNav = cacheData(() => curriculumApi2023.topNav(), ["top-nav"]);
```
carlmw, line 13: This cache will be shared with all pages, so this will make a big difference.

src/hocs/withPageErrorHandling.tsx
```diff
@@ -56,10 +56,9 @@ function withPageErrorHandling<T extends PageParams>(
         /**
          * Report error to error reporting service
          */
-        const { params, searchParams } = props;
+        const { params } = props;
         await errorReporter(page)(error, {
           ...(await params),
-          ...(await searchParams),
```
carlmw, line 62: Concerned that this would disable ISR

src/app/(core)/teachers/programmes/[slug]/units/[unitSlug]/lessons/getCachedUnitData.ts
```diff
@@ -0,0 +1,16 @@
+import { cache } from "react";
+
+import curriculumApi2023 from "@/node-lib/curriculum-api-2023";
+import { cacheData } from "@/node-lib/cache";
+
+export const getCachedUnitData = cache(
```
Comment (carlmw) [inline src/app/(core)/teachers/programmes/[slug]/units/[unitSlug]/lessons/getCachedUnitData.ts:6]: Cache entry is shared with the download success page.

src/app/(core)/teachers/eyfs/[subjectSlug]/page.tsx
Comment (njblore) [reply src/app/(core)/teachers/eyfs/[subjectSlug]/page.tsx:20]: do we know if/how this differs from having `generateStaticParams`? it sound like it's equivalent to returning an empty array I'm just not clear on why you would use one over the other?

src/app/(core)/layout.tsx
njblore: we could probably increase revalidate time for the top nav as well as it's so stable 👀

njblore (approved):
looks great! 🤙

src/app/(core)/teachers/eyfs/[subjectSlug]/page.tsx
Comment (carlmw) [reply src/app/(core)/teachers/eyfs/[subjectSlug]/page.tsx:20]: My understanding is that `generateStaticParams` is used to provide a set of route params upfront that you want prebuilt at build time. So you might use it if there are some params that are very popular and you want the cache primed.

src/app/(core)/layout.tsx
carlmw: Yeah, we might want to look at tags to make invalidation easier to manage. No reason this couldn't be cached for a very long time and only invalidated when something actually changes.

RBrownsett (approved):
Approved - I can see a hit in the vercel logs for unit and lesson pages

PR #4265, opened 2026-06-02
Feat/lesq 2039/improve lot picker seo

couplm:
## Description

Music year: 1950

- Creates visually hidden links on the subject phase picker
- Copies [Tash's work](https://github.com/oaknational/Oak-Web-Application/pull/4255) 😅 
- Only used in one place so made it less reusable as couldn't think of other use cases

## How to test

2. Disable JavaScript using debugger settings menu/`cmd + shift + p`
3. Open inspect element and search for `visually-hidden`
4. You should see a lot of visually hidden links

## Screenshots

How it should now look:
image

## Checklist

- [ ] Design sign-off


RBrownsett (approved):
Tested - visually hidden links show on the lot picker on the home page and also on the oak curricula page /about-us/oaks-curricula. No regressions found

src/common-lib/urls/urls.ts
```diff
@@ -1363,3 +1364,37 @@ export function resolveOakHref(props: ResolveOakHrefProps): string {
     return "/";
   }
 }
+
+export function generateSubjectPhasePickerLinks(
```
Taha-Hassan-Git: Could we move this into the SubjectPhasePicker component? Or at least out of this file.

Taha-Hassan-Git (commented):
Great work! Seems really weird to generate a bunch of links and then remove them from the DOM, but I guess that's what we'll have to do. Just a small thing with where `generateSubjectPhasePickerLinks` should live.

src/components/SharedComponents/SubjectPhasePicker/SubjectPhasePicker.tsx
```diff
@@ -48,6 +51,57 @@ const TruncatedFlex = styled(OakFlex)`
...
+}) => {
+  const [afterInitialRender, setAfterInitialRender] = useState(false);
+
+  useEffect(() => {
+    setAfterInitialRender(true);
+  }, []);
+
+  if (!afterInitialRender && !shouldHide) {
+    return (
+      <SubjectPhasePickerLinks
```
njblore: can we reuse `MaybeVisuallyHidden` rather than recreate it here?

src/common-lib/urls/urls.ts
couplm: Sure! I initially thought it'd be more re-usable and relevant but it definitely makes more sense to be in the subject phase picker component

src/components/SharedComponents/SubjectPhasePicker/SubjectPhasePicker.tsx
couplm: Makes sense! I had tried similar but couldn't get it to work, looking now I think I had the logic backwards 🥴 I'll update it now :)

src/components/SharedComponents/SubjectPhasePicker/SubjectPhasePicker.test.tsx
```diff
@@ -33,6 +38,57 @@ describe("Component - subject phase picker", () => {
   beforeEach(() => {
     jest.clearAllMocks();
   });
+
+  test("renders subject phase picker SEO links in HTML before hydration", () => {
```
couplm, line 42: Had problems getting the tests to pass as a console error always appeared after the recent changes. Tried looking around but it seems it would keep appearing due to useLayoutEffect coming from the VisuallyHidden import...? So I caught the console error and it will throw a new one if the error is something else.
As always, open to suggestions on how to fix this better!

njblore: 😬 ah that's a shame, might be a bit cleaner to just mock `useLayoutEffect` in this case? An alternative is we could stop using `VisuallyHidden` and replace with our own component which only has the styles  - we were using `VisuallyHidden` as a convenience but if it's becoming inconvenient that might be preferable!

couplm: Was unsure if it's related with how I implemented things. But I amended it and hopefully it's what you meant! Glad there is no hard requirement for `VisuallyHidden`.

src/components/SharedComponents/SubjectPhasePicker/SubjectPhasePicker.tsx
```diff
@@ -141,6 +142,40 @@ const isPhaseSelectionComplete = (
...
+  const resolveTeacherProgrammeHref = (subjectPhaseSlug: string) =>
+    resolveOakHref({
+      page: "teacher-programme",
+      subjectPhaseSlug,
+      tab: "units",
+    });
+
+  return subjects.flatMap((subject) =>
+    subject.phases.flatMap((phase) => {
+      const subjectPhaseSlug = `${subject.slug}-${phase.slug}`;
```
njblore: just a wee nitpick can we make sure we're using the shared constructor functions whenever we're creating subject phase slugs and programme slugs (ie. `getSubjectPhaseSlug` / `createTeacherProgrammeSlug`) rather than using string literals to build on the fly - just saves ourselves a headache down the line if we need to update them for any reason it becomes much harder to track them all down! this could be particularly easy to miss because it's only in the initial page html 😬

njblore (approved):
nice work!

src/components/SharedComponents/SubjectPhasePicker/SubjectPhasePicker.tsx
couplm: Thank you! Don't worry, I keep forgetting we have these as they're tucked away like you say. You reminding me is helpful so I get into the right habits with this repo :)

PR #4288, opened 2026-06-09
Feat/lesq 2025/topnav phase menu

couplm:
## Description

Music year: 1956

- Top nav has phase menu for primary/secondary in the **teachers** view (not just pupils)
- Secondary phase view shows all secondary subjects directly (KS3 + KS4 combined), with a "Key stages" option to drill into KS3/KS4 separately
- Primary phase view shows all primary subjects directly, with a "Key stages" option for KS1/KS2/EYFS
- KS4 subjects with exam boards (e.g. Science, Maths, English) open an exam board/tier panel in both phase and key-stage views
- Science and other parent subjects are condensed into a single entry aggregating exam boards from all child programmes
- KS3-only subjects are filtered out of the combined secondary phase view to avoid duplicates
- Tried to move focus navigation more towards dropdown focus manager instead of being in exam board panel
- Also tried to make dropdown focus manager clearer by separating concerns into the focus tree. Maybe it's more confusing now 🫨 Open to feedback on all of this so that I can learn 😭
- Should be keyboard navigable and follow current flow wrapping from parent to child and back to parent's sibling (is there a word for that?)
- Fixed broken `aria-controls` reference when "Key stages" tab is selected

## How to test

1. Go to [OWA Preview URL from Vercel bot comment]/
2. Click on **Secondary** in the top nav
3. You should see a "Secondary" option (phase view, all subjects) and a "Key stages" option (drill into KS3/KS4)
4. Click **Secondary** — all secondary subjects should be listed together; KS4 subjects (e.g. Science, Maths) should open an exam board/tier panel on click
5. Click **Key stages** — KS3 and KS4 tabs should appear; switching tabs should update the subject grid
6. Click **Primary** — all primary subjects should be listed; "Key stages" should show KS1/KS2/EYFS tabs including EYFS subjects linking to `/teachers/eyfs/<subject>`
7. Verify keyboard navigation: Arrow keys move between items in the left column; Tab/Shift-Tab moves through subjects; Escape closes the dropdown

## Screenshots

How it should now look:

## Checklist

- [ ] Design sign-off


src/browser-lib/avo/Avo.ts
```diff
@@ -1529,6 +1529,7 @@ export const FilterType = {
   'LESSON_COHORT_FILTER': 'Lesson Cohort filter',
   'PATHWAY_FILTER': 'Pathway filter',
   'UNIT_FILTER': 'Unit filter',
+  'PHASE_FILTER': 'Phase filter',
```
couplm, line 1532: Is it fine to add this here or is there something else that needs to be done? Maybe an existing one does the same thing, just unclear from names alone

src/node-lib/curriculum-api-2023/queries/topNav/getTeachersNavData.ts
couplm, line 1: Github definitely shouldn't hide this diff. This is probably the worst part of what I wrote 😭
Disclaimer: copilot helped me with the logic here and I tried to tidy it up after and check it all. I am not happy with how long and confusing it all is, but there were so many cases to think of 😵‍💫

couplm:
@copilot please make a list of paths to be tested by manual QA and considerations for this PR

RBrownsett:
@couplm nice work with the thought going into the a11y considerations. One thing I noticed (and got me thinking) The use of tablist on the new section is semantic only, its not wired to the arrow-key handler (which allows users to move up and down the tab list using the arrow keys) handleKeystageArrowKeys is still only used on teachers-primary-ks1, teachers-primary-ks2 etc. However I'm starting to wonder if we should move away from tab semantics to disclosure semantics on the left hand buttons, what do you think? I've put a suggested code change at the bottom of ticket https://app.notion.com/p/oaknationalacademy/Topnav-Phase-menu-dropdown-35f26cc4e1b18075b45ccd243cd56a43?source=copy_link

RBrownsett:
FYI W3C disclosure navigation pattern reading https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/

RBrownsett:
@couplm I think that works much better. Just one thing, it's still on the inner key stage list (KS1, KS2 etc.). That one also needs the role="tablist" removed for the same reason. Just removing the role attribute and keeping the ref and onKeyDown as they are should be fine. 
image

src/components/AppComponents/TopNav/DropdownFocusManager/focusTree.ts
```diff
@@ -0,0 +1,139 @@
...
+  isTeachersBrowseItem,
+} from "@/node-lib/curriculum-api-2023/queries/topNav/topNav.schema";
+
+export const createFocusId = (
+  area: string,
+  slug: string,
+  parentId?: string,
+): string => {
+  if (parentId === undefined) return `${area}-${slug}`;
+  if (!slug) return `${area}-${parentId}`;
```
Taha-Hassan-Git: If the type of slug doesn't allow it to be undefined then I don't think we need this check here.

src/components/AppComponents/TopNav/ExamBoardPanel/ExamBoardPanel.tsx
```diff
@@ -35,37 +41,22 @@ const ExamBoardButton = styled(OakPrimaryInvertedButton)`
...
-    return () => {
-      focusManager.unregisterChildren(parentId);
-    };
-  }, [focusManager, parentId, examboards]);
+  const subjectId =
+    createFocusId(
+      "teachers",
+      `teachers-${phaseSlug}-${viewType}`,
+      selectedSubject.slug,
+    ) ?? `teachers-${phaseSlug}-${viewType}-${selectedSubject.slug}`;
```
Taha-Hassan-Git: I think createFocusId always returns a string, so we shouldn't need the `??` operator. Even so the point of this function is to centralise the id-creation logic, so best not to escape it like this.

src/components/AppComponents/TopNav/ExamBoardPanel/ExamBoardPanel.tsx
```diff
@@ -213,42 +132,48 @@ const ExamBoardPanel = ({
         $flexDirection={"column"}
         $gap={"spacing-8"}
         $reset
-        id={`topnav-teachers-ks4-examboards-${selectedSubject.slug}`}
-        onKeyDown={handleListKeyDown}
+        id={`topnav-teachers-${viewType}-examboards-${selectedSubject.slug}`}
```
Taha-Hassan-Git: I think we should define the panel id once and use the variable name in both places, instead of re-writing it here.

src/components/AppComponents/TopNav/ExamBoardPanel/ExamBoardPanel.tsx
```diff
@@ -213,42 +132,48 @@ const ExamBoardPanel = ({
...
+            examboard.buttonTitle.toLowerCase() !==
+              examboard.programmeFactors.tier?.slug
+              ? `${examboard.buttonTitle} (${examboard.programmeFactors.tier?.description})`
+              : examboard.buttonTitle;
+          const buttonId =
+            createFocusId(
+              "teachers",
+              subjectId,
+              getExamBoardFocusSlug(examboard),
+            ) ?? `${subjectId}-${getExamBoardFocusSlug(examboard)}`;
```
Taha-Hassan-Git: I think this function always returns a string.

src/components/AppComponents/TopNav/DropdownFocusManager/focusTree.ts
```diff
@@ -0,0 +1,139 @@
...
+  groupSlug,
+  subject,
+}: {
+  phaseSlug: "primary" | "secondary";
+  groupSlug: string;
+  subject: TeachersBrowseChildItem["children"][number];
+}): FocusTreeNode => {
+  const subjectId = createFocusId(
+    "teachers",
+    `teachers-${phaseSlug}-${groupSlug}`,
```
Taha-Hassan-Git, line 54: Seems unintuitive to me to build this subjectId with "teachers" in it twice, what's the thinking behind this?

couplm:
> @couplm nice work with the thought going into the a11y considerations. One thing I noticed (and got me thinking)...

> FYI W3C disclosure navigation pattern reading https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/

Thanks for this! I wasn't aware of the difference before, I will try to remember it. And I made the updates suggested on the Notion page 😊 

> @couplm I think that works much better. Just one thing, it's still on the inner key stage list (KS1, KS2 etc.). That one also needs the role="tablist" removed for the same reason. Just removing the role attribute and keeping the ref and onKeyDown as they are should be fine. 

Sorry, I meant to finish this off but got waylaid with meetings. I amended it as you suggested but also made it so we can move between phase and keystage views with the arrow keys. Hopefully it has the correct semantics!

src/browser-lib/avo/Avo.ts
Taha-Hassan-Git: This is a generated file, we shouldn't be making edits to it. If you want to make a change to the analytics pipeline then you need to follow this doc: https://www.notion.so/oaknationalacademy/Avo-5ff21cbd33994c84907c83c6af5a25ee

It's slightly intimidating but actually relatively straightforward. Let me know if you need a hand.

src/components/AppComponents/TopNav/TopNavDropdown/TopNavDropdown.tsx
```diff
@@ -187,7 +202,64 @@ const TeachersPhaseSection = ({
...
+        width={"spacing-160"}
+        selected={isOpen(slug)}
+        onClick={() => onClickFn(slug)}
+        onKeyDown={(e) => focusManager.handleKeyDown(e, buttonId)}
+        id={buttonId}
+        aria-label={
+          title === "EYFS" ? "Early years foundation stage" : undefined
+        }
+      >
+        {title.replace("KS", "Key stage ")}
```
Taha-Hassan-Git: Instead of this, we could pass down the keyStageTitle from the data.

Taha-Hassan-Git (commented):
It works! A mammoth amount of work, so well done. I'm only halfway through reviewing but wanted to send early so you can see my comment about Avo branches. Taking a look at getTeachersNavData now 😫 .

src/components/AppComponents/TopNav/TopNavDropdown/TopNavDropdown.tsx
couplm: Went with a compromise and moved the replace to the consumer. It makes this function cleaner without referencing keystages as it could be phase, but also avoids having to map anything

src/browser-lib/avo/Avo.ts
couplm: Thank you!! Just checking before I do it, it definitely needs adding, right? Was unsure if the others encompass it or not, but it doesn't seem like it from their names

src/components/AppComponents/TopNav/DropdownFocusManager/focusTree.ts
couplm: As in passing teachers as the area and then teachers as part of the slug? I am not sure if I can condense it further without messing up the ID. I may be misunderstanding though :)

RBrownsett:
> > @couplm nice work with the thought going into the a11y considerations. One thing I noticed (and got me thinking)...
> 
> > FYI W3C disclosure navigation pattern reading https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/
> 
> Thanks for this! I wasn't aware of the difference before, I will try to remember it. And I made the updates suggested on the Notion page 😊
> 
> > @couplm I think that works much better. Just one thing, it's still on the inner key stage list (KS1, KS2 etc.). That one also needs the role="tablist" removed for the same reason. Just removing the role attribute and keeping the ref and onKeyDown as they are should be fine.
> 
> Sorry, I meant to finish this off but got waylaid with meetings. I amended it as you suggested but also made it so we can move between phase and keystage views with the arrow keys. Hopefully it has the correct semantics!

Just checked, thats working really well!

src/browser-lib/avo/Avo.ts
Taha-Hassan-Git: Hmmm, looks to me like you need to add it. Sorry!

src/components/AppComponents/TopNav/DropdownFocusManager/focusTree.ts
couplm: I realise now it's confusing, now it should be more explicit :)

RBrownsett:
The exam board panel is a little cramped compared to the designs - Not a huge issue, just wondering if its easy to make the exam panel wider? Though would that mean creating an explicit two column layout and giving exam boards a fixed min width? 
image

src/browser-lib/avo/Avo.ts
couplm: No worries, just better to check than to add and then remove it 😅

couplm:
> The exam board panel is a little cramped compared to the designs - Not a huge issue, just wondering if its easy to make the exam panel wider? Though would that mean creating an explicit two column layout and giving exam boards a fixed min width? 

Good point, I can definitely do this just not sure what the right call to make is here for spacing etc. Could @messmike please advise?

messmike:
> Preview

Do you have a preview link @couplm that I can look at?

couplm:
> > Preview
> 
> Do you have a preview link @couplm that I can look at?

Yep, [this one](https://oak-web-application-website-7qiqfaprz.vercel.thenational.academy/) should be fine. Thank you. The link is down due to waiting for something else 😅

messmike:
Would it be possible to try a minWidth for the exam board panel of `spacing-240`? 
Screenshot 2026-06-11 at 16 04 53

couplm:
Adding a link to the [avo pr](https://www.avo.app/schemas/5PhajbVijwhXVKIJtGMT/branches/KLEaOJfVLl1zGM5M0MJw-/diff) for visibility. Hope it is correct 🥴

RBrownsett:
@couplm the spacing fix looks good! Just noticed that the primary section via the hamburger is blank - will this be fixed when Carls PR get merged (video below)? The new property 'phase filter', is that part of the browse refined event? 

https://github.com/user-attachments/assets/8695e5cb-1965-487f-aec6-45be18752d25

couplm:
> @couplm the spacing fix looks good! Just noticed that the primary section via the hamburger is blank - will this be fixed when Carls PR get merged (video below)? The new property 'phase filter', is that part of the browse refined event?
> Screen.Recording.2026-06-15.at.10.42.11.mov

Thanks Richard, yes it should be resolved by Carl's work. And the phase filter is part of the browse refined event. It will need signing off from another dev just to make sure this is the correct implementation :)

src/components/AppComponents/TopNav/DropdownFocusManager/focusTree.ts
```diff
@@ -0,0 +1,139 @@
...
+  const examboardSlug = examBoard.programmeFactors?.examboard?.slug;
+  const tierSlug = examBoard.programmeFactors?.tier?.slug;
+
+  if (examboardSlug && tierSlug) {
+    return `${examboardSlug}-${tierSlug}`;
+  }
+
+  if (examboardSlug || tierSlug) {
+    return examboardSlug ?? tierSlug ?? "";
+  }
```
carlmw:
```suggestion
  if (examboardSlug) {
    return examboardSlug;
  }
  
  if (tierSlug) {
    return tierSlug;
  }
```

just a nit but this would reduce the complexity and you don't need the nullish fallback.

src/components/AppComponents/TopNav/TopNavDropdown/TopNavDropdown.tsx
```diff
@@ -187,7 +201,75 @@ const TeachersPhaseSection = ({
...
+  };
+
+  const isTopLevelOpen = (slug: string) => selectedTopLevel === slug;
+  const isKeystageOpen = (slug: string) => selectedViewType === slug;
+  const topLevelButtonIds = [
+    createFocusId("teachers", `teachers-${phaseData.slug}`, phaseData.slug),
+    ...(hasKeystageChildren
+      ? [createFocusId("teachers", `teachers-${phaseData.slug}`, "keystages")]
+      : []),
+  ];
```
carlmw, line 269:
```suggestion
  const topLevelButtonIds = [
    createFocusId("teachers", `teachers-${phaseData.slug}`, phaseData.slug),
  ];

  if (hasKeystageChildren) {
    topLevelButtonIds.push(
      createFocusId("teachers", `teachers-${phaseData.slug}`, "keystages"),
    );
  }
```

that spread and ternary could be replaced with an if. More readable imo

src/components/AppComponents/TopNav/TopNavDropdown/TopNavDropdown.tsx
```diff
@@ -370,7 +444,8 @@ const PupilsSection = ({
                   onClick={onClose}
                   onKeyDown={
                     focusManager && buttonId
-                      ? (e) => focusManager.handleKeyDown(e, buttonId)
+                      ? (e: React.KeyboardEvent<HTMLAnchorElement>) =>
+                          focusManager.handleKeyDown(e, buttonId)
```
carlmw, line 456:
```suggestion
                      ? (e) =>
                          focusManager.handleKeyDown(e, buttonId)
```

type annotation here is redundant, it comes from React.

src/node-lib/curriculum-api-2023/queries/topNav/getTeachersNavData.ts
```diff
@@ -11,48 +16,208 @@ import {
...
+
+  return programmes.filter(isNonLegacyProgramme);
+};
+const getSubjectDisplayTitle = (programme: TopNavProgramme) => {
+  const subjectDisplayName =
+    programme.actions?.programme_field_overrides?.subject ??
+    programme.programme_fields.subject;
+  const pathwayTitle = programme.programme_fields.pathway ?? "";
+
+  return subjectDisplayName + (pathwayTitle ? ` (${pathwayTitle})` : "");
```
carlmw, line 39:
```suggestion
  return `${subjectDisplayName}${pathwayTitle && ` (${pathwayTitle})`}`;
```

just a nit, but the mix of concatenation and template strings make this a little harder to reason.

though tbh i don't think the template string is much better. 🏄🏻

src/node-lib/curriculum-api-2023/queries/topNav/getTeachersNavData.ts
```diff
@@ -103,47 +272,225 @@ export const getExamBoardsForKS4Subject = ({
...
+  );
+
+  return [
+    {
+      slug: phaseSlug,
+      title: `${phaseSlug[0]?.toUpperCase()}${phaseSlug.slice(1)}`,
+      description: "",
+      children: orderedChildren,
+    },
+  ] as Omit<TeachersBrowseChildItem, "type">[];
```
carlmw:
```suggestion
  ] satisfies Omit<TeachersBrowseChildItem, "type">[];
```

prefer `satisfies` where not casting. `as` will throw away all type safety.

carlmw (approved):
This looks to have been a slog.

Just some nits from me 🥞

src/components/AppComponents/TopNav/TopNavDropdown/TopNavDropdown.tsx
couplm: I may leave this one as something was kicking off about types not being added, just to avoid the pain of changing it then having to add it back again to get the build to pass 😮‍💨

couplm:
> This looks to have been a slog.
> 
> Just some nits from me 🥞

Thanks so much Carl! DIdn't know code suggestions were a thing 🤯 really useful and helped a lot 😁

RBrownsett (approved):
Great work, a11y considerations are well thought out. Checked in different browsers with no issues.

PR #4305, opened 2026-06-17
CI Upgrade to Node version 24

charsville:
## Description

A lot of our CI/CD was running on older versions of node 20/22. 

Ideally we would move to node 26, as this will become the LTS from October 2026, however some of the actions we use in our CI/CD do not officially support 26 yet, and still target 24. So to be on the safe side this PR bumps to 24, and hopefully we can bump to 26 from October. ([Node releases](https://nodejs.org/en/about/previous-releases))

We also had some issues with Pa11y timing out. One potential cause of this could be due to concurrency overloading the available resources on the runner. This PR reduces the concurrency from 10 to 5. The risk here is that we increase the time it takes for the CI to run, however, it may have the inverse effect because more waves with smaller concurrent runs may make the individual waves complete faster.

- Updates package.json engines
- Updates nvmrc
- Reduce pa11y ci concurrency from 10 to 5
- Adds --no-sandbox flag to pa11y CI as per docs recommendation
- Updates setup-node versions to v6
- Updates checkout to v5
- Updates slack-notify to v2.4.0
- Updates labeler to v6

## How to test
Monitor all CI/CD runs on this PR for regressions or failures

## Checklist

- [ ] Design sign-off


njblore (approved):

rjkward (approved):
thanks for bumping all these, much appreciated 👍

PR #4307, opened 2026-06-17
fix: upgrade posthog-js to 1.388.1 (drop ip=0 from /flags requests)

mantagen:
## Why

We had reports that **uBlock Origin's Lite filter list blocks requests to `/flags` when an `ip=` parameter appears in the query string**, regardless of domain. uBlock have since fixed this upstream, but anyone on the **Lite** version will still hit it.

Older versions of `posthog-js` sent `ip=0` on every request, which trips that filter and breaks flag fetches for affected users.

> Older versions of posthog-js included `ip=0` on every request, which trips this filter. Upgrading to 1.387.0+ removes `ip=0` on every request.

Ref: [Slack thread](https://oaknationalacademy.slack.com/archives/C03SQBG3L78/p1781700733101719)

## What

- Bump `posthog-js` `1.369.0` → `1.388.1` (≥ the 1.387.0 fix that removes `ip=0`).

## ⚠️ CI will fail the release-age check — needs a team decision

`1.388.1` was published 2026-06-16/17, which is **newer than the repo's `minimumReleaseAge` policy window** (`pnpm-workspace.yaml`, 10080 min / 7 days). So `pnpm install` (pre-push hook and CI) will reject it as too new until it ages out (~2026-06-23).

The newest *policy-compliant* version is `1.384.2`, which does **not** contain the fix — so pinning to it would be pointless.

**Options for the team:**
1. Add `posthog-js` (+ `@posthog/core`, `@posthog/types`) to `minimumReleaseAgeExclude` in `pnpm-workspace.yaml` to land the fix now (bypasses the 7-day safeguard for posthog packages).
2. Hold this PR until ~2026-06-23 when `1.387.0`+ ages past the window, then merge with the policy intact.

Deliberately **not** modifying the policy in this PR — leaving CI red so the trade-off is visible and can be decided by the team.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mantagen:
Closing due to lack of appetite to fix, likely due to insufficient user impact

PR #4308, opened 2026-06-18
Exam board selector LESQ-2018

carlmw:
## Description

Music year: [2019](https://www.youtube.com/watch?v=4NRXx6U8ABQ)

- Add **Exam board (KS4)** radio filter to the teacher programme units page, shown after Year group when viewing a secondary KS4 exam-board programme (e.g. `science-secondary-aqa`)
- Switching exam board navigates via route slug change (`router.push` + `resolveOakHref`), not filter state
- Preload `examboardFilterDimensions` per board via new slim `curriculumSequenceSlugs` query and `buildExamboardFilterDimensions`, fetched in parallel inside `getProgrammeData`
- `getPreservedQuery` keeps compatible `keystages`, `years`, `tiers`, `pathways`, and `child_subjects` when changing board; drops values not valid for the destination board
- Thread `examboardFilterDimensions` through page → `ProgrammeView` → `UnitSequenceView` → desktop/mobile filters
- Refactor programme page data fetching to use granular `cacheData` helpers for phase options, overview, sequence, and sequence slugs
- Add `pathways` to `teacher-programme` URL query type in `urls.ts`

## How to test

1. Go to https://oak-web-application-website-git-feat-lesq-2108exam-board-faf7bb.vercel.thenational.academy/teachers/programmes/science-secondary-aqa/units?keystages=ks4&years=10&child_subjects=biology
2. Confirm the filter sidebar shows **Year group**, then **Exam board (KS4)**, with **AQA** selected and other boards (e.g. Edexcel, OCR) visible as radio buttons
3. Confirm units shown are KS4 Year 10 Biology units for AQA

**Also verify cross-board navigation preserves/drops filters correctly:**

4. On the same page, select **Foundation** tier (if shown) and ensure **Biology** remains selected under child subjects
5. Click **Edexcel**
6. URL should change to `/teachers/programmes/science-secondary-edexcel/units?...` (slug updates to `science-secondary-edexcel`)
7. Query params should retain `keystages=ks4`, `years=10`, and `child_subjects=biology` if Edexcel has Biology; tier should remain only if Edexcel supports the selected tier, otherwise it should be dropped from the URL

**Also verify the filter is hidden outside KS4 exam-board context:**

8. Go to https://oak-web-application-website-git-feat-lesq-2108exam-board-faf7bb.vercel.thenational.academy/teachers/programmes/science-secondary-aqa/units?keystages=ks3
9. **Exam board (KS4)** should not appear (KS3-only context)

**And verify no regression for non-exam-board programmes:**

10. Go to https://oak-web-application-website-git-feat-lesq-2108exam-board-faf7bb.vercel.thenational.academy/teachers/programmes/maths-primary/units
11. Page should load normally with Year group filter only — no **Exam board (KS4)** filter

## Screenshots

Screenshot 2026-06-18 at 11 05 25

## Checklist

- [ ] Design sign-off


src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ProgrammeFiltersExamBoard.tsx
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  return values.filter((value) => compatibleSlugs.includes(value));
+}
+
+function toQueryParam(values: string[]): string | undefined {
+  return values.length > 0 ? values.join(",") : undefined;
+}
+
+export function getPreservedQuery(
```
Comment (carlmw) [inline src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ProgrammeFiltersExamBoard.tsx:81]: filters the active filter params down to those that can actually be applied to the target examboard.

src/app/(core)/teachers/programmes/[slug]/[tab]/buildExamboardFilterDimensions.ts
```diff
@@ -0,0 +1,79 @@
...
+    boardDimensions.childSubjectSlugs.add(unit.subject_slug);
+  }
+};
+
+/**
+ * Derives which tier, pathway, and child-subject filters are valid for each KS4
+ * exam board option. Used when switching boards so incompatible query params are
+ * dropped rather than carried over.
+ */
+export function buildExamboardFilterDimensions(
```
Comment (carlmw) [inline src/app/(core)/teachers/programmes/[slug]/[tab]/buildExamboardFilterDimensions.ts:38]: We don't assume that all child-subjects, tiers and pathways are available for each examboard in a parallel programme. So to ensure that incompatible filters don't hide all units on the target programme we need to know what is valid for the target programme. So we pull the slugs for all units in all parallel examboard programmes to produce a comprehensive list of those that are valid in each.

src/node-lib/cache/index.ts
```diff
@@ -1,13 +1,26 @@
...
+ * cache returns. Caching transformed data couples the cache to the transformed schema;
+ * deploys that change that logic can serve stale shapes until revalidation.
+ * MV data is stable; bump `keyParts` or `revalidateTag` when the GQL query changes.
+ *
+ * `tags` are for invalidation (`revalidateTag`), not identity. Prefer page-level
+ * or domain tags (e.g. `curriculum-api`, `programme-page`) over revalidating shared
+ * key segments (e.g. `curriculum-overview` in multiple pages' `keyParts`).
+ *
+ * @see Next 16: replace with `"use cache"` + `cacheLife` / `cacheTag`.
  */
```
carlmw, line 24: Following yesterday's issues I figured it was a good idea to add some notes to help reduce the chance of issues in the future. Nothing bullet proof, so we still need a better process to prevent issues.

src/node-lib/curriculum-api-2023/queries/curriculumSequenceSlugs/curriculumSequenceSlugs.gql
```diff
@@ -0,0 +1,11 @@
...
+  $where: published_mv_curriculum_sequence_b_13_0_21_bool_exp
+) {
+  units: published_mv_curriculum_sequence_b_13_0_21(where: $where) {
+    examboard_slug
+    tier_slug
+    pathway_slug
+    subject_slug
+    subject_parent_slug
+  }
+}
```
carlmw, line 11: This needs to pull a lot of rows, so instead of using the existing `curriculumSequence` query this returns just the slugs for all units.

src/app/(core)/teachers/programmes/[slug]/[tab]/getProgrammeData.ts
```diff
@@ -1,13 +1,82 @@
...
+    curriculumApi2023.curriculumSequenceSlugs({
+      subjectSlug,
+      phaseSlug,
+      includeNonCurriculum: true,
+    }),
+  [PAGE_KEY, "curriculum-sequence-slugs"],
+  {
+    tags: [PAGE_KEY, CURRICULUM_API_CACHE_TAG],
+  },
+);
```
Comment (carlmw) [inline src/app/(core)/teachers/programmes/[slug]/[tab]/getProgrammeData.ts:79]: Splitting the cache so that it is closer to the query level. Another mitigation against yesterday's cache issues. Query changes should be less frequent 🤞🏻.

Though I nearly fell into the trap of bundling a lot of logic into the query which could bite in future if it needed changes.

Something to watch out for 😬

src/app/(core)/teachers/programmes/[slug]/[tab]/getProgrammeData.ts
```diff
@@ -1,13 +1,82 @@
...
+import { isExamboardSlug } from "@/pages-helpers/pupil/options-pages/options-pages-helpers";
+
+const PAGE_KEY = "programme-page-data";
+
+const getCachedCurriculumPhaseOptions = cacheData(
+  async () =>
+    curriculumApi2023.curriculumPhaseOptions({
+      includeNonCurriculum: true,
+    }),
+  [PAGE_KEY, "curriculum-phase-options"],
```
Comment (carlmw) [inline src/app/(core)/teachers/programmes/[slug]/[tab]/getProgrammeData.ts:28]: Prefixing with `PAGE_KEY` to create a namespace in the cache for just the programme page.

Something to be aware of (I wasn't because i've only used `"use cache";` in Next 16 - these keys can clash across the entire app. So if another function was only keyed as `"curriculum-phase-options"` with no arguments we'd see hits for whichever primed the cache first! This isn't the case with `"use cache";` which also keys on function identity.

So prepending `PAGE_KEY` ensures these are unlikely to clash (unless someone re-uses the same key combination on another page).

We might want to consider some sort of constants module to drive this? Or wrapping the curriculum API SDK directly and baking the caching in?

Needs thought.

src/app/(core)/teachers/programmes/[slug]/[tab]/getProgrammeData.ts
```diff
@@ -1,13 +1,82 @@
...
+const getCachedCurriculumOverview = cacheData(
+  async (subjectSlug: string, phaseSlug: string) =>
+    curriculumApi2023.curriculumOverview({
+      subjectSlug,
+      phaseSlug,
+      includeNonCurriculum: true,
+    }),
+  [PAGE_KEY, "curriculum-overview"],
+  {
+    tags: [PAGE_KEY, CURRICULUM_API_CACHE_TAG],
```
Comment (carlmw) [inline src/app/(core)/teachers/programmes/[slug]/[tab]/getProgrammeData.ts:43]: Tags would have helped us quickly deal with yesterday's incident. We could have invalidated the cache for a tag from within Vercel and avoided the rollback. https://vercel.com/oak-national-academy/oak-web-application-website/cdn/caches

`PAGE_KEY` would let us invalidate the cache for the entire page if we had issues.

I've also added `CURRICULUM_API_CACHE_TAG` so that we have ability to invalidate all entries related to the curriculum API. This might be useful going forward if we want to extend caching beyond the 2 hour ttl and instead have no expiry. We could trigger invalidation of that tag whenever the MVs finish refreshing.

RBrownsett:
@carlmw nice work. Couple of comments:

1. Is it possible to keep the focus within the examboard radio selector when moving between exam boards with a keyboard? In the preview tabbing into the exam board selector forces a page reload and moves focus to the top of the page. If we want to keep this behaviour then we might want to announce it as a link making it clear to user that they are being taken away from the filters. Preferably it would be better to return focus to the selected radio button after the page reloads - We might want to consider the mobile filter which closes the filter section on selection of an exam board, can the filter be kept open after a page reload? 

2. I think a follow on ticket and design will be needed to cover the pathway example below - allowing a user to switch between GCSE and Core

https://github.com/user-attachments/assets/ed4463c3-976e-4826-b9ac-c15fe26da2cb

src/app/(core)/teachers/programmes/[slug]/[tab]/getProgrammeData.ts
Comment (njblore) [reply src/app/(core)/teachers/programmes/[slug]/[tab]/getProgrammeData.ts:28]: 🤔  wrapping the sdk feels like the most foolproof solution for the long term, but perhaps not super trivial to set up, esp if we need to handle differently in pages vs app router?

src/app/(core)/teachers/programmes/[slug]/[tab]/getProgrammeData.ts
Comment (njblore) [reply src/app/(core)/teachers/programmes/[slug]/[tab]/getProgrammeData.ts:43]: `We could trigger invalidation of that tag whenever the MVs finish refreshing.` this sounds like it could be a nice win!

messmike:
> @carlmw nice work. Couple of comments:
> 
> 1. Is it possible to keep the focus within the examboard radio selector when moving between exam boards with a keyboard? In the preview tabbing into the exam board selector forces a page reload and moves focus to the top of the page. If we want to keep this behaviour then we might want to announce it as a link making it clear to user that they are being taken away from the filters. Preferably it would be better to return focus to the selected radio button after the page reloads - We might want to consider the mobile filter which closes the filter section on selection of an exam board, can the filter be kept open after a page reload?
> 2. I think a follow on ticket and design will be needed to cover the pathway example below - allowing a user to switch between GCSE and Core
> 
>  Screen.Recording.2026-06-18.at.14.37.01.mov

@RBrownsett here are designs for this example of pathways: https://www.figma.com/design/ktldjkEqAqv0X8PGzxrZyA/%F0%9F%9A%80-Integrated-journeys-v2?node-id=20021-22232&m=dev

Pathway example - filters

RBrownsett:
> > @carlmw nice work. Couple of comments:
> > 
> > 1. Is it possible to keep the focus within the examboard radio selector when moving between exam boards with a keyboard? In the preview tabbing into the exam board selector forces a page reload and moves focus to the top of the page. If we want to keep this behaviour then we might want to announce it as a link making it clear to user that they are being taken away from the filters. Preferably it would be better to return focus to the selected radio button after the page reloads - We might want to consider the mobile filter which closes the filter section on selection of an exam board, can the filter be kept open after a page reload?
> > 2. I think a follow on ticket and design will be needed to cover the pathway example below - allowing a user to switch between GCSE and Core
> > 
> > Screen.Recording.2026-06-18.at.14.37.01.mov
> 
> @RBrownsett here are designs for this example of pathways: https://www.figma.com/design/ktldjkEqAqv0X8PGzxrZyA/%F0%9F%9A%80-Integrated-journeys-v2?node-id=20021-22232&m=dev
> 
> Pathway example - filters

Brill thanks @messmike - @carlmw I can put this into a separate ticket if needed

src/app/(core)/teachers/programmes/[slug]/[tab]/getProgrammeData.ts
Comment (carlmw) [reply src/app/(core)/teachers/programmes/[slug]/[tab]/getProgrammeData.ts:28]: Yeah, might have to make it an opt-in so long as pages router is still in use.

My worry with abstracting to caching to that point is that you don't think about it when changing the query. We should most certainly include a hash of the `.gql` file (and possibly one of the entire GQL schema) as keys in that case to ensure that it doesn't pull stale cache entries.

carlmw:
> @carlmw nice work. Couple of comments:
> 
> 1. Is it possible to keep the focus within the examboard radio selector when moving between exam boards with a keyboard? In the preview tabbing into the exam board selector forces a page reload and moves focus to the top of the page. If we want to keep this behaviour then we might want to announce it as a link making it clear to user that they are being taken away from the filters. Preferably it would be better to return focus to the selected radio button after the page reloads - We might want to consider the mobile filter which closes the filter section on selection of an exam board, can the filter be kept open after a page reload?

@messmike @RBrownsett it's Friday and too hot 🫠 for my brain.

but this is really tricky to work around

**Returning focus might be possible**
when moving to another programme the entire content is reloaded, so we'd have to pass a query param to forward the focus requirement and quickly remove it once they've navigated (a bit messy but possible). Though it doesn't solve the real usability issue

**These should be links**
I should have seen this earlier, but the examboard "filters" are not filters they're navigation (like they are in the top nav and the [LHS of the unit page](https://oak-web-application-website-git-feat-lesq-2108exam-board-faf7bb.vercel.thenational.academy/teachers/programmes/combined-science-secondary-ks4-foundation-aqa/units/atomic-structure-and-the-periodic-table/lessons)). They're incongruous within the filters section as anything but links. Links let the user open the parallel programme in another tab which is a major UX tick.

Unfortunately we can't just wrap `OakRadioAsButton` in a link or add `role="link"` to the component and call it a day. 

Even more unfortunate is that the appearance of `OakRadioAsButton` isn't the same as `OakSecondaryButton` (which can be used as a link)

@messmike we've noted the same issue before which is why we use `OakSecondaryButton` on the [unit page for the tier and subject navigation IIRC](https://oak-web-application-website-git-feat-lesq-2108exam-board-faf7bb.vercel.thenational.academy/teachers/programmes/combined-science-secondary-ks4-foundation-aqa/units/atomic-structure-and-the-periodic-table/lessons)) 

So I see few ways forward:

1. I try to override the styles of `OakSecondaryButton` + `OakPrimaryButton` so that they match `OakRadioAsButton` (tricky and a little brittle) 
2. We do some work in oak-components to make `OakRadioAsButton` usable as a link (a considerable amount of work but would make for a more flexible component library)
3. We just use `OakSecondaryButton` + `OakPrimaryButton` like we do on the units page and accept that the examboard links will look different to the filters

👇🏻 this is what we'd get with #3

Screenshot 2026-06-19 at 10 29 40

Part of me thinks that the visual difference between the two controls is actually better for usability - it suggests two different interactions (though I do also want to take the path of least resistance 😇)

RBrownsett:
> > @carlmw nice work. Couple of comments:
> > 
> > 1. Is it possible to keep the focus within the examboard radio selector when moving between exam boards with a keyboard? In the preview tabbing into the exam board selector forces a page reload and moves focus to the top of the page. If we want to keep this behaviour then we might want to announce it as a link making it clear to user that they are being taken away from the filters. Preferably it would be better to return focus to the selected radio button after the page reloads - We might want to consider the mobile filter which closes the filter section on selection of an exam board, can the filter be kept open after a page reload?
> 
> @messmike @RBrownsett it's Friday and too hot 🫠 for my brain.
> 
> but this is really tricky to work around
> 
> **Returning focus might be possible** when moving to another programme the entire content is reloaded, so we'd have to pass a query param to forward the focus requirement and quickly remove it once they've navigated (a bit messy but possible). Though it doesn't solve the real usability issue
> 
> **These should be links** I should have seen this earlier, but the examboard "filters" are not filters they're navigation (like they are in the top nav and the [LHS of the unit page](https://oak-web-application-website-git-feat-lesq-2108exam-board-faf7bb.vercel.thenational.academy/teachers/programmes/combined-science-secondary-ks4-foundation-aqa/units/atomic-structure-and-the-periodic-table/lessons)). They're incongruous within the filters section as anything but links. Links let the user open the parallel programme in another tab which is a major UX tick.
> 
> Unfortunately we can't just wrap `OakRadioAsButton` in a link or add `role="link"` to the component and call it a day.
> 
> Even more unfortunate is that the appearance of `OakRadioAsButton` isn't the same as `OakSecondaryButton` (which can be used as a link)
> 
> @messmike we've noted the same issue before which is why we use `OakSecondaryButton` on the [unit page for the tier and subject navigation IIRC](https://oak-web-application-website-git-feat-lesq-2108exam-board-faf7bb.vercel.thenational.academy/teachers/programmes/combined-science-secondary-ks4-foundation-aqa/units/atomic-structure-and-the-periodic-table/lessons))
> 
> So I see few ways forward:
> 
> 1. I try to override the styles of `OakSecondaryButton` + `OakPrimaryButton` so that they match `OakRadioAsButton` (tricky and a little brittle)
> 2. We do some work in oak-components to make `OakRadioAsButton` usable as a link (a considerable amount of work but would make for a more flexible component library)
> 3. We just use `OakSecondaryButton` + `OakPrimaryButton` like we do on the units page and accept that the examboard links will look different to the filters
> 
> 👇🏻 this is what we'd get with #3
> 
> Screenshot 2026-06-19 at 10 29 40
> Part of me thinks that the visual difference between the two controls is actually better for usability - it suggests two different interactions (though I do also want to take the path of least resistance 😇)

Its certainly too hot for this! Looking at option 3 and having used the exam filters with the keyboard and seeing the focus issue (moving the user away from the filters) do we move the exam board out of this filter view? Appreciate thats a big change and maybe something we do as a feature after its gone out, but from a user point of view (especially in mobile) its quite disjointing for the filter draw to close when going through the options. 

Also its a little misleading, in RE for example, a user selects the filters they want e.g. a thread and subject category, then select the exam board filter and it resets and removes the filters they applied for the threads and subject categories.

carlmw:
Yeah, it certainly doesn't match expectations Let me see if I can trigger autofocus with some sort of temporary query param. The trick will be ensuring it doesn't jump focus when you open in a new tab which I don't think would be expected

messmike:
> > > @carlmw nice work. Couple of comments:
> > > 
> > > 1. Is it possible to keep the focus within the examboard radio selector when moving between exam boards with a keyboard? In the preview tabbing into the exam board selector forces a page reload and moves focus to the top of the page. If we want to keep this behaviour then we might want to announce it as a link making it clear to user that they are being taken away from the filters. Preferably it would be better to return focus to the selected radio button after the page reloads - We might want to consider the mobile filter which closes the filter section on selection of an exam board, can the filter be kept open after a page reload?
> > 
> > 
> > @messmike @RBrownsett it's Friday and too hot 🫠 for my brain.
> > but this is really tricky to work around
> > **Returning focus might be possible** when moving to another programme the entire content is reloaded, so we'd have to pass a query param to forward the focus requirement and quickly remove it once they've navigated (a bit messy but possible). Though it doesn't solve the real usability issue
> > **These should be links** I should have seen this earlier, but the examboard "filters" are not filters they're navigation (like they are in the top nav and the [LHS of the unit page](https://oak-web-application-website-git-feat-lesq-2108exam-board-faf7bb.vercel.thenational.academy/teachers/programmes/combined-science-secondary-ks4-foundation-aqa/units/atomic-structure-and-the-periodic-table/lessons)). They're incongruous within the filters section as anything but links. Links let the user open the parallel programme in another tab which is a major UX tick.
> > Unfortunately we can't just wrap `OakRadioAsButton` in a link or add `role="link"` to the component and call it a day.
> > Even more unfortunate is that the appearance of `OakRadioAsButton` isn't the same as `OakSecondaryButton` (which can be used as a link)
> > @messmike we've noted the same issue before which is why we use `OakSecondaryButton` on the [unit page for the tier and subject navigation IIRC](https://oak-web-application-website-git-feat-lesq-2108exam-board-faf7bb.vercel.thenational.academy/teachers/programmes/combined-science-secondary-ks4-foundation-aqa/units/atomic-structure-and-the-periodic-table/lessons))
> > So I see few ways forward:
> > 
> > 1. I try to override the styles of `OakSecondaryButton` + `OakPrimaryButton` so that they match `OakRadioAsButton` (tricky and a little brittle)
> > 2. We do some work in oak-components to make `OakRadioAsButton` usable as a link (a considerable amount of work but would make for a more flexible component library)
> > 3. We just use `OakSecondaryButton` + `OakPrimaryButton` like we do on the units page and accept that the examboard links will look different to the filters
> > 
> > 👇🏻 this is what we'd get with #3
> > Screenshot 2026-06-19 at 10 29 40
> > Part of me thinks that the visual difference between the two controls is actually better for usability - it suggests two different interactions (though I do also want to take the path of least resistance 😇)
> 
> Its certainly too hot for this! Looking at option 3 and having used the exam filters with the keyboard and seeing the focus issue (moving the user away from the filters) do we move the exam board out of this filter view? Appreciate thats a big change and maybe something we do as a feature after its gone out, but from a user point of view (especially in mobile) its quite disjointing for the filter draw to close when going through the options.
> 
> Also its a little misleading, in RE for example, a user selects the filters they want e.g. a thread and subject category, then select the exam board filter and it resets and removes the filters they applied for the threads and subject categories.

Hi @carlmw do you have the screengrab for the third option you mentioned? thanks

carlmw:
> > @carlmw nice work. Couple of comments:
> > 
> > 1. Is it possible to keep the focus within the examboard radio selector when moving between exam boards with a keyboard? In the preview tabbing into the exam board selector forces a page reload and moves focus to the top of the page. If we want to keep this behaviour then we might want to announce it as a link making it clear to user that they are being taken away from the filters. Preferably it would be better to return focus to the selected radio button after the page reloads - We might want to consider the mobile filter which closes the filter section on selection of an exam board, can the filter be kept open after a page reload?
> 
> @messmike @RBrownsett it's Friday and too hot 🫠 for my brain.
> 
> but this is really tricky to work around
> 
> **Returning focus might be possible** when moving to another programme the entire content is reloaded, so we'd have to pass a query param to forward the focus requirement and quickly remove it once they've navigated (a bit messy but possible). Though it doesn't solve the real usability issue
> 
> **These should be links** I should have seen this earlier, but the examboard "filters" are not filters they're navigation (like they are in the top nav and the [LHS of the unit page](https://oak-web-application-website-git-feat-lesq-2108exam-board-faf7bb.vercel.thenational.academy/teachers/programmes/combined-science-secondary-ks4-foundation-aqa/units/atomic-structure-and-the-periodic-table/lessons)). They're incongruous within the filters section as anything but links. Links let the user open the parallel programme in another tab which is a major UX tick.
> 
> Unfortunately we can't just wrap `OakRadioAsButton` in a link or add `role="link"` to the component and call it a day.
> 
> Even more unfortunate is that the appearance of `OakRadioAsButton` isn't the same as `OakSecondaryButton` (which can be used as a link)
> 
> @messmike we've noted the same issue before which is why we use `OakSecondaryButton` on the [unit page for the tier and subject navigation IIRC](https://oak-web-application-website-git-feat-lesq-2108exam-board-faf7bb.vercel.thenational.academy/teachers/programmes/combined-science-secondary-ks4-foundation-aqa/units/atomic-structure-and-the-periodic-table/lessons))
> 
> So I see few ways forward:
> 
> 1. I try to override the styles of `OakSecondaryButton` + `OakPrimaryButton` so that they match `OakRadioAsButton` (tricky and a little brittle)
> 2. We do some work in oak-components to make `OakRadioAsButton` usable as a link (a considerable amount of work but would make for a more flexible component library)
> 3. We just use `OakSecondaryButton` + `OakPrimaryButton` like we do on the units page and accept that the examboard links will look different to the filters
> 
> 👇🏻 this is what we'd get with #3
> 
> Screenshot 2026-06-19 at 10 29 40
> Part of me thinks that the visual difference between the two controls is actually better for usability - it suggests two different interactions (though I do also want to take the path of least resistance 😇)

@messmike yes that is what the screenshot here shows 

610372869-6419554d-7cdb-400b-9ea2-c751a1bcfcbd

src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ExamBoardFocus/ExamBoardFocusProvider.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * Exam board selection navigates to a new programme slug, remounting the filter
+ * and dropping focus. Descendants append ephemeral query params
+ * {@link FOCUS_KS4_OPTION_QUERY_PARAM} (and {@link OPEN_FILTERS_MODAL_QUERY_PARAM}
+ * from the modal tree) on push; this provider reads those params on load,
+ * re-opens the mobile filters modal when needed, focuses the matching radio in
+ * the correct scope, then strips the params from the URL.
+ */
+export function ExamBoardFocusProvider({
```
Comment (carlmw) [inline src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ExamBoardFocus/ExamBoardFocusProvider.tsx:44]: New a11y requirement was introduced to restore focus to the examboard panel after navigating to the new programme page.

This is complicated by the fact that we render filters in two places at once: sidebar on desktop and modal on mobile. Both stay mounted in the DOM (one is just hidden), so when a teacher changes exam board the app navigates to a new programme URL, the filter remounts, and keyboard focus is lost. We need to put focus back on the right radio button for accessibility, and on mobile we also need to reopen the filters modal if the change happened there. A single shared “focus manager” can’t tell which of the two filter trees the user actually used, so we split responsibilities: one small provider owns whether the mobile modal is open, another reads short‑lived URL flags after navigation and runs the focus‑restore flow, and a lightweight scope wrapper around each filter tree marks it as “page” or “modal” so restore and navigation know which subtree to target—without passing extra props through every layer or breaking the layout.

carlmw:
@RBrownsett I have introduced the logic to restore focus to the examboard panel after navigation. Hopefully the combination of page nav + focus movement doesn't cause any unintended a11y issues 🙏🏻 

I think we still need to make the radios into proper links to finish the job.

messmike:
> > > @carlmw nice work. Couple of comments:
> > > 
> > > 1. Is it possible to keep the focus within the examboard radio selector when moving between exam boards with a keyboard? In the preview tabbing into the exam board selector forces a page reload and moves focus to the top of the page. If we want to keep this behaviour then we might want to announce it as a link making it clear to user that they are being taken away from the filters. Preferably it would be better to return focus to the selected radio button after the page reloads - We might want to consider the mobile filter which closes the filter section on selection of an exam board, can the filter be kept open after a page reload?
> > 
> > 
> > @messmike @RBrownsett it's Friday and too hot 🫠 for my brain.
> > but this is really tricky to work around
> > **Returning focus might be possible** when moving to another programme the entire content is reloaded, so we'd have to pass a query param to forward the focus requirement and quickly remove it once they've navigated (a bit messy but possible). Though it doesn't solve the real usability issue
> > **These should be links** I should have seen this earlier, but the examboard "filters" are not filters they're navigation (like they are in the top nav and the [LHS of the unit page](https://oak-web-application-website-git-feat-lesq-2108exam-board-faf7bb.vercel.thenational.academy/teachers/programmes/combined-science-secondary-ks4-foundation-aqa/units/atomic-structure-and-the-periodic-table/lessons)). They're incongruous within the filters section as anything but links. Links let the user open the parallel programme in another tab which is a major UX tick.
> > Unfortunately we can't just wrap `OakRadioAsButton` in a link or add `role="link"` to the component and call it a day.
> > Even more unfortunate is that the appearance of `OakRadioAsButton` isn't the same as `OakSecondaryButton` (which can be used as a link)
> > @messmike we've noted the same issue before which is why we use `OakSecondaryButton` on the [unit page for the tier and subject navigation IIRC](https://oak-web-application-website-git-feat-lesq-2108exam-board-faf7bb.vercel.thenational.academy/teachers/programmes/combined-science-secondary-ks4-foundation-aqa/units/atomic-structure-and-the-periodic-table/lessons))
> > So I see few ways forward:
> > 
> > 1. I try to override the styles of `OakSecondaryButton` + `OakPrimaryButton` so that they match `OakRadioAsButton` (tricky and a little brittle)
> > 2. We do some work in oak-components to make `OakRadioAsButton` usable as a link (a considerable amount of work but would make for a more flexible component library)
> > 3. We just use `OakSecondaryButton` + `OakPrimaryButton` like we do on the units page and accept that the examboard links will look different to the filters
> > 
> > 👇🏻 this is what we'd get with #3
> > Screenshot 2026-06-19 at 10 29 40
> > Part of me thinks that the visual difference between the two controls is actually better for usability - it suggests two different interactions (though I do also want to take the path of least resistance 😇)
> 
> @messmike yes that is what the screenshot here shows
> 
> 610372869-6419554d-7cdb-400b-9ea2-c751a1bcfcbd

@carlmw I think this will be fine to go with the third option - thanks!

RBrownsett:
@carlmw restoring focus after navigation is a good improvement and addresses the main issue of users being returned to the top of the page. Thinking about the semantics, can we leave them as radio buttons? Links would imply navigation choices rather than filter selection. If we had them out of the filter and somewhere else on the page then links would be the better solution.

carlmw:
> @carlmw restoring focus after navigation is a good improvement and addresses the main issue of users being returned to the top of the page. Thinking about the semantics, can we leave them as radio buttons? Links would imply navigation choices rather than filter selection. If we had them out of the filter and somewhere else on the page then links would be the better solution.

@RBrownsett sure if you think that is best. Feels a bit wrong to me since it does perform a navigation. But then I guess in context that is an implementation detail.

RBrownsett:
> > @carlmw restoring focus after navigation is a good improvement and addresses the main issue of users being returned to the top of the page. Thinking about the semantics, can we leave them as radio buttons? Links would imply navigation choices rather than filter selection. If we had them out of the filter and somewhere else on the page then links would be the better solution.
> 
> @RBrownsett sure if you think that is best. Feels a bit wrong to me since it does perform a navigation. But then I guess in context that is an implementation detail.

I think so. I'll flag it as a future improvement, possibly move the selection away from the filters and follow the same implementation as subject categories on the science unit pages

RBrownsett (approved):
Approved, full testing notes on the ticket. Agreed that pathways and events will be handled in a separate ticket in the next sprint. Mobile view now keeps the modal open when selecting an exam board

src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ExamBoardFocus/ExamBoardFocusProvider.tsx
Comment (njblore) [reply src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ExamBoardFocus/ExamBoardFocusProvider.tsx:44]: Ohh boy 😬 I remember on the old unit listing page we had similar filters but we set it up so the filters did not apply until you click the "apply filters" button on mobile, I wonder if we could adopt that system here to remove the need to restore focus on mobile? (I suspect the only reason we apply the filters like we do here is because we've just adapted how the curriculum visualiser did them, rather than making a conscious product decision on it)

src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ProgrammeFiltersExamBoard.tsx
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  function onExamBoardChange(selectedSlug: string) {
+    const subjectPhaseSlug = getSubjectPhaseSlug({
+      subject: slugs.subjectSlug,
+      phaseSlug: slugs.phaseSlug,
+      examBoardSlug: selectedSlug,
+    });
+
+    router.push(
```
Comment (njblore) [inline src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ProgrammeFiltersExamBoard.tsx]: should we do a replace here to match the behaviour of the other filters?

src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ExamBoardFocus/ExamBoardFocusProvider.tsx
Comment (carlmw) [reply src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ExamBoardFocus/ExamBoardFocusProvider.tsx:44]: Yeah. I've struggled a bit with the dissonance of the exam board panel being treated as a filter when really it loads different content.

@RBrownsett was suggesting we lift the exam board filters out of the filter group and treat them as links which fits the model I have in my head.

I think inertia probably lead us down this path it's easy enough to use the same patterns and the examboard filter is a bolt on which runs against the grain.

One of those things you don't think about until you're building it unfortunately.

src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ProgrammeFiltersExamBoard.tsx
Comment (carlmw) [reply src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ProgrammeFiltersExamBoard.tsx]: 👍🏻 good shout

src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ExamBoardFocus/ExamBoardFocusProvider.tsx
Comment (njblore) [reply src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ExamBoardFocus/ExamBoardFocusProvider.tsx:44]: Yeah making them proper links would definitely help, maybe just something to consider if we revisit this again in future

njblore (approved):

PR #4310, opened 2026-06-18
Refactor/lesq 2065/search pages to app router

couplm:
## Description

Music year: 1956

- Move search pages to app router
- Honestly I don't know much about this having never worked with the pages system before, so when going in I thought it'd be simple but lots of things went wrong and I ended up shaving some yaks unexpectedly.
- I had copilot help with some of it, mainly the bit where it uses `compat` router to try and mediate between `next/router` and `next/navigation` so while copilot understands these React nuances better than I do, it could be janky in a way I don't understand yet.

## How to test

2. Search using the search bar
3. You should be able to see search results and navigate across pages correctly

## Checklist

- [ ] Design sign-off


couplm:
@copilot please write a summary of paths to test during manual QA

src/app/(core)/teachers/search/helpers/index.ts
```diff
@@ -7,25 +14,119 @@ import {
...
+    return "Subject filter" as FilterTypeValueType;
+  } else if (isContentTypeFilter) {
+    return "Content type filter" as FilterTypeValueType;
+  } else if (isExamBoardFilter) {
+    return "Exam board filter" as FilterTypeValueType;
+  } else if (slug === "new") {
+    return "Lesson Cohort filter" as FilterTypeValueType;
+  } else {
+    return "Unknown filter" as FilterTypeValueType;
+  }
```
carlmw: casting here hides type errors. "Unknown filter" isn't a valid `FilterTypeValueType`. 

If we're hitting that branch then we're probably encountering a bug and that will suppress it and send junk to tracking.

Since this is only tracking it's probably better to fail gracefully. So we can return `null` and use that to skip tracking with invalid properties.

Then we can report the error and deal with it. We're likely to hit this branch if data is released without schemas being updated.

`search/unknown-filter-type` isn't a real error code here, so would need an update in OakError to add a case for it.

```suggestion
export const getFilterType = (slug: string): FilterTypeValueType | null => {
  const isKeystageFilter = keystageSlugs.safeParse(slug).success;
  const isYearFilter = yearSlugs.safeParse(slug).success;
  const isSubjectFilter = subjectSlugs.safeParse(slug).success;
  const isContentTypeFilter = slug === "lesson" || slug === "unit";
  const isExamBoardFilter = examboardSlugs.safeParse(slug).success;

  if (isKeystageFilter) {
    return "Key stage filter";
  } else if (isYearFilter) {
    return "Year filter";
  } else if (isSubjectFilter) {
    return "Subject filter";
  } else if (isContentTypeFilter) {
    return "Content type filter";
  } else if (isExamBoardFilter) {
    return "Exam board filter";
  } else if (slug === "new") {
    return "Lesson Cohort filter";
  } else {
    reportError(new OakError({
      code: "search/unknown-filter-type",
      meta: {
        slug,
      },
    }));
    return null;
  }
```

src/app/(core)/teachers/search/page.test.tsx
```diff
@@ -19,43 +18,16 @@ const yearGroups = fixture.yearGroups;
 jest.mock("next/navigation");
 
 (usePathname as jest.Mock).mockReturnValue("/");
+(useSearchParams as jest.Mock).mockReturnValue(new URLSearchParams());
```
carlmw: just a nit, but Jest has a utility now that properly types mocked functions in place 🏄🏻 

```suggestion
jest.mocked(usePathname).mockReturnValue("/");
jest.mocked(useSearchParams).mockReturnValue(new URLSearchParams());
```

src/app/(core)/teachers/search/Search.view.tsx
```diff
@@ -88,8 +87,23 @@ const Search: FC<SearchProps> = (props) => {
...
+    const getParam = (param: string) => searchParams?.get(param) ?? "";
+    return {
+      keyStages: getParam("keyStages") || undefined,
+      examBoards: getParam("examBoards") || undefined,
+      contentTypes: getParam("contentTypes") || undefined,
+      subjects: getParam("subjects") || undefined,
+      yearGroups: getParam("yearGroups") || undefined,
+      curriculum: getParam("curriculum") || undefined,
+      page: getParam("page") || undefined,
+    };
```
carlmw: another nit, your helper here can also deal with reducing falsey values to `undefined`

```suggestion
    const getParam = (param: string) => searchParams?.get(param)?.trim() || undefined;
    return {
      keyStages: getParam("keyStages"),
      examBoards: getParam("examBoards"),
      contentTypes: getParam("contentTypes"),
      subjects: getParam("subjects"),
      yearGroups: getParam("yearGroups"),
      curriculum: getParam("curriculum"),
      page: getParam("page"),
    };
```

src/app/(core)/teachers/search/Search.view.tsx
```diff
@@ -548,6 +560,13 @@ const Search: FC<SearchProps> = (props) => {
                     hits={results}
                     allKeyStages={allKeyStages}
                     query={query}
+                    paginationNavigation={{
+                      route: pathname || "/",
+                      searchParams: searchParams
+                        ? new URLSearchParams(searchParams.toString())
+                        : null,
+                      push: (url: string) => router?.push(url),
+                    }}
```
carlmw, line 567: Pretty sure cloning searchParams is redundant here. `searchParams` is `ReadonlyURLSearchParams` which fits into `URLSearchParams` and anything that relies on it should be cloning rather than trying to mutate.

If you had a legitimate reason to clone feel free to ignore. 

```suggestion
                    paginationNavigation={{
                      route: pathname || "/",
                      searchParams,
                      push: (url: string) => router?.push(url),
                    }}
```

src/app/(core)/teachers/search/SearchView.tsx
```diff
@@ -0,0 +1,57 @@
...
+    allKeyStages,
+    allYearGroups,
+    allSubjects,
+    allContentTypes,
+    allExamBoards,
+    legacy: [{ slug: "new", title: "Show new only" }],
+    navigation: {
+      searchParams: searchParams
+        ? new URLSearchParams(searchParams.toString())
+        : null,
```
carlmw: Another clone that might not be needed? Still just a nit.

src/context/Search/useSearch.test.ts
```diff
@@ -49,70 +56,87 @@ jest.mock("@/common-lib/error-reporter/errorReporter", () => ({
     (...args: []) =>
       reportError(...args),
 }));
-const fetch = jest.spyOn(global, "fetch") as jest.Mock;
-jest.mock("next/dist/client/router", () => require("next-router-mock"));
```
carlmw, line 53: We should still be able to use `next-router-mock` for the compat import. The interface is the same.

`jest.mock("next/compat/router", () => require("next-router-mock"));`

Would spare us the big mock on LN70:93

might need some wider changes in the tests to use the mock interface, so no stress on this one.

src/context/Search/useSearch.ts
```diff
@@ -20,40 +24,51 @@ type UseSearchQueryReturnType = {
...
-    push,
-  } = useRouter();
+  const router = useCompatRouter();
+  const searchParams = useMemo(() => {
+    if (navigation?.searchParams) {
+      return navigation.searchParams;
+    }
+
+    return new URLSearchParams(router?.asPath?.split("?")[1] ?? "");
+  }, [navigation?.searchParams, router?.asPath]);
```
carlmw: `useSearchParams` works under both pages and app router so this could be reduced to:

```suggestion
  const appSearchParams = useSearchParams();
  const searchParams = navigation?.searchParams ?? appSearchParams;
```

Plus I think this hook is now only used under app router so maintaining back compat isn't needed

src/components/SharedComponents/Pagination/usePagination.ts
```diff
@@ -21,70 +18,104 @@ export type PaginationProps = {
...
   }, [currentPage, items, pageSize]);
 
   const firstItemRef = useRef<HTMLAnchorElement | null>(null);
   const paginationTitle =
     totalPages > 1 ? ` | Page ${currentPage} of ${totalPages}` : "";
-  const nextPageUrlObject = isLastPage
-    ? { pathname: router.asPath }
-    : { pathname: pathname, query: Object.fromEntries(nextPageParams) };
+
+  const baseParams = new URLSearchParams(searchParams.toString());
```
carlmw: Don't think cloning is needed if this is only used to build new `URLSearchParams`

src/components/SharedComponents/Pagination/usePagination.ts
```diff
@@ -21,70 +18,104 @@ export type PaginationProps = {
...
-  );
-  const currentPage = isNaN(pageNumber) ? 1 : pageNumber;
 
-  const nextPageParams = new URLSearchParams(encode(router.query));
-  nextPageParams.set("page", (currentPage + 1).toString());
+  const router = useRouter();
+  const route = navigation?.route || router?.asPath?.split("?")[0] || "/";
+  const searchParams =
+    navigation?.searchParams ||
+    new URLSearchParams(router?.asPath?.split("?")[1] ?? "");
```
carlmw: `usePathname` and `useSearchParams` can be used under both pages and app router. Only caveat being that they will be `null` on initial render. In this case that seems ok as it will settle quickly and the fallbacks will deal with .

```suggestion
  const pathname = usePathname();
  const appSearchParams = useSearchParams();
  const route = navigation?.route ?? pathname ?? "/";
  const searchParams = navigation?.searchParams ?? appSearchParams ?? new URLSearchParams();
```

src/components/SharedComponents/Pagination/usePagination.ts
```diff
@@ -21,70 +18,104 @@ export type PaginationProps = {
...
+      params.delete("page");
+    } else {
+      params.set("page", String(bounded));
+    }
+
+    if (navigation?.push) {
+      navigation.push(toHref(route, params));
+    } else {
+      void router?.push(toHref(route, params));
+    }
```
carlmw, line 124: My only concern with this approach is that under app router if `navigation` isn't set nothing will happen when clicking buttons. We'd pick that up in QA but the types don't reveal that.

We could add a comment to explain that or import both routers and fallback through them until one works.

```typescript
import { useRouter as useAppRouter } from "next/navigation";
import { useRouter as useCompatRouter } from "next/compat/router";

...

  const appRouter = useAppRouter();
  const compatRouter = useCompatRouter();

... 

  if (navigation?.push) {
    navigation.push(toHref(route, params));
  } else if (compatRouter) {
    compatRouter.push(toHref(route, params));
  } else {
    appRouter.push(url);
  }
```

carlmw (commented):
Porting pages to app router is a PITA. I've done a fair few now and it's finicky to maintain pages + app router compatibility. Many of the new app router hooks also work with pages (with the caveat that they're null on first render) so I think you can eliminate most uses of `useRouter`. it's only needed if you're doing actual routing

src/app/(core)/teachers/search/Search.view.tsx
```diff
@@ -15,13 +15,16 @@ import {
 import styled from "styled-components";
 import { useFeatureFlagVariantKey } from "posthog-js/react";
 
-import { SearchProps } from "./search.view.types";
+import { SearchProps } from "../../../../components/TeacherViews/Search/search.view.types";
```
njblore: keep an eye on the automatic imports we want to use the `@` alias wherever possible for readability, ie. `@/components/TeacherViews/Search/search.view.types`

src/app/layout.tsx
```diff
@@ -92,7 +93,9 @@ export default function RootLayout({
                   >
                     <AnalyticsWrapper>
                       <AppHooks />
-                      <SaveCountProvider>{children}</SaveCountProvider>
+                      <MenuProvider>
```
njblore, line 96: I am pretty sure MenuProvider is legacy and was just used for the old hamburger menu, is it needed in the app router? if so, I'd be questioning why 👀

src/components/SharedComponents/KeyStageKeypad/KeyStageKeypad.tsx
```diff
@@ -28,7 +28,9 @@ export type KeyStageKeypadProps = {
   keyStages: KeypadItem[];
 } & KeyStageOnClick;
 
-const KeypadLink: FC<KeypadItem & KeyStageOnClick> = (props) => {
```
njblore, line 31: I think this whole component is defunct and can be removed 🔥

src/app/(core)/teachers/search/page.tsx
```diff
@@ -0,0 +1,27 @@
...
+    index: false,
+    follow: true,
+  },
+};
+
+async function SearchPage() {
+  const curriculumData = await curriculumApi2023.searchPage();
+
+  return (
+    <Suspense fallback={null}>
```
njblore: does this Suspense do anything with a null fallback? and I'm not sure we want a suspense here anyway as this should be a statically rendered page right?

couplm: I think that an error told me to add it in, and it made sense to me at the time from a quick search but it is quite unfamiliar to me. I'll remove it again and check the error in case something else is wrong that's causing the warning

src/app/(core)/teachers/search/Search.view.tsx
couplm: I think I was having typing issues, for some reason it wasn't accepting them as the same in another area and can't remember if this was the same now. Will try this :)

src/app/layout.tsx
couplm: I got an error telling me it was necessary to add. I had to add it in to get other stuff moved over, but now they are moved I'll give removing it a go and see if I can figure out why it was required before

src/app/(core)/teachers/search/Search.view.tsx
carlmw: yeah, likely because `ReadonlyURLSearchParams` fits into `URLSearchParams`, but it will throw if you try to mutate the params. It's a defence inside Next to prevent you from altering the return of `useSearchParams` outside the router.

src/components/SharedComponents/Pagination/usePagination.ts
```diff
@@ -21,70 +23,108 @@ export type PaginationProps = {
...
-    ? { pathname: router.asPath }
-    : { pathname: pathname, query: Object.fromEntries(nextPageParams) };
+
+  const baseParams = new URLSearchParams(searchParams.toString());
+
+  const prevParams = baseParams;
+  if (currentPage - 1 <= 1) {
+    prevParams.delete("page");
+  } else {
+    prevParams.set("page", String(currentPage - 1));
```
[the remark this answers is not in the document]
carlmw: Sorry this was me not being clear in my comment, a clone is needed, but only once at the point of mutation.

The double cloning `baseParams = new URLSearchParams(searchParams)` -> `prevParams = new URLSearchParams(baseParams)` could just be `prevParams = new URLSearchParams(searchParams)`

RBrownsett:
@couplm looking good - just a small issue that on page load it shows the footer before focusing to the top of the page. I seem to remember seeing this issue when we made the the other pages in the app router and footer rendering before the main content. 

https://github.com/user-attachments/assets/6db18b5f-72f9-4c48-9f8e-7366db6c45f6

couplm:
> @couplm looking good - just a small issue that on page load it shows the footer before focusing to the top of the page. I seem to remember seeing this issue when we made the the other pages in the app router and footer rendering before the main content.
> Screen.Recording.2026-06-24.at.14.12.08.mov

@RBrownsett, thanks for flagging! Please could you try again when ready? I have had a look with throttling on and it appears to be OK, just a white screen fallback. Provided that's what we are wanting? Thanks!

RBrownsett:
@couplm page load now looks good, nice. One regression I've noticed is when you select a filter the page scrolls to the top. Before the page remained on the filters and user could select other without scrolling back down. I think adding some like { scroll: false }) when the search params are passed. Next app router docs seem to state that should work:

https://nextjs.org/docs/app/api-reference/functions/use-router#disabling-scroll-to-top

couplm:
> @/couplm page load now looks good, nice. One regression I've noticed is when you select a filter the page scrolls to the top. Before the page remained on the filters and user could select other without scrolling back down. I think adding some like { scroll: false }) when the search params are passed. Next app router docs seem to state that should work:
> 
> https://nextjs.org/docs/app/api-reference/functions/use-router#disabling-scroll-to-top

Thanks @RBrownsett ! I hadn't noticed that. Should be fixed now :)

RBrownsett (approved):
Tested across different devices and browsers, no further issues found. Events checked. Nice work

src/app/(core)/teachers/search/helpers/index.ts
```diff
@@ -1,15 +1,27 @@
-import { ParsedUrlQuery } from "querystring";
+import { ParsedUrlQuery } from "node:querystring";
```
carlmw: Not sure why this import changed but while the types for `node:querystring` are the same as the deprecated `querystring` package they're not the same package. 

Using node packages here introduces the possibility of them being pulled into the client side (though in this case it wont happen because this get tree shaken since the type is only used). It would likely blow up on you if a node built in package were imported - but who knows!

```suggestion
import type { ParsedUrlQuery } from "querystring";
```

Unless there was a genuine reason to change this i'd walk it back and make it a type only import.

`querystring` is a long dead package and everything should be using `URLSearchParams`, but that is a problem for another day.

Give me a shout if there is an issue

src/app/(core)/teachers/search/helpers/index.test.ts
```diff
@@ -1,23 +1,28 @@
-import { ParsedUrlQuery } from "querystring";
-
-import searchPageFixture from "../../node-lib/curriculum-api-2023/fixtures/searchPage.fixture";
+import { ParsedUrlQuery } from "node:querystring";
```
carlmw: Same package/type issue here
```suggestion
import type { ParsedUrlQuery } from "querystring";
```

carlmw (approved):
Just a thing about using a different package for type imports. Go go go 🥪

src/app/(core)/teachers/search/helpers/index.ts
couplm: I was confused reading this as I didn't remember doing it or know what it was about, but I can see now it is Sonarqube doing its thing. Thanks for this, wouldn't have known the difference otherwise :)

PR #4316, opened 2026-06-23
Chore/lesq 1935/remove app header [LESQ-1935]

njblore:
## Description

Music year: 1956

- Remove old `AppHeader` and related components
- Rename `Layout` -> `AppLayout` and remove extra wrapper component that was named AppLayout
- Remove reference to header height in old theme
- Create a `TopNavMinimal` component that doesn't use any context to be used on the error view page 
- On migrating the error view to use `TopNavMinimal` I realised the app router has no root level error boundary, so I added an `error.tsx` alongside the root layout which is applied as an error boundary. Unfortunately testing this using react dev tools doesn't seem to work with app router, I was able to verify it works locally

## How to test

1. Go to [OWA Preview URL from Vercel bot comment]
2. There should be no regressions


RBrownsett:
@njblore can main be merged into this PR for testing? I'm getting the top nav error that was fixed this morning. thanks

njblore:
@RBrownsett sorry I think it moved it into `ready to review` when I requested the copilot review! I'm just wrapping up the error handling I'll move it back across when actually ready

njblore:
@copilot can you get together a succinct list of paths or pages to review based on the changes in this PR for our human QA to review

src/hooks/useIsNewTopnavEnabled.ts
```diff
@@ -1,9 +0,0 @@
-import { useFeatureFlagEnabled } from "posthog-js/react";
-
-import { ENABLE_NEW_TOPNAV } from "@/config/flags";
```
carlmw, line 3: We could probably remove this config module too, it's just a single const

src/app/error.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const router = useRouter();
+
+  useEffect(() => {
+    errorReporter("app::error-boundary")(error);
+  }, [error]);
+
+  return (
+    <>
+      <TopNavMinimal />
+      <OakMaxWidth $alignItems={"flex-end"}>
```
carlmw, line 31: Not something for now, but we could add an error boundary that wraps the children in the core layout. That would keep the top nav and footer alive through client errors thrown inside the page.

carlmw (approved):
Yaks were shaved and ~1,500 LOC were removed 🍦

src/app/error.tsx
njblore: good shout, added a ticket 🚀

RBrownsett (approved):
Approved, very useful testing steps! Thanks

PR #4323, opened 2026-06-24
Feat/lesq 2068/onboarding pages to app router

Taha-Hassan-Git:
## Description

[Music year: 2017](https://www.youtube.com/watch?v=dp45V_M4Akw)

- Moves onboarding pages over to the app router
- Uses a layout component to enforce auth
- Tests for the above behaviour
- adds optional send_instantly argument to track event, enables by default on onboarding completed events

## Issue(s)

Fixes #lesq 2068

## How to test

2. Click on \_\_\_\_\_\_

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist

- [ ] Design sign-off


Taha-Hassan-Git:
@copilot Our QA is going to test the deployed application to ensure there are no regressions after this refactor. Help them by generating a check list of critical paths to test.

Taha-Hassan-Git:
Wow, testing this is a MUCH bigger headache than building it, sorry @RBrownsett  😢

src/components/TeacherComponents/OnboardingForm/onboardingDataQueryParam.ts
```diff
@@ -1,39 +1,39 @@
...
     return null;
   }
 }
 
 /**
- * Accumulates onboarding form data in a query param
+ * Accumulates onboarding form data, returning the encoded value for the `state`
+ * query param. Merges `data` on top of any previously encoded `state`.
+ *
+ * The caller is responsible for placing the returned value back into the URL,
```
njblore: What's the thinking behind having the caller handle wrapping and unwrapping the state from the query params, rather than centralise it here?

RBrownsett:
> Wow, testing this is a MUCH bigger headache than building it, sorry @RBrownsett 😢

@Taha-Hassan-Git haha yeah, very detailed list (but useful!). All the core journeys seem to be working, just a few issues listed below: 

1. Events seem to be hit or miss. I signed up around 10 times and only saw the userOnboardingCompleted twice. It only seems to fire when I'm signing up from a page already in the app router e.g. the unit page? I tried the home page in owa staging and the event fired, so I think it should be working 
2. userOnboardingProgressed I cannot get to a stage where that event is fired. Possible issue that is already present but might be worth investigating? 
3. Small visual regression when signing in from a page in app router e.g. unit or programme page, there is a visual flash of the Nearly done page (even though the user is already onboarded)  

https://github.com/user-attachments/assets/3ca14dcf-c031-49d5-80de-697a9e3eef10

src/components/TeacherComponents/OnboardingForm/onboardingDataQueryParam.ts
Taha-Hassan-Git: Hmm, the original idea was that it would work with either `router.query` or `useSearchParams`, but that's not flexibility that we actually need. So I'm going to centralise it as you suggested!

njblore: Aha that makes sense! cool

src/browser-lib/analytics/withQueue.ts
```diff
@@ -42,8 +43,14 @@ const withQueue = <T>(
...
 
-  const track: EventFn = (eventName, props) => {
-    queueEvent({ type: "track", eventName, props });
+  const track: EventFn = (eventName, props, options) => {
+    queueEvent({ type: "track", eventName, props, options });
+    // Instant events (e.g. those fired immediately before a navigation) can't
+    // wait for the next interval tick or the request may be cancelled before it
+    // leaves the browser. Flush now, while still respecting consent state.
+    if (options?.sendInstantly) {
+      conditionallyProcessQueue();
```
njblore, line 52: I'm not too sure what the queue system is solving for, so also not sure if we should be aware of any downsides bypassing it like this?

njblore (approved):
nice!🍦

src/browser-lib/analytics/withQueue.ts
Taha-Hassan-Git: It was 4 years ago but the person who put this stuff in was @mantagen. Maybe he has some insights.

Taha-Hassan-Git: It's this PR:

https://github.com/oaknational/Oak-Web-Application/pull/538

njblore: oooh right I see so the queue is just a way to process events with respect to the consent state, it seems like that's the main goal rather than there being an explicit need to queue. Which your change respects so seems like it should be all good

RBrownsett (approved):
Approved, looks good. Full testing notes on the Notion ticket

src/browser-lib/analytics/withQueue.ts
mantagen: Wow that feels like a _long_ time ago!

PR #4325, opened 2026-06-25
feat: add focus indicator and open in google slides button to lesson …

couplm:
…slides

## Description

Music year: [1967](https://www.youtube.com/watch?v=N_MqfF0WBsU) - it is really hard to find songs for anything before the 70s now

- Added a focus indicator to the slide deck embed.
- Took liberty to add the 'Open in Google Slides' button after the slide deck embed. Happy to change this but saw it added in the ticket as feedback and makes sense. Put it after the slide deck so users can focus the deck, see what it is called, then choose to open in a new tab. But maybe this is the wrong behaviour.

## How to test

1. Go to any OWA lesson with a slide deck
2. Focus the slide deck
3. You should see a focus indicator and the Open in Google Slides option

## Checklist

- [ ] Design sign-off


src/components/TeacherComponents/LessonOverviewPresentation/LessonOverviewPresentation.tsx
```diff
@@ -20,28 +42,54 @@ const LessonOverviewPresentation: FC<LessonOverviewPresentationProps> = ({
...
+        <OakSmallTertiaryInvertedButton
+          element={Link}
+          href={presentationEditUrl}
+          target="_blank"
+          rel="noopener noreferrer"
+          iconName="external"
+          isTrailingIcon
+          data-testid="open-presentation-in-google-slides"
+        >
+          Open in Google Slides
```
couplm: Open to feedback on this being something else! The recommended solutions included this but wasn't on the ticket. Just felt like adding focus might not be enough 🤷

src/components/TeacherComponents/LessonOverviewPresentation/LessonOverviewPresentation.tsx
```diff
@@ -20,28 +42,54 @@ const LessonOverviewPresentation: FC<LessonOverviewPresentationProps> = ({
...
-            border: "none",
-          }}
-          loading="eager"
-        />
-      </AspectRatio>
-    </OakBox>
+    <OakFlex $flexDirection="column" $gap="spacing-12">
+      <OakBox $ba={["border-solid-m"]} $width={"100%"}>
+        <AspectRatio ratio={isWorksheetPortrait ? "2:3" : "16:9"}>
+          <SlideEmbedFocusContainer
```
njblore: could we use [`OakFocusIndicator`](https://components.thenational.academy/?path=/docs/components-messaging-and-feedback-oakfocusindicator--docs) here?

src/components/TeacherComponents/LessonOverviewPresentation/LessonOverviewPresentation.tsx
njblore: I'm not 100% but I think we don't want to do this because we lose sight of users who download if they open in slides and download there

messmike: Good shout, might be good to check with @JoStrawOak on this point

JoStrawOak: That's correct. It's something that was introduced with the Cycle 2 work but we want to discourage. Anecdotally Stuart has observed a few computing teachers doing this. The fact that we see 10% lower downloads from users arriving via google than via bing suggests this might be the case. I actually want to discuss this in squad planning tomorrow to see what alternative options we have for the iframe (I dug out your 2 year old spike Tash). Suggest for now we avoid actively encouraging opening in google slides

couplm: Makes sense, thanks all! I will keep it to just focus.

RBrownsett (approved):
Approved- looks good! Focus now shows

messmike (approved):

src/components/TeacherComponents/LessonOverviewPresentation/LessonOverviewPresentation.tsx
```diff
@@ -23,25 +32,36 @@ const LessonOverviewPresentation: FC<LessonOverviewPresentationProps> = ({
...
+    <OakFlex $flexDirection="column" $gap="spacing-12">
+      <OakBox $ba={["border-solid-m"]} $width={"100%"}>
+        <AspectRatio ratio={isWorksheetPortrait ? "2:3" : "16:9"}>
+          <OakFocusIndicator
+            tabIndex={0}
+            style={{ width: "100%", height: "100%" }}
+            role="group"
+            data-testid="overview-presentation-focus-target"
+            aria-label={`${isAdditionalMaterial ? "Document" : "Slide deck"} preview for ${title}`}
+          >
```
njblore, line 71: focus is working but I'm not seeing the correct styling from `OakFocusIndicator` (should be grey and yellow border) - probably not a blocker since it does the job? 

Image

njblore (approved):
🔎

src/components/TeacherComponents/LessonOverviewPresentation/LessonOverviewPresentation.tsx
couplm: Thanks for pointing this out, I hadn't noticed the visual difference and figured `OakFocusIndicator` would've done it automatically, but can see now that it only seems to do links and buttons. I added some CSS to this file for now but admittedly not the best long term fix

PR #4326, opened 2026-06-25
Added `<SupportYou/>` component 

orangemug:
## Description
Added `<SupportYou/>` component 

## Issue(s)
Fixes `ADOPT-2109`

## How to test

1. Go to https://oak-web-application-storybook-git-feat-adopt-2109-suppor-188cbf.vercel.thenational.academy/?path=/story/components-genericpagescomponents-supportyou--default
2. Check storybook matches designs

**Note**: The image will differ, that will be added to sanity when we gets that bit done.

## Checklist

- [ ] Design sign-off


rightsaidjames (changes requested):
Two things:
* As per the [Figma a11y spec](https://www.figma.com/design/UDdUjFa4MJdM8AKCVxQJ5I/Final-designs---About-us?node-id=1-41481&m=dev), the icon on the button link should have alt text of 'Opens in new tab'. This can alternatively be achieved by setting an aria-label on the enclosing link of `Get in touch with an expert (opens in new tab)`
* The heading for this component should be a H2, but it seems to be set as H1 with no mechanism to change this via the component props.

rightsaidjames (approved):

src/components/GenericPagesComponents/SupportYou/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+};
+export function SupportYou({
+  title,
+  headingTag = "h2",
+  body,
+  link,
+  image,
+}: SupportYouProps) {
+  return (
+    <OakBox $pv={["spacing-56", "spacing-80", "spacing-80"]}>
```
alexcupit, line 33: Out of curiosity, is there a preference over whether to repeat or not if the value doesn't change between tablet and desktop? I think I had this on an nc-interface PR and feedback was to change to:
```
$pv={["spacing-56", "spacing-80"]}
```

alexcupit (approved):
LGTM! 

Does this also need adding into the Oak's Curricula page at some point? I saw it's marked as finished in figma but looks like it might be needed there as well as Oak's Impact? https://www.figma.com/design/UDdUjFa4MJdM8AKCVxQJ5I/Final-designs---About-us?node-id=1-32961&m=dev

src/components/GenericPagesComponents/SupportYou/index.tsx
orangemug: So I personally always do 3 values, but that's just "what I do" I don't think we have a standard approach

PR #4338, opened 2026-07-01
feat: show screen reader specific message using aria live polite on d…

couplm:
…ownload form error

## Description

Music year: 1966

- Added announcement for errors on the download form
- It seems that the copy link button on the media page does work...? It is possibly an NVDA issue though but I am unable to test with NVDA as it is Windows only

## How to test

1. Go to [OWA Preview URL from Vercel bot comment](https://oak-web-application-website-9781hwj31.vercel.thenational.academy//teachers/programmes/music-secondary-ks3/units/the-role-of-improvision-harmony-and-rhythm-in-blues-composition/lessons/the-use-of-the-drum-kit-in-grunge/media?video=12207)/teachers/programmes/music-secondary-ks3/units/the-role-of-improvision-harmony-and-rhythm-in-blues-composition/lessons/the-use-of-the-drum-kit-in-grunge/media?video=12207
2. Open Voiceover/other screen reader
3. You should focus the copy link button and press it
4. You should see the text 'Link Copied'
5. Click 'Back to Lesson'
6. Click 'Download all resources'
7. Deselect all resources, do not enter a school and do not check 'My school is not listed' and uncheck the terms and conditions checkbox
8. You should see the errors read by the screen reader

## Checklist

- [ ] Design sign-off


src/components/TeacherComponents/SharePageLayout/SharePageLayout.tsx
```diff
@@ -55,7 +55,7 @@ const SharePageLayout: FC<SharePageLayoutProps> = (props) => {
   const hasFormErrors = Object.keys(props.errors).length > 0;
   const validationErrorMessages = getFormErrorMessages(props.errors);
   const hasValidationSummary = validationErrorMessages.length > 0;
-  const validationSummaryAnnouncement = `To complete correct the following: ${validationErrorMessages.join(". ")}`;
+  const validationSummaryAnnouncement = `To complete, correct the following: ${validationErrorMessages.join(". ")}`;
```
Taha-Hassan-Git: This phrase is repeated a few times across different components. Probably better to extract into a const or a function?

Taha-Hassan-Git (commented):
Nice work, I have a non-blocking suggestion for an arguably slight improvement, but perhaps you already considered and decided against this.

At the moment the sighted and screen reader experience is slightly different. Sighted users get an unordered list and screen readers announce a string of text.

I think if you put the error messages in an always mounted div with `role="status"` and `aria-live="polite"`, it should be announced whenever the text content changes. We should be able to remove some complexity with parallel components and useEffects etc.

RBrownsett:
@couplm sorry my bad, I should have been clearer on the ticket. The media copy button works with voice over but not with Nvida (which the auditors use as its one of the most popular screen readers). Not sure if its possible to add the suggestion they made? I have a windows device so I can test it. 

The download error message is fixed which is good :-)

src/components/TeacherComponents/SharePageLayout/SharePageLayout.tsx
couplm: Thanks for these, Taha! Yep, I initially did it differently as Copilot was telling me it would not work if the validation was re-triggered. I couldn't see the issue happening on the page but added stuff to be safe. But actually I don't think it is necessary and I agree it would be better to have it simpler than address something that is probably not an issue :) I have updated to reflect your suggestions 🪩

Taha-Hassan-Git: Nice, and it still works? That's great!

couplm:
> @couplm sorry my bad, I should have been clearer on the ticket. The media copy button works with voice over but not with Nvida (which the auditors use as its one of the most popular screen readers). Not sure if its possible to add the suggestion they made? I have a windows device so I can test it.
> 
> The download error message is fixed which is good :-)

Sorry @RBrownsett I misunderstood it! I have updated it now and hopefully it works with NVDA 🤞

RBrownsett:
@couplm I'm learning with this ticket as well, still not working :-/ From some research adding aria-live="polite" (like they suggested) to a button itself doesn't reliably announce changes to the button's accessible name.

For WCAG 4.1.3 (Status Messages), the recommended pattern is to use a separate, persistent live region (role="status") and update its text when the copy action succeeds. Often a lot of the these accessibility tickets tend to be a bit whack a mole sadly. Let me know what you think.

Example code:

```
<button type="button" onClick={copyLink}>
  {copied ? 'Link copied' : 'Copy link'}
</button>

<span className="sr-only" role="status">
  {copied ? 'Link copied' : ''}
</span>
```

RBrownsett:
@couplm if it does become a pain we could split it out of this ticket and create a new one for the media page and I can do some more research

couplm:
> @couplm if it does become a pain we could split it out of this ticket and create a new one for the media page and I can do some more research

This might be best if you don't mind. I had a quick look and it would be possible but out of the scope of this ticket perhaps, at least without some weird fix (as far as I can tell). We might need to amend the `OakCopyLinkButton` to allow us to specify an action `onClick`, so the span isn't always there. Right now there is nothing to hook into for the `copied` check. This feels a bit out of scope. Or maybe there is a better way entirely that we can get this working :)

RBrownsett:
> > @couplm if it does become a pain we could split it out of this ticket and create a new one for the media page and I can do some more research
> 
> This might be best if you don't mind. I had a quick look and it would be possible but out of the scope of this ticket perhaps, at least without some weird fix (as far as I can tell). We might need to amend the `OakCopyLinkButton` to allow us to specify an action `onClick`, so the span isn't always there. Right now there is nothing to hook into for the `copied` check. This feels a bit out of scope. Or maybe there is a better way entirely that we can get this working :)

Yep, I agree makes sense. I'll pull this part out and make a new ticket

RBrownsett (approved):
Approved - download error messaging status alerts fixed. I have moved the media page copy link issue into this ticket https://app.notion.com/p/oaknationalacademy/Announce-status-messages-media-page-copy-link-39226cc4e1b1802a9edbfd32184658ae?source=copy_link

Taha-Hassan-Git (approved):
LGTM!

PR #4343, opened 2026-07-02
Test/lesq 2113/playwright setup

RBrownsett:
## Description

Creation of front end browsers tests. 

Adds initial Playwright E2E setup and first teacher lesson download journey test coverage. This is the first of two tickets, the CI workflow will follow in the second ticket. 

If these tests prove viable then improvements will be made:

- Move away from static lesson test and utilise a lesson from the database
- If and when additional tests are created a move to the POM framework will be accessed. 
- Increase browser coverage e.g. safari webkit 

Changes
Added Playwright dependency and config
Added E2E test for teacher lesson download flow
Ensured Jest does not run Playwright specs
Updated testing documentation and README link
Ran targeted spec locally and it passed

## Issue(s)

N/A 

## How to test

1. Check out this PR branch locally.
2. Ensure Node 24 is active: `nvm use 24`
3. Install dependencies: `pnpm install`
4. Start the app in one terminal
5. In a second terminal, run the Playwright teacher E2E spec: `pnpm run test:e2e -- src/tests/e2e/teacher/lesson-page.spec.ts`
6. Confirm both tests pass: **teacher can click download all resources on lesson page** and **teacher can complete download flow and download lesson assets**
7. Optional: open Playwright HTML report: `pnpm exec playwright show-report`
8. Optional: verify Jest does not pick up E2E specs: `pnpm run test:ci` Confirm it runs unit tests only and does not execute files under src/tests/e2e/

I am interested to see how the first test performs locally. In the CI there will be 1 retry however this is set to 0 locally and I did see in the initial version the test was sometimes flaky on a cold start.

## Screenshots

Playwright report example:

image

## Checklist

- [ ] Design sign-off


src/tests/e2e/teacher/lesson-page.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+  "/teachers/programmes/science-secondary-ks3/units/cells/lessons/the-common-processes-of-all-living-organisms";
+
+test("teacher can click download all resources on lesson page", async ({
+  page,
+}) => {
+  await page.goto(lessonPath);
+
+  const downloadAllButton = page.getByRole("link", {
+    name: "Download all resources",
+  });
```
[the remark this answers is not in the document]
RBrownsett: Only labelled download all in mobile. This is a desktop test, also the solution will not work as there are three elements with that name. I could update it to const downloadAllButton = page.locator('[data-testid="download-all-button"]:visible').first(); which may work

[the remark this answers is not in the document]
RBrownsett: I have updated to downloadAllButton = page.locator('[data-testid="download-all-button"]:visible').first();

docs/testing.md
```diff
@@ -39,6 +39,38 @@ Logic tests and snapshot tests should be in different files because their "failu
 
 The current tests live [here](../e2e_tests/browser/engineering/). The naming is intended to leave room for creating some product-facing tests using feature files, this is subject to change.
 
+## Playwright E2E
```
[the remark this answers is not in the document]
RBrownsett: This has been updated

docs/testing.md
```diff
@@ -35,9 +35,36 @@ Tests live next to the code they are testing wherever possible. Next does not al
...
+
+### Commands
+
+- `pnpm run test:e2e` runs all Playwright tests.
+- `pnpm run test:e2e -- src/tests/e2e/teacher/lesson-page.spec.ts` runs a single spec.
+- `pnpm run test:e2e:ci` runs Playwright with the HTML report enabled.
+
+### Local Execution
+
+- If `BASE_URL` is not set, Playwright uses `http://localhost:3000`, so the app must be running locally.
```
Taha-Hassan-Git: You can set something like: 

```
webServer: {
		command: 'pnpm run dev',
		url: 'http://localhost:3000',
		reuseExistingServer: true,
	}
```

In the config so we don't have to make sure the dev server is running before starting the tests.

Taha-Hassan-Git (commented):
Looks great! Before this lands though I would love to have [fixtures](https://playwright.dev/docs/test-fixtures) set up. It'll be mostly a bunch of boilerplate at the moment, but it'll be really useful as we add more and more pages and behaviours.

RBrownsett:
> fixtures

Good point I'll do some reading and get them added - looks like an improved version of the POM framework. Certainly would make sense as we build them out.

njblore (approved):
Nice work! Looking good and the tests pass locally ⭐ I am interested to hear more about the issues with flakiness you mentioned, as that sounds like something that could become problematic at a larger scale / in ci. I'm wondering if there's something we can do at the global level or via config to address it? Something to think about as we expand the test suite perhaps

couplm:
> Nice work! Looking good and the tests pass locally ⭐ I am interested to hear more about the issues with flakiness you mentioned, as that sounds like something that could become problematic at a larger scale / in ci. I'm wondering if there's something we can do at the global level or via config to address it? Something to think about as we expand the test suite perhaps

Anecdotally, using Cypress in my last job we would use `retries` and increase the number of ms that a step would wait before marking as failed. This would help, since sometimes tests just failed because they ran quickly on our laptops but slowly in CI. There were of course nightmare tests that you could never pinpoint but these 2 methods caught most of them 😅

RBrownsett:
> > Nice work! Looking good and the tests pass locally ⭐ I am interested to hear more about the issues with flakiness you mentioned, as that sounds like something that could become problematic at a larger scale / in ci. I'm wondering if there's something we can do at the global level or via config to address it? Something to think about as we expand the test suite perhaps
> 
> Anecdotally, using Cypress in my last job we would use `retries` and increase the number of ms that a step would wait before marking as failed. This would help, since sometimes tests just failed because they ran quickly on our laptops but slowly in CI. There were of course nightmare tests that you could never pinpoint but these 2 methods caught most of them 😅

Yeah, flaky tests are the worst and lose confidence in the suite, which is what I'm keen to avoid. So far I've set it to 1 retry in the ci (which will be the follow in ticket) and 0 for locally. I might up that to 1 as well. I've put a couple of things in place to try and combat the flakiness: 

- waiting for dom content to be loaded before interacting with the page 
- waiting until there is no element with the data-test id of loading (30 second wait)

Also I was trying to do too many asserts, when really I needed to keep the test clean and assert for specific behaviour. For example the **user can complete download flow and download lesson assets** test I was asserting the zip was downloaded in the browser and then checking that the success page was displayed. The problem is the download happens in the browser before the success page loads. One solution is too add a wait, but I think it might be cleaner to keep the tests specific to the behaviour. I could add a further test that checks the success page loads rather than using the download asset test. What do you think?

docs/testing.md
RBrownsett: Updated config and added a fixture file that handles navigation and page loads

Taha-Hassan-Git: Last thing to do is edit the readme so people know you don't have to have the server running.

Taha-Hassan-Git (approved):
I pulled this branch down and ran the tests, this is all working great at the moment. Just a couple of docs things. Firstly the devs will need to install playwright browsers before they can run the tests using this command: `pnpm exec playwright install` . We should definitely add that to the ReadMe

docs/testing.md
RBrownsett: Good spot, updated

PR #4344, opened 2026-07-02
Added oaks-impact stats component

orangemug:
## Description
Added oaks-impact stats component

## Issue(s)
Fixes `ADOPT-2113`

Enable the feature flag for oaks-impact

1. Go to https://oak-web-application-website-git-feat-adopt-2113-impact-stats.vercel.thenational.academy/about-us/oaks-impact
2. Check the stats sections matches the designs

Go to https://oak-web-application-storybook-git-feat-adopt-2113-impact-stats.vercel.thenational.academy/?path=/story/components-genericpagescomponents-oaksimpactstats--default and check it matches the designs.

## Screenshots
Screenshot 2026-07-07 at 12 47 11
Screenshot 2026-07-07 at 12 46 19

## Checklist

- [ ] Design sign-off


src/components/GenericPagesComponents/OaksImpactStats/OaksImpact.test.tsx
```diff
@@ -0,0 +1,31 @@
+import { OaksImpactStats } from "./";
+
+import { renderWithProvidersByName } from "@/__tests__/__helpers__/renderWithProviders";
+
+const render = renderWithProvidersByName(["oakTheme"]);
+
+describe("SupportYou", () => {
```
CatMillerCM:
```suggestion
describe("Oak's Impact Stats", () => {
```

src/components/GenericPagesComponents/OaksImpactStats/OaksImpactStats.test.tsx
CatMillerCM, line 1: OaksImpact.test.tsx > OaksImpactStats.test.tsx

src/components/GenericPagesComponents/OaksImpactStats/OaksImpact.test.tsx
orangemug: Not sure I like this, I think this should be `OaksImpactStats` as that's what we're testing. Also that's what we do elsewhere in the codebase

src/components/GenericPagesComponents/OaksImpactStats/OaksImpactStats.test.tsx
orangemug: Updated

src/components/GenericPagesComponents/OaksImpactStats/OaksImpact.test.tsx
orangemug: Fixed in latest

src/components/GenericPagesComponents/OaksImpactStats/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+  };
+  items: {
+    image: {
+      alt: string;
+      src: string;
+    };
+    headingText: string;
+    body: string;
+  }[];
+};
```
CatMillerCM: I'm wondering here if it would be easiest to follow the naming conventions from Sanity?
So the top level would be "title" and "body" which I think still work passing to OakHeading and OakP.
The "link" is probably fine as is but could always change to "cta", and change "text" to "label" (it's also getting passed into aria label and used as label for button so think that kinda makes sense)
I think "items" makes sense and is a convention we've used before, but could always make it specifically "stats"?
Then within them it might be nicer to have "heading" and "text" to match

src/components/GenericPagesComponents/OaksImpactStats/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+            <OakFlex $flexDirection={"column"} $gap={"spacing-40"}>
+              {props.items.map((item, index) => {
+                return (
+                  <Fragment key={index}>
+                    {index > 0 && (
+                      <OakBox
+                        $width="100%"
+                        $background="bg-decorative2-very-subdued"
+                        $height="spacing-2"
+                        $borderRadius="border-radius-s"
```
CatMillerCM, line 72: very nit 😬 could we add curly brackets are these values please so it aligns with rest of page, sorryyy haha

src/components/GenericPagesComponents/OaksImpactStats/OaksImpact.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+  it("renders correctly", () => {
+    const data = {
```
CatMillerCM: could we use the fixture here?

src/components/GenericPagesComponents/OaksImpactStats/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+    };
+    headingText: string;
+    body: string;
+  }[];
+};
+export function OaksImpactStats(props: Readonly<OaksImpactStatsProps>) {
+  return (
+    <OakFlex
+      $background={"bg-decorative2-main"}
+      $ph={["spacing-32", "spacing-40", "spacing-100"]}
```
CatMillerCM: In the previous about-us pages, we were using [NewGutterMaxWidth](https://github.com/oaknational/Oak-Web-Application/blob/bca1d9f1412786007c37080a4f6b224050ce9efb/src/components/GenericPagesComponents/NewGutterMaxWidth/index.tsx#L4) a fair amount in the components instead of passing these values top level, however it does differ slightly.

I've used it over in the [case studies PR](https://github.com/oaknational/Oak-Web-Application/pull/4357), but I wonder if there's a wider discussion on alignment here and the best way to ensure consistency across all pages (not even just about-us)

src/components/GenericPagesComponents/OaksImpactStats/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+      $pv={"spacing-80"}
+    >
+      <OakMaxWidth>
+        <OakFlex
+          $flexDirection={["column", "row", "row"]}
+          $gap={["spacing-32", "spacing-16", "spacing-120"]}
+        >
+          <OakFlex
+            $flexDirection={"column"}
+            $pr={["spacing-32", "spacing-32", "spacing-0"]}
```
CatMillerCM, line 30: Might be being obtuse here, but how come we need spacing-0 as right margin (or right margin specifically at all) here for desktop? It seems like behave nicely on the branch, but on storybook it makes it look a bit wonky? Is there a better way or is this needed?

Image

src/components/GenericPagesComponents/OaksImpactStats/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+                return (
+                  <Fragment key={index}>
+                    {index > 0 && (
+                      <OakBox
+                        $width="100%"
+                        $background="bg-decorative2-very-subdued"
+                        $height="spacing-2"
+                        $borderRadius="border-radius-s"
+                      />
+                    )}
```
CatMillerCM, line 74: instead of fragment and this index conditional block, could we use another flex and then pass a borderTop conditional? Feel like it keeps it a bit neater without excess blocks?
Something like
```
                  <OakFlex
                    key={index}
                    flexDirection={"column"}
                    $gap={"spacing-8"}
                    $borderColor={"bg-decorative2-very-subdued"}
                    $bt={index > 0 ? "border-solid-m" : undefined}
                  >
```
Seems to work locally for me just would need a bit of padding tweaking to get it centred between them?

Also maybe we should use `items.src` for the key instead of index?

src/components/GenericPagesComponents/OaksImpactStats/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+                        $flexDirection={"column"}
+                        $borderRadius={"border-radius-m2"}
+                        $gap={"spacing-4"}
+                      >
+                        <OakBox
+                          $color={"text-primary"}
+                          $font={"heading-light-1"}
+                        >
+                          {item.headingText}
+                        </OakBox>
```
CatMillerCM, line 101: can we use OakHeading here?

src/components/GenericPagesComponents/OaksImpactStats/OaksImpact.test.tsx
orangemug: Fixed in latest

src/components/GenericPagesComponents/OaksImpactStats/index.tsx
orangemug: It's not a heading, it's just some text. `<OakHeading/>` is for `h1-h6` tags and means something from an a11y sense

src/components/GenericPagesComponents/OaksImpactStats/index.tsx
orangemug: I'm going to ignore this one, if we want that sort of alignment we should get prettier to do it for us. Not worth a devs time fixing this kind of thing

src/components/GenericPagesComponents/OaksImpactStats/index.tsx
orangemug: An additional `<div/>` here seems better because we have perfect alignment between design/dev, and we haven't got to make changes.

> Also maybe we should use items.src for the key instead of index?

Indexes are frowned upon, but for something that never changes it's fine. The `src` although unlikely to be the same, could in theory be the same. Ideally all arrays of things should have an `id`, but here indexes are safe and we can ignore the lint warnings

src/components/GenericPagesComponents/OaksImpactStats/index.tsx
orangemug: Matches designs, see https://www.figma.com/design/UDdUjFa4MJdM8AKCVxQJ5I/Final-designs---About-us?node-id=1-40574&m=dev

src/components/GenericPagesComponents/OaksImpactStats/index.tsx
orangemug: Will update once https://github.com/oaknational/Oak-Web-Application/pull/4361 merges

CatMillerCM (approved):

src/components/GenericPagesComponents/OaksImpactStats/index.tsx
orangemug: Changed to use same naming/types as sanity

src/components/GenericPagesComponents/OaksImpactStats/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { NewGutterMaxWidth } from "../NewGutterMaxWidth";
+
+import { oaksImpactPageStatsSectionSchema } from "@/common-lib/cms-types/aboutPages";
+import CMSImage from "@/components/SharedComponents/CMSImage/CMSImage";
+import { PortableTextWithDefaults } from "@/components/SharedComponents/PortableText";
+import { getLinkHref } from "@/utils/portableText/resolveInternalHref";
+
+export type OaksImpactStatsProps = z.infer<
+  typeof oaksImpactPageStatsSectionSchema
+>;
```
CatMillerCM, line 19: think this is my knowledge gap, but why do we need to z.infer here?

orangemug: `z.infer<typeof schema>` returns you the type the schema would resolve to, so `{key: type}`. Whereas `typeof schema` will be a `ZodSchema`

rightsaidjames (approved):
LGTM! I have also tested the CMS functionality.

PR #4346, opened 2026-07-03
Feat/improve error boundary [LESQ-2138]

njblore:
## Description

Music year: 1988

- Add an error boundary for the `(core)` route which covers all of the teachers browse journey. As this sits below the core layout the Topnav and footer will be functional in this error view
- Add a `global-error` boundary at the app route, this will catch any issues in the root layout
- Add a component `SimulateErrorControls` to enable toggling error boundary locally, in theory this is possible with react dev tools but I haven't been able to get it work in app router
- Update the error boundary layout with a `Retry` button, this could resolve some rendering issues that ocur due to temporary blips in network connectivity, race conditions etc.

## How to test
Running locally set this env var in your `.env` file to enable the controls: `NEXT_PUBLIC_SIMULATE_ERROR="true"`

1. Go to localhost:3000/teachers/programmes/history-primary/units
2. You should see the controls for core and root layout
3. You should be able to toggle to trigger a rendering error
4. Click `Retry` and the page should re-render successfully
5. In Bugnsag you should see an error report for the level where the error was triggered: `app::global-layout`, `app::root-layout` or `app::core-layout`
7. With the env var not set, and in the preview environment, the error controls should not show

## Screenshots

**Error controls:**
Screenshot 2026-07-06 at 11 43 30

**Core Error Boundary (topnav functional):**
Screenshot 2026-07-06 at 11 44 28 

**Root Error Boundary:**
Screenshot 2026-07-06 at 11 44 52  

**Global Error Boundary:**
Screenshot 2026-07-06 at 11 45 33


src/app/(core)/layout.tsx
```diff
@@ -20,6 +21,7 @@ export default async function CoreLayout({
     return (
       <>
         <TopNav {...topNavProps} />
+        <SimulateErrorControls errorBoundaryLevel="root" />
         <main id="main">{children}</main>
```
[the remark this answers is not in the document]
njblore: I don't think this is correct, the nearest boundary is the one above it in the hierarchy not a sibling

src/app/global-error.tsx
```diff
@@ -0,0 +1,61 @@
...
+            $ph="spacing-100"
+          >
+            <ErrorFallback
+              error={error}
+              reset={reset}
+              reportError={reportError}
+              actionSlot={
+                <OakFlex $flexDirection="column" $gap="spacing-24">
+                  <OakTertiaryButton
+                    onClick={() => globalThis.location.reload()}
```
njblore, line 47: we don't have access to `router` here

src/app/components/ErrorHandling/ErrorFallback.tsx
```diff
@@ -0,0 +1,82 @@
...
+        </OakHeading>
+
+        <OakP $mb="spacing-24">Let's get you back to browsing</OakP>
+        {actionSlot ?? (
+          <OakFlex $flexDirection="column" $gap="spacing-24">
+            <OakTertiaryButton
+              onClick={() =>
+                startTransition(() => {
+                  reset();
+                  router.refresh();
```
couplm, line 64: Wonder if we can put metrics on the retry button? So we can see if the retry works and also how annoyed people might be 😅

couplm (approved):
this is cool!! 🤠

src/app/components/ErrorHandling/ErrorFallback.tsx
njblore: Yeah I thought that too but we don't have it set up for logging non-errors in bugsnag afaik, maybe one to look at when we move to sentry 👀

RBrownsett (approved):
Look good! Tested locally and in the PR. Errors logged in BugSnag dev as expected.

PR #4350, opened 2026-07-06
Chore/reduce prepush strain more

Taha-Hassan-Git:
## Description

Limits the amount of memory that tests will use up when running checks locally, only runs tests on changes since main on pre-push.

## Issue(s)

Fixes #

## How to test

1. pull branch and run tests locally
2. Push to remote origin and verify it only runs tests against main

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist

- [ ] Design sign-off


package.json
```diff
@@ -51,7 +51,8 @@
     "storybook:static-serve": "storybook build --quiet && http-server ./storybook-static --port=6006",
     "test": "jest --watch --collectCoverage=false",
     "test:coverage": "jest --watch",
-    "test:ci": "jest",
+    "test:ci": "jest --workerIdleMemoryLimit=1GB --maxWorkers=60%",
```
orangemug: This will slow down CI also though right? We probably don't want that do we?

package.json
```diff
@@ -51,7 +51,8 @@
     "storybook:static-serve": "storybook build --quiet && http-server ./storybook-static --port=6006",
     "test": "jest --watch --collectCoverage=false",
     "test:coverage": "jest --watch",
-    "test:ci": "jest",
+    "test:ci": "jest --workerIdleMemoryLimit=1GB --maxWorkers=60%", 
+    "test:prepush": "jest --changedSince=origin/main --collectCoverage=false",
```
orangemug, line 55: I guess the only concern is would we end up in a world where some tests only ever run on CI and we can't run them locally because of "some odd issue"

I'm unsure if I'm concerned, I guess tests could get flakier this way, thoughts @njblore ?

Taha-Hassan-Git: You can still run CI checks locally with `test:ci`, though I can understand some might be frustrated at something successfully pushing and then failing remotely.

package.json
Taha-Hassan-Git: This is true, I'll maybe add an extra script to run locally at lower cost, `test:ci-local` or `test:ci-light`. I'm not great at naming.

package.json
orangemug: It was more that most of us use CI on macos and the tests run on linux, so if we had a cross platform issue it could be a very long time for us to find it.

Not saying this is a blocker, just really wanted folks thoughts

package.json
orangemug: I'm confused why you're changing `test:ci`, don't you want to change `"test"` so running the tests doesn't use all the resources?

I also think workerIdleMemoryLimit should be a high percentage, because I think I would like the tests to run fast over everything else

If you want to override it on a per user bases you could set it to say 80% and them make it configuarable in the env, so something like

```
"test:ci": "jest --workerIdleMemoryLimit=${OWA_TEST_MEM_LIMIT:-80%} --maxWorkers=60%"
```

Which uses the shell defaults syntax

Taha-Hassan-Git: It's only really a problem when running all the tests. Running a single suite, or even a few is always completely fine. I could add it to all tests but it doesn't really seem necessary.

CatMillerCM (approved):

PR #4352, opened 2026-07-07
Fix/lesq 2063/hubspot fetch error [LESQ-2063]

njblore:
## Description

Music year: 2004

This error was being thrown on the new `eyfs` page, as far as I can tell this is the only place it was happening and would not have caused any lost sign ups.
The error was due to us trying to fetch the hubspot contact details on the EYFS page even when the user is signed out, they need to be signed in to download and for us to get hubspot details, so this constantly failed with`401 unauthorized` errors. Once the user signs in to download the issue is resolved.

To fix this I added a condition that the user must be signed in before we attempt to request hubspot details.
I additionally added some more error codes for hubspot api routes that we know can error as these were all being grouped under `hubspot/unknown` previously which isn't the most helpful.

## How to test

1. Go to https://oak-web-application-website-git-fix-lesq-2063hubspot-fet-c363dd.vercel.thenational.academy/teachers/eyfs/expressive-arts-and-design
3. While signed out: you should see no errors from hubspot in the console, and in the network tab you should see no requests to `contacts` from `fetchHubspotContactDetails`
4. While signed in: you should be able to download lessons, you should see a request to `hubspotFormSubmit`  with a valid payload 

Screenshot 2026-07-07 at 15 14 02


couplm (approved):
🤠

RBrownsett:
@njblore just checking I can see a request to contact-lookup when signed out. Is this a different endpoint to contacts?

image

njblore:
@RBrownsett yes `contact-lookup` is a different endpoint that doesn't require authentication

RBrownsett (approved):
Approved, tested with no issues, no regressions in lesson downloads and hubspot payload is correct. Good to go

PR #4356, opened 2026-07-08
Feat/browse analytics store [LESQ-2155]

njblore:
## Description

Music year: 1987

This PR initialises the `TeacherBrowseAnalyticsStore` as a centralised store to handle analytics in the teacher browse journey. To demonstrate I have hooked up two events on different pages (lesson and unit overviews) showing how the store is instantiated and used in components. In the future we could expand this store to handle much of the programme state for us, removing the need for prop drilling. 

- Create the store and provider
- Set up with two events for initial testing `unitDownloadInitiated` and `lessonResourceDownloadStarted`
    - This can coexist with the current tracking implementation until we have migrated all events into the store
- Add a method to get the session id to the analytics provider
- Create a `JourneyId` in the store using session id and programme slug, this can be appended to events. There is an open question about whether `programmeSlug` is appropriate here but that detail can be worked out later

## How to test

Unit download initiated
1. Go to https://oak-web-application-website-git-feat-browse-analytics-store.vercel.thenational.academy/teachers/programmes/english-primary-ks1/units/the-three-billy-goats-gruff-reading-and-writing/lessons
2. While signed in, click on download unit
3. You should see an analytics event `Unit Download Initiated` with the correct properties
4. Go to the lesson download success page
5. Click on download unit
6. You should see an analytics event `Unit Download Initiated` with the correct properties

Lesson Resource Download Started
1. Go to https://oak-web-application-website-git-feat-browse-analytics-store.vercel.thenational.academy/teachers/programmes/english-primary-ks1/units/the-three-billy-goats-gruff-reading-and-writing/lessons/reading-and-responding-to-the-three-billy-goats-gruff
2. Click download in the header
3. You should see an analytics event `Lesson Resource Download Started` with the correct properties, including the button value "all"
5. Go back to the lesson page
6. Click on download links on individual resources
7. You should see an analytics event `Lesson Resource Download Started` with the correct properties, including the button value which should respect the button that was clicked (eg. "slide deck")


src/context/TeacherBrowseAnalytics/teacherBrowseAnalytics.types.ts
```diff
@@ -0,0 +1,238 @@
...
+};
+export type LessonPathwayData = ProgrammePathwayData &
+  UnitPathwayData & {
+    lessonName: string;
+    lessonSlug: string;
+    lessonReleaseDate: string;
+  };
+
+// All Track Fns used in the teacher browse journey
+export type TeacherBrowseTrackFns = {
```
njblore, line 90: As it stands these are all the events used in the teacher browse journey that will need migrating. Some of these events will be deprecated and / or replaced as part of the event tracking upgrade work

Taha-Hassan-Git: I'm not sure of the benefit of this type at the moment. If it is necessary could we get these type definitions from avo instead somehow?

src/app/(core)/teachers/programmes/[slug]/units/[unitSlug]/lessons/[lessonSlug]/Components/LessonHeader/LessonHeader.tsx
```diff
@@ -50,39 +30,12 @@ const LessonHeader = (props: LessonHeaderProps) => {
...
-    examBoardTitle,
-    tierTitle,
-    pathwayTitle,
-    lessonSlug: currentLessonSlug,
-    lessonName: lessonTitle,
-    lessonReleaseDate,
-    isLegacy: false,
-  });
+  const { lessonResourceDownloadStarted } = useTeacherBrowseAnalytics(
+    (s) => s.track,
```
Comment (Taha-Hassan-Git) [inline src/app/(core)/teachers/programmes/[slug]/units/[unitSlug]/lessons/[lessonSlug]/Components/LessonHeader/LessonHeader.tsx]: Took a while to figure out what this was, can we stick to using `store` instead of `s` ? Makes it much more obvious what's happening.

Taha-Hassan-Git (approved):
Great change! Looking forward to migrating the rest of our functions into this store.

src/context/TeacherBrowseAnalytics/teacherBrowseAnalytics.types.ts
njblore: Once we've migrated all the tracking to the store this will be the type defining the `track` segment, which currently only has two functions. The arguments to each fn here are different to the avo definitions because once the store is populated it will supplement and provide the majority of the required properties - I've added it here now so we have a view of what we need to migrate and what properties each fn will require

RBrownsett (approved):
Approved. Tested both events, correct properties are pulled through. Download Resource Button Name tested across all assets. Regression test of lesson downloads. 

Out of interest where is the journeyId logged?

njblore:
@RBrownsett it's not logged anywhere at the moment as it's not used in any events, if you run the code locally you can add a console log to see it but I'd suggest we verify properly once we start to include it in events

PR #2107, opened 2023-11-22
feat: send hubspot share data LESQ-484

JBR90:
## Description

Music year: {owa_music_year}

-Sends data to HubSpot from share page on submit.

## How to test

1. Go to https://deploy-preview-2107--oak-web-application.netlify.thenational.academy
any share page, share resources, check HubStop received event
download events should still be captured

## Checklist

- [ ] Added or updated tests where appropriate
- [ ] Manually tested across browsers / devices
- [ ] Considered impact on accessibility
- [ ] Design sign-off
- [ ] Approved by product owner
- [ ] Does this PR update a package with a breaking change

PR #4361, opened 2026-07-13
Feat/adopt 2107/oaks impact sanity

alexcupit:
## Description

Music year: 2004

- introduces gql query for oaks impact page, using new dark mode image fragment/types 
- pulls through into Oak's impact page header component (excl video as on separate ticket)
- also added data into sanity [here](https://cms.thenational.academy/structure/pages;aboutUsNew;newAboutCorePage.oaksImpact)

## Issue(s)

Fixes #2107

## How to test

1. Go to [[OWA Preview URL from Vercel bot comment]/path/to/page](https://oak-web-application-website-git-feat-adopt-2107oaks-impa-3c5ad9.vercel.thenational.academy/about-us/oaks-impact)
2. You should see content from sanity

## Checklist

- [ ] Design sign-off


src/node-lib/sanity-graphql/queries/imageWithAltTextAndDarkMode.fragment.gql
```diff
@@ -0,0 +1,18 @@
+fragment ImageWithAltTextAndDarkMode on ImageWithAltTextAndDarkMode {
+  altText
+  isPresentational
+  asset {
+    ...ImageAsset
+  }
+  darkModeImage {
```
orangemug, line 7: Didn't notice this in the original CMS PR, but as we'll likely be using this new type "all the places" we should probably be sure about the naming.

`asset` and `darkModeImage.asset` seems somewhat odd, thoughts?

**Note:** We can fix this in a follow up ticket PR if we think it's worth changing, just throwing thoughts out there

orangemug (approved):
One comment, approved

src/node-lib/sanity-graphql/queries/imageWithAltTextAndDarkMode.fragment.gql
alexcupit: Yeah I see that. I didn't want to break anything that's already using `asset` to avoid lots of changes to existing code, and it would be nice to drop the `asset` from `darkModeImage` but since it uses existing sanity fragments for images, I don't think that's possible? At least the `gql-codegen:sanity` errored if it was a flatter query.

orangemug: Can you add a ticket to this sprint and we can discuss later on. We should get others thoughts outside out squad also once we'd made a decision, as it'll likely be used in "all the places" moving forward

CatMillerCM (approved):
lgtm!

PR #4363, opened 2026-07-14
Fix/lesq 2132/my lib english category [LESQ-2132]

njblore:
## Description

Music year: 1983

- update programme link on MyLibrary to use new `teacher-programme` url config
    - this required updating the query to pull through data to construct the `subjectPhaseSlug`
    - also updated the query params able to be passed through to the url query to include keystage
- add an extra transformation to convert `&` into `and` when getting the subject category slug 
    - unfortunately we only have the subject category title available in the MV, for a longer term fix we can either update the MV to pull the slug through, or use the slugified titles throughout the journey. 

## How to test

1. Go to https://oak-web-application-website-git-fix-lesq-2132my-lib-engl-9e2774.vercel.thenational.academy/teachers/programmes/english-primary/units?keystages=ks2&subject_categories=reading-writing-and-oracy
2. Save a unit to My Library
3. Go to My Library and click on the link to go to English KS2 programme
4. You should be navigated to same url as in step 1 with units showing
5. Regression test for url update on my library page:
    1. units in different keystages, 
    2. programmes go to correct pathway (core and gcse)
    3. examboards are maintained
    4. Subjects with parents go to the parent subject programme (ie. KS4 Biology -> Science) 

## Screenshots

Screenshot 2026-07-14 at 07 55 16

How it should now look:
Screenshot 2026-07-14 at 07 56 30


couplm (approved):

Taha-Hassan-Git (approved):
Looks good! But I think it would be nice to have a test to guard against regressions. Maybe a test of the MyLibrary component to check it's rendering the correct url?

RBrownsett:
@njblore carried out the testing steps and no issues. I did try to save a unit for rule of law and digital literacy but they were not added to the library. Is that more to do with staging data rather than the change made in this PR (I tried the same in the dependagames PR and those subjects did not show in my library either)

njblore:
@RBrownsett Ahh yes I think it's because of the rule of law data on staging is not matching production - basically we have a situation where in dev and preview environments we use the production curriculum data for data rendering, but we use the staging educator api which is synced with the staging curriculum database. So we save the unit based on the production data but try to fetch it based on the staging data, so if they don't match we can't find it. Staging rule of law unit slugs: 
Screenshot 2026-07-14 at 13 34 24

RBrownsett (approved):
Approved, looks good. Regression tested my library with no issues.

PR #4374, opened 2026-07-20
Feat/use case studies refs oaks impact

CatMillerCM:
## Description

- add case study schemas
- update query
- use case studies references in oak's impact
- update fixture and tests

## Issue(s)

Fixes ADOPT-2126

## How to test

1. Go to https://oak-web-application-website-git-feat-use-case-studies-re-df6ce7.vercel.thenational.academy/about-us/oaks-impact
3. See case studies section using references from [Sanity](https://cms.thenational.academy/structure/pages;aboutUsNew;newAboutCorePage.oaksImpact)

## Screenshots
image
image

## Checklist

- [ ] Design sign-off


src/components/GenericPagesComponents/OaksImpactCaseStudies/index.tsx
```diff
@@ -30,27 +28,30 @@ export const OaksImpactCaseStudies = ({
...
-                key={caseStudy.heading}
-                heading={caseStudy.heading}
+                key={caseStudy.slug.current}
+                heading={caseStudy.video.title || ""}
                 headingLevel={"div"}
-                href={caseStudy.href}
-                imageSrc={caseStudy.imageSrc}
+                href={`#/case-studies/${caseStudy.slug.current}`}
+                imageSrc={
+                  getProxiedSanityAssetUrl(caseStudy.image?.asset?.url) ?? ""
```
orangemug, line 50: I did not know this existed (CC @alexcupit)

CatMillerCM: This is what stops the need for adding the cdn block into `next.config.ts` 🙌

src/components/GenericPagesComponents/OaksImpactCaseStudies/index.tsx
```diff
@@ -30,27 +28,30 @@ export const OaksImpactCaseStudies = ({
...
             as="ul"
-            $gap={"spacing-16"}
-            $flexDirection={["column", "row"]}
+            $gridTemplateColumns={["1fr", "repeat(3, 1fr)"]}
+            $cg={"spacing-16"}
+            $rg={"spacing-16"}
             $pa={"spacing-0"}
             $ma={"spacing-0"}
           >
             {caseStudies.map((caseStudy) => (
```
orangemug, line 42: I think we probably want to cap this at `3` as the UI will start to look wonky above that number. Maybe we should put a "max" in the CMS if we haven't already, but either way max 3 here

orangemug (changes requested):

src/components/GenericPagesComponents/OaksImpactCaseStudies/index.tsx
```diff
@@ -30,27 +28,33 @@ export const OaksImpactCaseStudies = ({
...
               <OakCard
                 as="li"
-                key={caseStudy.heading}
-                heading={caseStudy.heading}
+                key={caseStudy.slug.current}
+                heading={caseStudy.video.title || ""}
                 headingLevel={"div"}
-                href={caseStudy.href}
-                imageSrc={caseStudy.imageSrc}
+                href={`#/case-studies/${caseStudy.slug.current}`}
```
alexcupit, line 48: I know the path isn't yet live, so not sure if this a placeholder, but looks like this is pointing to: 
`https://oak-web-application-website-git-feat-use-case-studies-re-df6ce7.vercel.thenational.academy/about-us/oaks-impact#/case-studies/tackling-blank-page-syndrome-with-aila`

src/components/GenericPagesComponents/OaksImpactCaseStudies/index.tsx
CatMillerCM: Hmm, I have actually just updated the grid to be `{["1fr", "repeat(auto-fit, minmax(200px, 1fr))"]}` so that you can add/remove list and they shrink/grow a bit, but depending on screen size they would drop to the next row if they were to go under 200px width of course. But this gives more flexibility. I'm wondering how much we'd want to enforce this? I feel like there's currently the scenario where we would want it to take 2 lis (other 2 case studies), and possible 4 in the future?
Maybe a thought for @HelenHarlow, do we want to limit the number of cards that can go in this section?

src/components/GenericPagesComponents/OaksImpactCaseStudies/index.tsx
alexcupit: Interesting! Looks like `OakCard` uses `OakImage` so no need for `CMSImage` here and therefore relies on the `url` rather than `_id` too

src/components/GenericPagesComponents/OaksImpactCaseStudies/index.tsx
CatMillerCM: yes sorry shouldve put a comment, it's placeholder for now :)

src/components/GenericPagesComponents/OaksImpactCaseStudies/index.tsx
HelenHarlow: Heya! We only ever want to have 3 here. When we have more case studies, we will be linking off to the case study library, so this is just to allow the user to directly access a few/ get a taste for the content. Hope that answers the question?

CatMillerCM: @HelenHarlow thanks! My only worry is at the minute we only have 3 videos, so for go live this section would only be able to have 2? 
image
Realise it's slightly updated in the new designs but think it would be the same issue?
image

src/components/GenericPagesComponents/OaksImpactCaseStudies/index.tsx
orangemug: Yeah, I think we can do away with `CMSImage` in a bunch of places, yay!

src/components/GenericPagesComponents/OaksImpactCaseStudies/index.tsx
HelenHarlow: So on the impact page itself, there will always be 3, but yes, you're right, on the case study page for a short amount of time, we will need to display 2. The cards themselves shouldn't look any different, it's just that we'll display 2 cards over 8 columns rather than 3 cards over 12 columns. I'm going to add some examples to the documentation to show how this should look. I'm not sure if that quite answers your question so let me know if there's anything i misunderstood

CatMillerCM: Thanks, that makes sense. So do we think I should put a limit of 3 in Sanity as it will never be over that? A user could add 1 or 2, but I will make sure the size of the cards would still be the same and centred?

HelenHarlow: I  think the min would be 2 and the max would be 3, but yes that sounds good! Thanks

rightsaidjames (approved):
LGTM

orangemug (approved):

PR #4386, opened 2026-07-22
Fix/adopt 2164/shrink legacy logos

CatMillerCM:
## Description

Makes legacy logos smaller and bring in curriculum text

## Issue(s)

Fixes ADOPT-2164

## How to test

1. Go to https://oak-web-application-website-git-fix-adopt-2164shrink-leg-745126.vercel.thenational.academy/about-us/oaks-curricula
2. Scroll to Legacy section
3. Change screen size to observe responsive behaviour

## Screenshots

image
image

How it should now look:
image
image

## Checklist

- [ ] Design sign-off


src/components/GenericPagesComponents/CurriculumPartners/index.tsx
```diff
@@ -47,11 +47,11 @@ export function CurriculumPartners({
       <OakBox>
         <OakGrid
           as="ul"
-          $gridTemplateColumns={[
-            "repeat(3, 1fr)",
-            "repeat(5, 1fr)",
-            "repeat(6, 1fr)",
-          ]}
+          $gridTemplateColumns={
+            title === "Current"
```
orangemug: Use a prop like `useSmallMode` or something like that instead of a string

orangemug (changes requested):

src/components/GenericPagesComponents/CurriculumPartners/index.tsx
CatMillerCM: @orangemug How do you feel about using the new `size` prop? initially this was only to be used when multiple size-related css properties were updates, but after a couple of discussions i was thinking about making it more flexible so it's just anything that impacts size. Here we would then pass something like `size = "sm"` almost like a variant, with default as "md". However this would still be a string but would be a specific prop instead of relying on the title staying the same, and would follow the [current accepted values](https://app.notion.com/p/oaknationalacademy/Oak-Components-Standards-v3-35726cc4e1b180f58ccef8cc58274b90?source=copy_link#35726cc4e1b1803dae7dd68750d72af9).

What do you think?

orangemug: Yup sounds good to me

CatMillerCM: updated in [c0b5919](https://github.com/oaknational/Oak-Web-Application/pull/4386/commits/c0b59193eae04f0b509a600ae8704d81d3a54ebd) - let me know what you think of the approach. Good chance to test out standards :D

orangemug: Something seems off with the word "config" but general approach I like

orangemug (approved):
LGTM

HelenHarlow (approved):
LGTM

rightsaidjames (approved):
Re-approving following additional changes

src/components/GenericPagesComponents/CurriculumPartners/index.tsx
CatMillerCM: hmmm, I know what you mean, think Weronika initially called it a variantConfig and then Reece and I agreed it would be good to have a separate config file. If there's a better word though I'm all ears and happy to update the standards 🙌

PR #4389, opened 2026-07-23
fix(pupil): avoid blocking navigation on classroom submission checks

jamiemax:
## description

- prevent submission checks from blocking pupil navigation
- keep the expected submission check asynchronous during destination page hydration
- prevent duplicate submission requests caused by the navigation handler and destination hydration running in quick succession
- add coverage for intro, video, quiz, and classroom add-on flows

## issue(s)

fixes [pupil-1779](https://app.notion.com/p/39e26cc4e1b1807ea575f66ab4e7f35c)

## how to test

1. go to `/pupils/programmes/computing-primary-year-4/units/the-internet/lessons/sharing-information/overview` using the classroom assignment query parameters from the ticket
2. open the browser network panel and filter for `classroom/submission`
3. enter the introduction, starter quiz, video, or exit quiz
4. complete the section and click `continue lesson`
5. you should navigate immediately without waiting for the submission request
6. the click should not start an additional blocking submission request
7. a single asynchronous submission request during destination page hydration is expected
8. repeat the navigation and confirm there are no duplicate submission requests in quick succession
9. for an assignment that has been handed in or returned, confirm the hydrated read-only state redirects the pupil to the review page

## screenshots

n/a 

## checklist

- [x] added or updated tests where appropriate
- [ x] manually tested across browsers / devices
- [x] considered impact on accessibility
- [x] design sign-off not required because there are no visual changes
- [ ] approved by product owner
- [ ] does this pr update a package with a breaking change

### DIFF
--- a/src/components/PupilComponents/Views/Hooks/usePupilExperienceBase.test.ts
+++ b/src/components/PupilComponents/Views/Hooks/usePupilExperienceBase.test.ts
@@ -58,19 +58,40 @@ describe("usePupilExperienceBase", () => {
     expect(routerPush).toHaveBeenCalledWith("/pupils/lessons/lesson-1/review");
   });
 
-  it("blocks progression and redirects when refresh resolves read-only", async () => {
-    const refreshReadOnly = jest.fn().mockResolvedValue(true);
+  it("allows progression immediately from cached writable state without refreshing", () => {
+    const refreshReadOnly = jest.fn(
+      () => new Promise<boolean>(() => undefined),
+    );
     usePupilLessonProgress.getState().setRefreshReadOnly(refreshReadOnly);
 
     const { result } = renderHook(() => usePupilExperienceBase());
     let canProgress: boolean | undefined;
-    await act(async () => {
-      canProgress = await result.current.ensureCanProgress();
+    act(() => {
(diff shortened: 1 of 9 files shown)


paulsweeting (changes requested):
@jamiemax  - ❌  Unfortunately, we cannot proceed with this solution, as it allows the Student to hand in at the final question of the quiz, which would store the results in Google Cloud. This seems to fail if the student hands-in the application and then they get stuck in a loop. The check that the assignment isn't in a state of 'TURNED_IN' is necessary to prevent this.

The delay in the button might just need to stay, or as you suggested introduce a loading spinner on the button as a visual indicator that it's loading.

In this scenario below, not only does the student get stuck but the final score is NOT sent to the teacher allowing them to view what the Pupil scored.

Steps:
1 - Complete the Exit Quiz up to the final question
2 - Before clicking 'Review Lesson' click 'Hand in'
3 - Click 'Review lesson'

**The state is not stored in Google Cloud, Pupil gets stuck, and the final score is not sent to the teacher.**

https://github.com/user-attachments/assets/cecf586b-b4cb-4422-a8a2-45adb5f16751

jamiemax:
@paulsweeting no problem, working on a follow up fix for this now

PR #4393, opened 2026-07-23
Feat/lesq 2172/sign out redirect [LESQ-2172]

njblore:
## Description

Music year: 2019

- Add a wrapper around the clerk provider so we can set `afterSignOutRedirectUrl` to the current page - clerk uses the homepage by default
- add a matcher for pages gated behind login so we can redirect these to the home page after sign out and prevent a loop where the user is redirected back to sign in

## How to test

**_Note_**: the sign out redirect url can only be declared when the page first initialises, so any updates to filters/query params that happen after first loatd will not be maintained when signing out 😞 confirmed by Clerk on their discord

1. Go to https://oak-web-application-website-git-feat-lesq-2172sign-out-redirect.vercel.thenational.academy/teachers/programmes/art-primary/units
2. Sign in
3. Sign out
4. You should remain on the same page
6. Repeat the above steps for different page types, eg. a unit, lesson, blog, campaign etc
7. Sign in and go to my library
8. Sign out
9. You should be redirected to the home page


src/browser-lib/appProviders/clerkProviderWithRedirects.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+// Pathnames for pages that are gated behind login
+const authRequiredPathnameRegex = /teachers\/my-library/;
+
+export function ClerkProviderWithRedirects({
+  children,
+  fontFamily,
+}: Readonly<{ children: React.ReactNode; fontFamily: string }>) {
+  const signOutUrl = useMemo(() => {
+    if (typeof window !== "undefined") {
```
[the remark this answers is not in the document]
njblore: yes this is noted as a limitation, it should only affect limited circumstances eg. when applying filters, as the clerk provider does not accept a dynamic value for the sign out url

src/browser-lib/appProviders/clerkProviderWithRedirects.tsx
```diff
@@ -0,0 +1,70 @@
+"use client";
+import { ClerkProvider } from "@clerk/nextjs";
+import { useMemo } from "react";
+
+// Pathnames for pages that are gated behind login
+const AUTH_REQUIRED_PATH_PREFIXES = ["/teachers/my-library"];
```
njblore, line 6: I don't love this approach but can't think of a better alternative, other than using the middleware to gate routes which we have ruled out due to latency issues

couplm: I was thinking it'd be better if it lived somewhere more visible at least, like some sort of config file as I used a lot in Angular applications, however it doesn't seem we have anything similar here. It is probably fine living here as at least it is bundled with the code, just wondering if it would be lost track of in future, especially if the number of pages grows

couplm (approved):
📮

RBrownsett (approved):
Looks good! Tested against different pages (along with copyright lessons) and the page loaded in the correct state.

PR #4411, opened 2026-08-04
Better feature flag implementation 

orangemug:
## Description
That supports server-time, build-time feature flags & override via an environment variable

Hopefully we can now use either this in server-side usage

```ts
isFeatureFlagEnabled(
    context,
    "oaks-impact"
);
```

Or this in build time usage, just remove the `context` as it's not available

```ts
isFeatureFlagEnabledAtBuild("oaks-impact");
```

@rightsaidjames can override any feature flag with `FORCE_FEATURE_FLAG_*` so in this instance `FORCE_FEATURE_FLAG_OAKS_IMPACT=true` in vercel which means we can't accidentally merge with features enabled. 

When we go live we can either set `FORCE_FEATURE_FLAG_OAKS_IMPACT=true` on production, or make a code change to change the default value to true (preferred)

## Issue(s)
Fixes `ADOPT-2183`

## How to test

1. Go to https://oak-web-application-website-git-feat-adopt-2183-better-f-9a52f1.vercel.thenational.academy/about-us/oaks-impact
2. Check that it still 404s with flag disabled

## Checklist

- [ ] Design sign-off


src/utils/featureFlagServer.ts
```diff
@@ -0,0 +1,51 @@
+import { GetServerSidePropsContext } from "next";
+
+import { getFeatureFlag } from "@/node-lib/posthog/getFeatureFlag";
+import { getPosthogIdFromCookie } from "@/node-lib/posthog/getPosthogId";
+import getBrowserConfig from "@/browser-lib/getBrowserConfig";
+
+export const FLAGS = {
+  get "oaks-impact"() {
```
rightsaidjames, line 8: It looks like this const is hardcoded for Oak's impact, so the `isFeatureFlagEnabledAtBuild` won't work for other feature flags. Is this intentional?

orangemug: Yeah you'd need to setup a mapping at the start of the work. So when you start it you'd add

```tsx
{
  get "some-feature" () {
    return process.env.FORCE_FEATURE_FLAG_SOME_FEATURE ?? "false";
  }
}
```

Then you should be able to use either of the following everywhere you need to

```tsx
isFeatureFlagEnabledAtBuild("some-feature")
isFeatureFlagEnabled(context, "some-feature")
```

rightsaidjames (approved):

CatMillerCM (approved):

PR #4414, opened 2026-08-06
Update to use portabletext input below video

orangemug:
## Description
Update to use portabletext input below video in case study page

## Issue(s)
None

## How to test

1. Go to https://oak-web-application-website-git-fix-impact-page-use-test-5572d8.vercel.thenational.academy/about-us/oaks-impact
2. Click on a case study
3. The text below the video should be now coming from the CMS

## Screenshots
Highlighted text is now from CMS

Screenshot 2026-08-06 at 12 14 13

## Checklist

- [ ] Design sign-off


src/pages/about-us/case-studies/[slug].tsx
```diff
@@ -120,9 +120,24 @@ const AboutUsOaksImpactCaseStudy: NextPage<
...
+                          marks: [],
+                        },
+                      ],
+                      markDefs: [],
+                      style: "normal",
+                    };
+                  }),
                 ]}
-                body={"Testing text"}
+                body={caseStudy.textRaw ?? []}
```
Comment (alexcupit) [inline src/pages/about-us/case-studies/[slug].tsx]: could this not just be `caseStudy.textRaw`? But if it's `undefined` (as opposed to `[]`) then it looks like there is spacing difference? Just thinking in other uses, it might be left without the body and then the layout would look slightly different.

alexcupit (commented):
Looks good to me, and works with my branched sanity graphql too 🎉 

Just a small question on the default spacing if there is no body text beneath the video.

Image

src/pages/about-us/case-studies/[slug].tsx
Comment (orangemug) [reply src/pages/about-us/case-studies/[slug].tsx]: Fixed in https://github.com/oaknational/Oak-Web-Application/pull/4414/commits/a6f328c7f02b27fcddcc5fc4bd1de1888da58526

We should also probably change `<OakVideo/>` to correctly handle `[]` (that can come later)

alexcupit (approved):

rightsaidjames (approved):
I have highlighted the risk of mismatch between the WYSIWYG editor and the styling supported by this component, but this shouldn't be a blocker for go-live IMO.

LGTM

PR #4417, opened 2026-08-07
Fix/lesq 2179/hubspot submit errors [LESQ-2179]

njblore:
## Description

Music year: {owa_music_year}

I reproduced this bug through these steps:
    1. sign into OWA 
    2. go to Hubspot and delete your contact* (you may have to delete all your alias contacts as well)
    3. go to a lesson download page 
    4. see the download button never loads
    
*even better if you can do this with a completely fresh user as hubspot seems to keep some knowledge of "deleted" users

This scenario could happen if there was an error during user onboarding and the user contact is never created in Hubspot at the point of sign up.

The Hubspot client also does not handle a missing user well and throws an error that we were not accounting for in our logic, as we wait for the Hubspot response to resolve before turning the loading state of the button off, but were never getting there in this case.

To resolve I added an extra piece of state so we know when the Hubspot request has completed whether successfully or not.

## How to test

1. Go to https://oak-web-application-website-git-fix-lesq-2179hubspot-sub-60e09b.vercel.thenational.academy/
5. Follow the steps above to reproduce the error - I don't have access to production hubspot so I did this while running the changes locally and using the sandbox hubspot account
6. You should be able to download the lesson resources even when the request to contacts fails

## Screenshots

When the request fails:
Screenshot 2026-08-07 at 09 43 46

How it should now look:
Request still fails but button loads:
Screenshot 2026-08-07 at 09 45 27


src/pages/api/hubspot/contacts/index.ts
```diff
@@ -1,15 +1,9 @@
 import { Client as HubspotClient } from "@hubspot/api-client";
 import { NextApiRequest, NextApiResponse } from "next";
-import { z } from "zod";
 import { getAuth, clerkClient } from "@clerk/nextjs/server";
 
 import getServerConfig from "@/node-lib/getServerConfig";
 
-export const contactResponseSchema = z.object({
```
njblore, line 8: this was not used anywhere

src/components/TeacherComponents/hooks/downloadAndShareHooks/useSyncHubspotAndLocalStorage.tsx
```diff
@@ -0,0 +1,202 @@
+import { useUser } from "@clerk/nextjs";
+import { useState, useEffect } from "react";
+import { UseFormSetValue } from "react-hook-form";
+
```
njblore, line 4: this is mostly just pulled out from useResourceFormState, the main new thing is `noSchoolFromHubspot` in the final useEffect hook to set the `hubspotLoaded` state which resolves the download button loading state

src/components/TeacherComponents/hooks/downloadAndShareHooks/useResourceFormState.tsx
```diff
@@ -1,13 +1,10 @@
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
 import { useForm } from "react-hook-form";
 import { useSearchParams } from "next/navigation";
 import { useRouter } from "next/compat/router";
-import { useUser } from "@clerk/nextjs";
 import { zodResolver } from "@hookform/resolvers/zod";
```
njblore, line 6: Changes in this file are mostly refactors to try to improve readability, and extracting the hubspot/localstorage logic into a separate hook

Taha-Hassan-Git (approved):
This feels like side-effect spaghetti! I find it very difficult to follow, but just looking at the changes you introduced they make sense to me.

Having worked more closely with these files do you think there's potential to simplify them, or are they this complex for a good reason? Not for this PR, of course, but possible future tech debt work.

njblore:
@Taha-Hassan-Git yes it totally is 😞 I think to improve it more substantially we'd need to rethink the overall architecture because as it stands we have multiple pieces of state (the form, localstorage, hubspot) that each need to be kept in sync leading to inevitable chaining. I don't know that it would be worth the effort required, but we could potentially have a ticket to scope it out and see if there is a redesign that could simplify the process before fully investing into doing the actual work?

Taha-Hassan-Git:
@njblore I think it's worth looking into, hard to see it getting prioritised at any point but there's no harm in making a ticket.

RBrownsett:
@njblore testing proved tricky as you mentioned hubspot does keep some data when the account is deleted (screenshot of a deleted account below - it displays the no school error message)

image

I did test it by blocking the contacts url in the Networks tab. In production blocking causes the 'Loading' button issue to display but in the PR I am still able to download resources even when the contacts url is blocked. So I think we should be confident that it will fix the issue

RBrownsett (approved):
Approved

PR #4418, opened 2026-08-07
Feat/adopt 2171/oaks impact nav footer

CatMillerCM:
## Description

- Conditionally render Oak's Impact link in footer if isFeatureFlagEnabledStatic("oaks-impact") is true
- Conditionally render Oak's Impact link in top nav query if isFeatureFlagEnabledStatic("oaks-impact") is true
- Conditionally render Oak's Impact link in top nav fixture if isFeatureFlagEnabledStatic("oaks-impact") is true
- Rejig top nav menu items slightly according to Helen
- Conditionally render Oak's Impact link in Explore more about Oak section if isFeatureFlagEnabledStatic("oaks-impact") is true
- Mock the feature flag to be false in about-us page tests so it doesn't update snaps right now
- Create a `featureFlagChecks` util folder which houses a static (const only) and a server side (posthog and const) check in separate files. 
- Also updated the env var to support client side
- These changes stop errors from trying to import server side in static build pages (even if only wanting to use static)
- Happy for renaming suggestions!
- Update tests to reflect all changes

I have updated the feature flag var to be `true` for this branch here: https://vercel.com/oak-national-academy/oak-web-application-website/settings/environment-variables

## Issue(s)

Fixes ADOPT-2171

## How to test

1. Go to https://oak-web-application-website-git-feat-adopt-2171oaks-impa-e1737b.vercel.thenational.academy/about-us/oaks-curricula
2. Click on 'About Us' in the top nav
3. You should see Oak's Impact added
4. Scroll to the 'Explore more about Oak` section
5. You should see an 'Oak's impact' button and the layout as 2x2
6. Scroll to the footer
7. You should see 'Oak's Impact' underneath Oak's curricula

If you flip the var to `false` in vercel env vars for the branch, or pull down and flip it locally, you should not see Oak's Impact in the top nav, footer, or explore more about oak section.

## Screenshots

When feature flag is `false`
image
image
image

When feature flag is `true`
image
image
image

## Checklist

- [x] Design sign-off


orangemug (approved):
LGTM

src/pages/about-us/oaks-impact.tsx
```diff
@@ -57,7 +57,7 @@ const OaksImpact: NextPage<OaksImpactPageProps> = ({ topNav, pageData }) => {
 };
 
 export const getServerSideProps: GetServerSideProps = async (context) => {
-  const isImpactPageEnabled = await isFeatureFlagEnabled(
+  const isImpactPageEnabled = await isFeatureFlagEnabledServer(
```
alexcupit, line 60: Will this want changing to `getStaticProps` once we stop fetching the flag from posthog? Just wondering if it makes sense to change this now and then use the one feature flag source of truth since `isFeatureFlagEnabledServer` checks the `isFeatureFlagEnabledStatic` first anyway?

src/pages/about-us/case-studies/[slug].tsx
```diff
@@ -188,7 +188,7 @@ export const getServerSideProps: GetServerSideProps<
   AboutUsOaksImpactCaseStudyPageProps,
   URLParams
 > = async (context) => {
-  const isImpactPageEnabled = await isFeatureFlagEnabled(
+  const isImpactPageEnabled = await isFeatureFlagEnabledServer(
```
Comment (alexcupit) [inline src/pages/about-us/case-studies/[slug].tsx:174]: same here and whether this will change to `getStaticProps`?

alexcupit (commented):
Looks good! Just a question on whether it's worth removing the server check at this point and then we could use `getStaticProps` instead for those locations which I think we said we'd want to move to eventually?

src/pages/about-us/case-studies/[slug].tsx
Comment (CatMillerCM) [reply src/pages/about-us/case-studies/[slug].tsx:174]: @orangemug What was the reason we said we wanted to keep posthog feature flag in for now again? My brain is drawing a blank but I swear we discussed this haha

Comment (orangemug) [reply src/pages/about-us/case-studies/[slug].tsx:174]: It's easier for folk to test and should be "the way to do things", we'll need to change over in the next couple of days but it make things easier in the mean time with all these PRs branches flying about

Comment (alexcupit) [reply src/pages/about-us/case-studies/[slug].tsx:174]: Yeah that makes sense, but with the new const flag, doesn't that mean we're never checking against the value in posthog, since if there's a const it never makes the API call?

Comment (CatMillerCM) [reply src/pages/about-us/case-studies/[slug].tsx:174]: @alexcupit so it checks if the const is true and if so early exits (returns true), but if it's false it still goes to check posthog and can still have the feature switched on there.
I updated the tests to hopefully make this a bit clearer too: https://github.com/oaknational/Oak-Web-Application/pull/4418/changes#diff-908d7b2bee63ee931388700fb7dc5b03056ebe98b1a3f0311d73dda035201ee1

rightsaidjames (approved):
LGTM, I've now tested it with both  possible env var values.

PR #4420, opened 2026-08-07
Feat/adopt 2169/impact survey

CatMillerCM:
## Description

- Creates a `TrackScrolledTo` component that renders an OakBox and triggers the `track.scrolledTo` event when it enters the view port. Will only trigger the first time this happens so as not to spam the event with up and down scrolling.
- Creates a `useTrackExitIntended` hook that calls `track.exitIntended()` when the user leaves the viewport from the top of the window. Will only trigger the first time this happens so as not to spam the event with mouse in and out.
- Add tests for both of these
- Bring in both to `oaks-impact` and call `useTrackExitIntended();` and render `<TrackScrolledTo eventKey="support_you" />` above `SupportYou`

## Issue(s)

Fixes ADOPT-2169

## How to test

1. Go to https://oak-web-application-website-git-spike-adopt-2169impact-survey.vercel.thenational.academy/about-us/oaks-impact
2. Move your mouse off the top of the page
3. You should see the test survey appear
4. Click off the test survey
5. Scroll down to the `Support You` section
6. You should see the test survey appear again
7. Note: in posthog this will be configured to not happen to users irl
8. You can also check in the console that the avo event is sent (see screenshots below)

## Screenshots
image
image
image

## Checklist

- [ ] Design sign-off


src/node-lib/avo/Avo.ts
CatMillerCM, line 1: Reminder: This needs reverting before merge

src/components/SharedComponents/TrackScrolledTo/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+  useEffect(() => {
+    if (!ref.current) {
+      return;
+    }
+
+    const observer = new IntersectionObserver(
+      ([entry]) => {
+        if (entry?.isIntersecting && !hasTracked.current) {
+          hasTracked.current = true;
+          track.scrolledTo({ key: eventKey });
```
alexcupit, line 29: just a reminder, we thought `key` might need updating depending on Ewan's feedback

alexcupit (approved):
Looks good! Just a reminder we need @ewan-wakeman's review in Avo before merging: https://www.avo.app/schemas/5PhajbVijwhXVKIJtGMT/branches/A_5MN1Ql_Y47ciClDU4i3/diff

src/hooks/useTrackExitIntended.tsx
```diff
@@ -0,0 +1,32 @@
...
+    const handleMouseOutOfWindow = (e: MouseEvent) => {
+      if (hasTrackedExitIntent.current) {
+        return;
+      }
+
+      // Check if cursor moves outside the top of the window viewport
+      if (e.clientY <= 0 && !e.relatedTarget) {
+        hasTrackedExitIntent.current = true;
+        track.exitIntended();
+        window.removeEventListener("mouseout", handleMouseOutOfWindow);
```
orangemug, line 23: Do we only want to do this once? We can only assume that the user has tried to exit, I guess we'll never actually know for sure

Thoughts folks?

src/components/SharedComponents/TrackScrolledTo/TrackScrolledTo.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+jest.mock("@/context/Analytics/useAnalytics", () => ({
+  __esModule: true,
+  default: jest.fn(),
+}));
+
+const mockUseAnalytics = jest.mocked(useAnalytics);
+
+describe("TrackScrolledTo", () => {
+  let ioCallback: IntersectionObserverCallback;
```
orangemug: nit: observer callback is probably nicer, I assume i/o when I was reading this first time

src/hooks/useTrackExitIntended.tsx
CatMillerCM: I added logic so that both events only trigger the first time so as not to spam the event. I feel like this is probably more appreciated for our logs, but obviously for the user it wouldnt make much difference as the survey would only show once anyway if the setting was applied in posthog directly.

src/components/SharedComponents/TrackScrolledTo/TrackScrolledTo.test.tsx
CatMillerCM: good shout - I copied it from another test but have updated in this file in [eb305da](https://github.com/oaknational/Oak-Web-Application/pull/4420/commits/eb305da25bf3405399e8449e1080cd8308548eab)

rightsaidjames (approved):
Looks good! Have tested both events in Chrome, Firefox and Safari

orangemug (approved):

PR #4421, opened 2026-08-10
Feat/adopt 2173/impact sitemaps

CatMillerCM:
*Not to be merged until release ready*

## Description

- include oaks-impact in sitemaps
- include oaks-impact and a case study page in deployment urls for testing

## Issue(s)

Fixes ADOPT-2173

## Checklist

- [ ] Design sign-off


src/common-lib/urls/getDeploymentTestUrls.js
```diff
@@ -14,6 +14,8 @@ function getDeploymentTestUrls() {
     // About Us pages
     "/about-us/who-we-are",
     "/about-us/oaks-curricula",
+    "/about-us/oaks-impact",
```
orangemug, line 17: Can't we put this behind the build flag?

orangemug: This was what I meant @CatMillerCM https://github.com/oaknational/Oak-Web-Application/pull/4430

Just deploying that PR other see if it works

CatMillerCM: I personally don't think it's worth it - it's not actually a feature that we need to dev around or visually see, and even if we were to temp roll back I think it would be fine to keep in the urls here as it will exist. I only ran this PR a bit earlier so I could check on our pal pa11y in case any tickets needed to be done to make that pass. Otherwise it's just more code to add and rip out, what do you think?

orangemug: Yeah also those are JS files so we'd need to do all the checking https://github.com/oaknational/Oak-Web-Application/pull/4430

I think it's more to manage as you need to remember that you have/haven't added but lots of changes, this'll do for now

orangemug (approved):

src/common-lib/urls/getDeploymentTestUrls.js
orangemug: Yeah also those are JS files so we'd need to do all the checking https://github.com/oaknational/Oak-Web-Application/pull/4430

I think it's more to manage as you need to remember that you have/haven't added but lots of changes, this'll do for now

src/common-lib/urls/getDeploymentTestUrls.js
```diff
@@ -14,6 +14,8 @@ function getDeploymentTestUrls() {
...
+    "/about-us/case-studies/tackling-blank-page-syndrome-with-aila",
```
alexcupit: this was the case study with the regenerated slug: 

```suggestion
    "/about-us/case-studies/giving-teachers-a-starting-point-with-safe-curriculum-aligned-ai-tools",
```

alexcupit (changes requested):

src/common-lib/urls/getDeploymentTestUrls.js
CatMillerCM: nice spot, thank you!

rightsaidjames (approved):

alexcupit (approved):

PR #4439, opened 2026-08-13
Feat/lesq 2185/navigation

couplm:
## Description

Music year: 1998

- Update teacher browse store to make programmeState optional
- Added teacher browse store to new areas where applicable
- Moved `browseRefined` to `unitRefined` and `programmeRefined` as applicable
- Moved `browseRefinedAccessed` and `curriculumRefinedAccessed` to `programmeAccessed` as applicable

## How to test

2. Click on \_\_\_\_\_\_

## Checklist

- [ ] Design sign-off


couplm:
@copilot please make a list of manual QA testing paths for devs/qa reviewing the navigation events

src/components/AppComponents/LayoutSiteFooter/LayoutSiteFooter.tsx
```diff
@@ -229,19 +230,15 @@ const FooterLink: FC<LayoutFooterLinkProps> = (props) => {
           .map(toSentenceCase)
           .join(" ");
 
-        if (keystageDescriptions.safeParse(sentenceCaseText).success) {
-          track.browseRefinedAccessed({
-            platform: "owa",
-            product: "teacher lesson resources",
-            engagementIntent: "refine",
+        if (
+          track?.programmeAccessed &&
```
Taha-Hassan-Git: What happens if this track function isn't there? When would it not be?

src/components/GenericPagesComponents/CurriculumTab/CurriculumTab.tsx
```diff
@@ -112,16 +114,21 @@ const CurriculumTab: FC<CurriculumDownloadTabProps> = ({
...
             </OakFlex>
             <OakBox
               $display={["none", "none", "block"]}
               $maxWidth={"spacing-640"}
             >
-              <SubjectPhasePicker
-                {...curriculumPhaseOptions}
-                id="choose-curriculum-label-large"
-              />
+              <TeacherBrowseAnalyticsStoreProvider
```
Taha-Hassan-Git, line 123: I don't think this page is accessible from anywhere in the website anymore, we can just get rid of it! Can either do it here, or add it to tech debt if you prefer.

src/components/GenericPagesComponents/TeachersTab/TeachersTab.tsx
```diff
@@ -118,10 +119,15 @@ function PhasePickerWithLegend({
       >
         Explore curriculum plans and teaching resources
       </OakHeading>
-      <SubjectPhasePicker
-        {...curriculumPhaseOptions}
-        id={"teachers-subject-picker-" + layout}
-      />
+      <TeacherBrowseAnalyticsStoreProvider
```
Taha-Hassan-Git: Did you try wrapping the homepage in this provider? The subject phase picker might be the only place we're sending events for now, but I think it'll be a bit more future-proof if we just wrap the page, so we can use the hook wherever without needing a refactor.

src/components/SharedComponents/CategoryFilterList/CategoryFilterList.tsx
```diff
@@ -33,21 +35,26 @@ const CategoryFilterList = <T extends CategoryLinkProps>(
...
-              isSelected={getIsSelected(category.linkProps)}
-              setSelected={setSelected}
-              trackingProps={trackingProps}
-              {...category}
-            />
-          );
-        })}
-      </OakUL>
-    </nav>
+    <TeacherBrowseAnalyticsStoreProvider
```
Taha-Hassan-Git: Again I think page-level wrapping is a bit nicer. Unless there's some reason I'm overlooking.

src/components/TeacherViews/MyLibrary/MyLibrary.tsx
```diff
@@ -103,21 +107,14 @@ export default function MyLibrary(props: Readonly<MyLibraryProps>) {
                     subject_categories: collection.subjectCategoryQuery,
                   },
                 })}
-                trackBrowseRefined={() =>
-                  track.browseRefined({
-                    platform: "owa",
-                    product: "teacher lesson resources",
-                    engagementIntent: "refine",
+                trackProgrammeRefined={() => {
```
Taha-Hassan-Git: Instead of prop drilling here, we could just call the hook where we need to use the tracking function

src/context/TeacherBrowseAnalytics/TeacherBrowseAnalyticsProvider.tsx
```diff
@@ -82,6 +83,10 @@ export const TeacherBrowseAnalyticsStoreProvider = ({
     store.setState({ journeyId });
   }, [store, journeyId]);
 
+  useEffect(() => {
```
Taha-Hassan-Git: Seems like we're tracking programme state in the journeyId and again here. Is there a way to do this more efficiently? We could at least have them in the same useEffect, or track both on one piece of state?

src/context/TeacherBrowseAnalytics/TeacherBrowseAnalyticsProvider.tsx
```diff
@@ -103,3 +108,30 @@ export const useTeacherBrowseAnalytics = <T,>(
...
 };
+
+export const useTeacherBrowseAnalyticsOptional = <T,>(
+  selector: (store: TeacherBrowseAnalyticsStore) => T,
+): T | undefined => {
+  const teacherBrowseAnalyticsStoreContext = useContext(
+    TeacherBrowseAnalyticsStoreContext,
+  );
+
+  // Stable fallback so we can call `useStore` unconditionally (Rules of Hooks).
```
Taha-Hassan-Git: Not sure about this pattern. What's the thinking here? Maybe better to throw and report an error?

src/context/TeacherBrowseAnalytics/TeacherBrowseAnalyticsStore.ts
```diff
@@ -292,181 +481,122 @@ export const createTeacherBrowseAnalyticsStore = (
...
-      lessonMediaClipsStarted: (data) => {
+      teachingMaterialsSelected: (data) => {
         const { avo, programmeState, journeyId } = get();
 
-        if (programmeState.browseLevel !== "lesson") {
-          reportAnalyticsError({
-            event: "lessonMediaClipsStarted",
-            programmeState,
-          });
+        const lessonState = requireLessonState(
```
Taha-Hassan-Git, line 657: Is there any harm in still checking the browseLevel? Unlikely we would have a lesson state without being at a lesson browse level, but stranger things have happened!

src/context/TeacherBrowseAnalytics/TeacherBrowseAnalyticsStore.test.ts
```diff
@@ -0,0 +1,280 @@
...
+        filterType: FilterType.CONTENT_TYPE_FILTER,
+        filterValue: "units",
+        activeFilters,
+        unitName: "Cells",
+        unitSlug: "cells",
+      }),
+    );
+  });
+
+  test("unitRefined on programme-level state omits unit analytics and still tracks", () => {
```
Taha-Hassan-Git: When would we fire unit refined without unit state? 🤔

src/context/TeacherBrowseAnalytics/TeacherBrowseAnalyticsProvider.tsx
```diff
@@ -103,3 +108,30 @@ export const useTeacherBrowseAnalytics = <T,>(
 
   return useStore(teacherBrowseAnalyticsStoreContext, selector);
 };
+
+export const useTeacherBrowseAnalyticsOptional = <T,>(
```
njblore: Not sure I understand the reasoning for the optional version, won't this mean we won't track events on pages that don't have the provider?

src/components/AppComponents/LayoutSiteFooter/LayoutSiteFooter.tsx
```diff
@@ -229,19 +230,15 @@ const FooterLink: FC<LayoutFooterLinkProps> = (props) => {
...
+          keystageDescriptions.safeParse(sentenceCaseText).success
+        ) {
+          track.programmeAccessed({
```
njblore: is this ever reached? I don't think we have keystage links on the footer anymore, we might just be able to remove this

Taha-Hassan-Git (commented):
Hiya! Great work. I have a few questions and some suggested changes. This is only after a dev review, I haven't tested the actual events yet.

src/components/GenericPagesComponents/CurriculumTab/CurriculumTab.tsx
Taha-Hassan-Git: This will also require some redirects work, so don't worry about it for now. I can add it to tech debt.

src/context/TeacherBrowseAnalytics/teacherBrowseAnalytics.types.ts
```diff
@@ -125,7 +125,9 @@ export type VideoTrackingProperties = {
 // All Track Fns used in the teacher browse journey
 export type TeacherBrowseTrackFns = {
```
njblore, line 129: are we able to use this type in the store now? or are there still events missing?

src/components/TeacherViews/MyLibrary/MyLibrary.tsx
njblore: is this one that got missed previously? 👀 would be good to remove `useAnalytics` in this file fully

src/context/TeacherBrowseAnalytics/TeacherBrowseAnalyticsProvider.tsx
couplm: It was meant to allow the component to work regardless but I think I misunderstood the requirements so I will swap this out :)

src/context/TeacherBrowseAnalytics/TeacherBrowseAnalyticsProvider.tsx
couplm: It was copilot saying this way would work better so i committed its suggestion, but I am removing this anyway :)

src/context/TeacherBrowseAnalytics/TeacherBrowseAnalyticsStore.test.ts
couplm: I have changed the test name to hopefully make more sense? I was trying to get better code coverage after sonar flagged it

src/context/TeacherBrowseAnalytics/teacherBrowseAnalytics.types.ts
couplm: They don't correctly align yet, it seems some are missing and some properties are misaligned. I need to sort mine out as well

src/components/GenericPagesComponents/CurriculumTab/CurriculumTab.tsx
couplm: Thank you, Taha!

src/components/SharedComponents/CategoryFilterList/CategoryFilterList.tsx
couplm: Nope no reason, my reasoning was simply not knowing the convention so thought it was safer to wrap a smaller rather than bigger area

couplm:
@copilot  please make an updated list of manual QA testing paths for devs/qa reviewing the navigation events

src/components/SharedComponents/PostCategoryList/PostCategoryList.tsx
```diff
@@ -33,32 +34,37 @@ const PostCategoryList: FC<PostCategoryListProps> = (props) => {
...
-              ? resolveOakHref({ page: "blog-index" })
-              : resolveOakHref({ page: "webinar-index" })
-          }
-          iconName={"arrow-right"}
-          isTrailingIcon
-        >
-          {`Switch to ${page === "blog-index" ? "webinars" : "blogs"}`}
-        </OakTertiaryButton>
-      </OakFlex>
+      <TeacherBrowseAnalyticsStoreProvider
```
couplm, line 37: I placed this here as it was used across multiple components and I figured it was best to be in one place? But happy to move if another pattern is better

src/components/TeacherComponents/MyLibraryUnitCard/MyLibraryUnitCard.test.tsx
```diff
@@ -126,6 +142,30 @@ describe("MyLibraryUnitCard", () => {
     const lessonLink = screen.getByText("Lesson 0");
     const user = userEvent.setup();
     await user.click(lessonLink);
-    expect(mockTrackLessonAccessed).toHaveBeenCalledWith("lesson-0-published");
```
couplm, line 129: I'm not entirely sure how it was sent like this before and this event confused me since it seemed to be mainly/only used by the pupil side from a quick look at posthog, but what I did on the right seemed correct...

src/app/(core)/layout.tsx
```diff
@@ -26,7 +27,12 @@ export default async function CoreLayout({
         <TopNav {...topNavProps} />
         <SimulateErrorControls errorBoundaryLevel="root" />
         <main id="main">{children}</main>
-        <LayoutSiteFooter />
+        <TeacherBrowseAnalyticsStoreProvider
```
njblore, line 27: does the footer still need the provider?

src/components/AppComponents/TopNav/TeachersTopNavHamburger/TeachersTopNavHamburger.tsx
```diff
@@ -88,7 +89,12 @@ export function TeachersTopNavHamburger(props: Readonly<TeachersSubNavData>) {
           $height={"100%"}
           id="teachers-top-nav-hamburger"
         >
-          <HamburgerMenuContent {...props} hamburgerMenu={hamburgerMenu} />
+          <TeacherBrowseAnalyticsStoreProvider
+            programmeState={null}
```
njblore, line 93: I'm not 100% sure if it's a problem, but we're losing the context of the programme state here when the hamburger is accessed from a programme page, so the way we're measuring continuous journeys (using the `journeyId`) would count any navigation events as an end to the journey even if the user doesn't go through to  a new programme page.. I'm not sure of a good solution though :/ we might need to rethink how these pages are architected as the topnav is currently a level above where the programme data comes in so we can't just wrap it all in one provider right now...

couplm: Writing here for transparency/future reference. We decided to try constructing the journey ID using URL segments instead of programme state. Programme state is still used elsewhere but this approach allows us to build journey IDs that are the same across stores.

src/components/AppComponents/TopNav/TopNav.tsx
njblore: are we not updating these browseRefineds?

couplm: I think this is a separate event that is staying as is? It isn't on my list at least 😅

njblore: oooh yes sorry I thought it was `browseRefined` but it's `browseAccessed` 🤔  we have `browseRefinedAccessed` -> `programmeAccessed` in the doc, I'm thinking maybe this is the same thing?

couplm: ah so maybe this goes from browseAccessed to programmeAccessed?

njblore: yeah I think so, because otherwise we have this one browse event still hanging out without any of the new nav properties 😑 would be good to confirm with Jo/Ewan though if possible

couplm: I will check, thanks for raising it :)

couplm: I'm glad you noticed this, had a reply from Ewan agreeing that it makes sense to move over with an explanation as to why. I will implement this and the below is here for documentation.

> Previously browse described a series of screens where the only available action was to take another step along browse. This is different now as users can take action at each stage of the process. Therefore using this programmeAccessed, unitAccessed, unitSequenceAccessed and lessonAccessed feels much more descriptive of what the use is actually "achieving" at each stage. In PostHog we'll probably set up an "action" which groups together these two events (browseAccessed and programmeAccessed) if we ever need to look back at how journeys have been affected by introducing the change.

src/components/TeacherComponents/MyLibraryUnitCard/MyLibraryUnitCard.test.tsx
njblore: it looks like previously this was passed a function which handled calling the actual analytics based on a provided lessonSlug, whereas now it calls lessonAccessed itself so it includes all the props.

For cases like this you can use `objectContaining` to reduce the number of things we care about checking, eg:
```
 expect(mockTrackLessonAccessed).toHaveBeenCalledWith(
    expect.objectContaining({
        lessonSlug: "lesson-o-published
    })
 )
```

src/context/TeacherBrowseAnalytics/utils/getJourneySlugsContext.ts
```diff
@@ -0,0 +1,24 @@
...
+export type JourneySlugs = {
+  subjectSlug: string;
+  phaseSlug: string;
+};
+
+const useJourneySlugsContext = (): JourneySlugs => {
+  const pathname = usePathname();
+  const subjectPhaseSegment = pathname
+    ?.split("/")
+    .find((segment) => segment.includes("-"));
```
njblore: this feels a wee bit brittle, could we instead run `parseSubjectPhaseSlug` on each segment to find the first match against the whole url?

src/components/AppComponents/AppLayout/AppLayout.tsx
```diff
@@ -60,7 +61,12 @@ const AppLayout: FC<AppLayoutProps> = (props) => {
         {appVariant === "client-error" ? (
           <TopNavMinimal />
         ) : (
-          <TopNav {...topNavProps} />
+          <TeacherBrowseAnalyticsStoreProvider
```
njblore: just looking at how AppLayout is used on pages router pages, it seems to wrap every page so I'm wondering if we're better off wrapping AppLayout on each page in the analytics provider rather than wrapping individual components within, we don't have the same issue of the TopNav being at a higher level in pages router so that should be workable (I think), and reduce the number of places we need to add the provider. What do you think?

njblore (approved):
looks good! added a few comments but nothing blocking getting this to QA 🎏

src/components/AppComponents/AppLayout/AppLayout.tsx
couplm: Thanks for this, I've updated all the areas I could find and tried to test them.

I think it's now ready for merging the filter store work. Just commenting this publicly here for visibility again 😬

couplm:
@copilot please make a list of manual QA testing instructions, taking into account that this contains navigation analytics and filter store work.

RBrownsett:
@couplm just looking at the QA testing list point 5 - the Unit Overview Accessed event is not firing from my library. Should this be added to https://app.notion.com/p/oaknationalacademy/Blocked-I-want-updated-My-Library-events-3a626cc4e1b1805fb495c55730f9fb3c?source=copy_link or handled in this ticket

RBrownsett (approved):
Full testing details on QA notion ticket. Approved, the programmeAccessed and programmeEvents are firing correctly.

A follow on ticket has been created that will implement missing/broken events for unitOverAccessed (not always firing) and programmeAccessed from the unit page via the breadcrumb and view all units button. 

https://app.notion.com/p/oaknationalacademy/Update-programme-unit-and-lesson-navigation-events-3c926cc4e1b1806d9af0ff7ca470456c?source=copy_link

PR #4442, opened 2026-08-17
feat: turn on oaks impact feature flag

alexcupit:
## Description

Music year: 1967

Not sure if we want to control this through vercel env var or here in code? If vercel, will need to trigger a rebuild which I think only certain permissions allow for the deployed app, which is why I opted towards this

- Flips default oaks-impact feature bool to turn feature on

## Issue(s)

Fixes #ADOPT-2174

## How to test

2. Click on header/footer
3. You should see oak's impact and pages should load as expected

## Checklist

- [ ] Design sign-off

### DIFF
--- a/src/utils/featureFlagChecks/flags.tsx
+++ b/src/utils/featureFlagChecks/flags.tsx
@@ -1,6 +1,6 @@
 export const FLAGS = {
   get "oaks-impact"() {
-    return process.env.NEXT_PUBLIC_FORCE_FEATURE_FLAG_OAKS_IMPACT ?? "false";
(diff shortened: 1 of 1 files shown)


CatMillerCM (approved):
I think this is good for now, as also easy to revert back and redeploy, and ideally we'll rip it out anyway once we're confident everything is good! But let's double check @orangemug is also happy with this approach

rightsaidjames (approved):
LGTM!

orangemug:
> But let's double check @orangemug is also happy with this approach

Yup happy

orangemug (approved):

PR #4447, opened 2026-08-18
Feat/lesq 2156/filter store

Taha-Hassan-Git:
## Description

[Music year: 2005](https://www.youtube.com/watch?v=k85mRPqvMbE)

This PR moves programme filtering into a shared BrowseFilters context, with the URL as the source of truth for filter state. 

Changes

* Added a Zustand-backed `BrowseFiltersProvider` and `useBrowseFilters` hook to manage programme filters and analytics
* Moved curriculum filtering helpers into the new `BrowseFilters` context
* Added URL-backed storage that preserves unrelated query parameters and omits filters matching programme defaults
* Added support for keeping filter state in sync with browser back/forward navigation while preserving the server-resolved initial state
* Updated programme pages and filter components to consume the shared BrowseFilters state
* Removed the previous `useFilters` hook
* Added tests for the provider, store, URL storage, analytics behaviour, and query-string filter helpers

## Issue(s)

Fixes #

## How to test

2. Click on \_\_\_\_\_\_

## Screenshots

{screenshots}

How it should now look:
{screenshots}

## Checklist

- [ ] Design sign-off


src/utils/curriculum/by-pathway.ts
```diff
@@ -1,7 +1,8 @@
+import { filteringFromYears } from "../../context/BrowseFilters/filtering";
```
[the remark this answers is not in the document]
Taha-Hassan-Git: Thanks!

src/context/BrowseFilters/browseFiltersUrlStorage.ts
```diff
@@ -0,0 +1,70 @@
...
+  filters: Partial<CurriculumFilters>;
+};
+
+const replaceSearch = (search: string) => {
+  const url = search
+    ? `${window.location.pathname}?${search}`
+    : window.location.pathname;
+
+  globalThis.history.replaceState(null, "", url);
+};
```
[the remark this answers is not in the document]
Taha-Hassan-Git: Thanks!

src/context/BrowseFilters/BrowseFiltersStore.ts
```diff
@@ -0,0 +1,99 @@
...
+  // The storage needs the *current* default filter, which lives in the store
+  // it is being built for, so it reads back through this reference.
+  const storeRef: { current?: StoreApi<BrowseFiltersStore> } = {};
+
+  const storage = createBrowseFiltersUrlStorage(
+    () => storeRef.current?.getState().defaultFilter ?? defaultFilter,
+  );
+
+  const store = createStore<BrowseFiltersStore>()(
+    persist<BrowseFiltersStore, [], [], BrowseFiltersPersistedState>(
```
njblore, line 49: why are we persisting the store? aren't they essentially persisted automatically in query params?

src/app/(core)/teachers/programmes/[slug]/[tab]/Components/ProgrammeView.tsx
```diff
@@ -92,34 +84,7 @@ export const ProgrammeView = ({
...
-    }
-
-    programmeRefined({
-      componentType: "filter_link",
-      activeFilters: newFilters,
-      filterType,
-      filterValue,
-    });
-  };
+  const { filters, onChangeFilters } = useBrowseFilters();
```
Comment (njblore) [inline src/app/(core)/teachers/programmes/[slug]/[tab]/Components/ProgrammeView.tsx]: what do you think about moving these into the actual filter components, so we don't need to prop drill so much?

src/context/BrowseFilters/BrowseFiltersProvider.tsx
```diff
@@ -0,0 +1,64 @@
...
+  defaultFilter,
+  initialFilter,
+  children,
+}: BrowseFiltersProviderProps) => {
+  const searchParams = useSearchParams();
+  const [store] = useState(() =>
+    createBrowseFiltersStore({ defaultFilter, initialFilter }),
+  );
+
+  // Keeps the store in sync with back/forward navigation.
```
njblore: I'm probably being dumb but why is this necessary? won't it always be updated from the default filter state anyways?

src/context/BrowseFilters/browseFiltersUrlStorage.ts
```diff
@@ -0,0 +1,70 @@
...
+
+      replaceSearch(
+        searchStringWithFilters(
+          window.location.search,
+          value.state.filters as CurriculumFilters,
+          getDefaultFilter(),
+        ),
+      );
+    },
+    removeItem: () => {
```
njblore: should `removeItem` accept a key param? looks like it currently removes everything, is that right?

njblore: same for `getItem` as well, if the behaviour is what we want the names are at least unclear

src/context/BrowseFilters/BrowseFiltersStore.ts
njblore: ok I read down and I think I see that this is necessary to tie the url state in, is that right?

Taha-Hassan-Git: Yeah!

src/app/(core)/teachers/programmes/[slug]/[tab]/Components/ProgrammeView.tsx
Comment (Taha-Hassan-Git) [reply src/app/(core)/teachers/programmes/[slug]/[tab]/Components/ProgrammeView.tsx]: Oh yeah! Of course!

src/context/BrowseFilters/BrowseFiltersProvider.tsx
Taha-Hassan-Git: Yeah, it isn't necessary! I was thinking ahead too much.

src/context/BrowseFilters/browseFiltersUrlStorage.ts
Taha-Hassan-Git: It does look a bit strange, it's because we're using persist which has named "slots" for each bit of storage, we're only using a single storage slot, having a separate one for each filter would be way overkill. We don't really need granular control over the keys here, we just want to read them from the url and write them to the url. 

The named `getItem` and `setItem` methods are what zustand's persist system expects, so we can't call them something else. However I can make some named functions that these methods call that will make the intention a bit more clear? Perhaps with additional comments?

I think this part of the filter store will make a lot more sense when we want to persist filters deeper into the journey!

njblore: Ooh ok gotcha, yeah if we can add a bit of commentary to clarify that would probably help

njblore (approved):

njblore:
@copilot based on the changes in this PR can you write a concise test plan for a QA to review with

RBrownsett:
@Taha-Hassan-Git looks like there is a small regression that only affects KS4 sciences. For example if you are on AQA Chemistry then select AQA Biology in the top nav, the url updates but the page does not refresh. See video example below:

https://github.com/user-attachments/assets/24d82044-c0c7-4e66-b353-b07d83b526a7

RBrownsett (approved):
Regression tested filters and events, looks good to me. Approved

PR #4448, opened 2026-08-20
Added `useCookieFlag()` to store flags and hooked up to the implementation toolkit dismissible banner

orangemug:
## Description
Added `useCookieFlag()` to store flags that are available in the backend & frontend code

> [!NOTE]
> This shouldn't be used for anything else except flags (true/false) because cookies get sent on each request 

I've hook up `useCookieFlag()` to store the state of the dismissible inline banner for implementation guides.

When dismissing the inline banner you should see a cookie gets added `oak-flag-toolkit-modal-dismissed=1` you can remove this when testing.

Also worth noting that it has a "broadcast channel" that means that the state will be updated on any other open tabs. 

## Issue(s)
Fixes `ADOPT-2208`

## How to test
Enable `implementation-guides` feature flag

1. Go to https://oak-web-application-website-git-feat-adopt-2208-cookie-f-bee6c1.vercel.thenational.academy/teachers/programmes/english-primary/units
2. Dismiss the inline banner
3. Hard-reload
4. Banner should not appear

## Checklist

- [ ] Design sign-off


src/hooks/useCookieFlag/getCookieFlags.ts
```diff
@@ -0,0 +1,10 @@
+import { cookies } from "next/headers";
+
+export async function getCookieFlags() {
```
alexcupit: perhaps `getActiveCookieFlags`?

src/config/flags.ts
alexcupit, line 1: not to be confused with feature flag config. Perhaps feature flags could also live in this file to make the distinction clear? Or this file could just be `cookieFlagNames.ts`?

src/app/(core)/teachers/programmes/[slug]/[tab]/Components/ImplementationGuideCallout/index.tsx
```diff
@@ -8,32 +8,49 @@ import {
...
   subject,
   phase,
   subjectTitle,
   phaseTitle,
+  activeFlags,
 }: Readonly<ImplementationGuideCalloutProps>) {
+  const [bannerDismissed, setBannerDismissed] = useCookieFlag(
+    "toolkit-modal-dismissed",
+    {
+      flags: cookieFlags,
```
Comment (alexcupit) [inline src/app/(core)/teachers/programmes/[slug]/[tab]/Components/ImplementationGuideCallout/index.tsx]: perhaps we could rename the param as below, or even `cookieFlagNames`? 
```suggestion
     cookieFlags,
```

src/hooks/useCookieFlag/useCookieFlag.ts
```diff
@@ -0,0 +1,109 @@
...
+  {
+    activeFlags,
+    cookieStore = isClient ? window.cookieStore : undefined,
+    flags,
+  }: Opts,
+): [boolean, SetValue<boolean>] {
+  const key = deriveKey(rawKey);
+
+  // Cleanup any cookies that are not in the flags list.
+  const fn = useCallback(async () => {
```
alexcupit: Perhaps?
```suggestion
  const cleanupOldCookieFlags = useCallback(async () => {
```

src/hooks/useCookieFlag/useCookieFlag.ts
```diff
@@ -0,0 +1,109 @@
...
+    if (cookieStore && isClient) {
+      const currentCookies = await (cookieStore instanceof CookieStore
+        ? cookieStore.getAll()
+        : cookieStore.getAll(""));
+
+      for (const { name } of currentCookies) {
+        if (
+          name?.startsWith(PREFIX) &&
+          !flags.includes(name.replace(PREFIX + "-", ""))
```
alexcupit: are these the same? 
```suggestion
          !flags.includes(rawKey)
```

src/hooks/useCookieFlag/useCookieFlag.ts
orangemug: Updated in latest

src/hooks/useCookieFlag/useCookieFlag.ts
orangemug: Updated in latest

src/hooks/useCookieFlag/getCookieFlags.ts
orangemug: Updated in latest

src/config/flags.ts
orangemug: Updated in latest

src/hooks/useCookieFlag/useCookieFlag.ts
```diff
@@ -0,0 +1,105 @@
...
+  }, [cookieStore, key]);
+
+  const setValue = useCallback(
+    async (value: SetStateAction<boolean>) => {
+      if (!cookieStore) throw new Error("cookieStore is not available");
+      const newValue =
+        typeof value === "function" ? value(currentValueLocal) : value;
+
+      if (newValue) {
+        await cookieStore.set(key, "1");
```
orangemug: Set this to 400 days which is the longest, we could reset this _onload_ but feels overkill for now 

```
maxAge: 60 * 60 * 24 * 400, // 400 days
```

rightsaidjames: I think 400 days is too long - that would mean the banner, once dismissed, would likely never be seen again unless cookies were cleared or a different device was used - unless the banner lasted well over a year. I think we originally discussed 60 days as a starting point, but we can make a team decision on this.

orangemug: Yeah I think that's what we want from the ticket

> We could even set a short-ish expiry date (60 or 90 days?) so it would effectively clean up after itself.

And 

> Good point on expiry, guess if we set the max-age to 60 days then updated the time each time we check it, it would mean that it would never actually expire for regular users. We could then add proper cleanup down the road

But just decided to implement proper cleanup instead

orangemug: Updated to 400 days

src/hooks/useCookieFlag/README.md
```diff
@@ -0,0 +1,35 @@
+# useCookieFlag
+We want a way to set flags for a user on the server & client. This is for things like when a popup is dismissed
+
+Aims
```
alexcupit: nit: does this want to be a heading? 
```suggestion
## Aims
```

src/hooks/useCookieFlag/README.md
alexcupit, line 1: should this also include a mention of adding to config/cookieFlagNames.ts?

alexcupit (approved):
looks good, just a couple of v minor points on the README but not essential to address now!

src/app/(core)/teachers/programmes/[slug]/[tab]/Components/ImplementationGuideCallout/ImplementationGuideCallout.test.tsx
```diff
@@ -61,4 +64,24 @@ describe("ImplementationGuideCallout", () => {
...
+        subject="rshe-pshe"
+        phase="ks2"
+        subjectTitle="RSHE (PSHE)"
+        phaseTitle="Primary"
+        onClick={() => {}}
+        activeFlags={["oak-flag-toolkit-modal-dismissed"]}
+      />,
+    );
+
+    // Check if the message is rendered correctly
```
Comment (CatMillerCM) [inline src/app/(core)/teachers/programmes/[slug]/[tab]/Components/ImplementationGuideCallout/ImplementationGuideCallout.test.tsx]:
```suggestion
```

src/app/(core)/teachers/programmes/[slug]/[tab]/Components/ImplementationGuideCallout/ImplementationGuideCallout.test.tsx
```diff
@@ -61,4 +64,24 @@ describe("ImplementationGuideCallout", () => {
     );
     expect(message).toBeInTheDocument();
   });
+
+  test("disabled when flag present", () => {
```
Comment (CatMillerCM) [inline src/app/(core)/teachers/programmes/[slug]/[tab]/Components/ImplementationGuideCallout/ImplementationGuideCallout.test.tsx]: sorry bit of a nit but i often use tests to understand what's going on - am i right in thinking that the banner just doesnt render (early return) if the active flag is passed? could this be 'should not render when...', or does it still appear in the dom but disabled?

rightsaidjames (approved):
LGTM - cookie is set when banner is dismissed, returns if cookie is deleted. In incognito mode the banner doesn't appear because it depends on the feature flag.

If the feature flag is enabled but cookies are rejected, the cookie is still set (as it should be, as it's a functional cookie)

Expiry date is as expected.

src/app/(core)/teachers/programmes/[slug]/[tab]/Components/ImplementationGuideCallout/ImplementationGuideCallout.test.tsx
Comment (orangemug) [reply src/app/(core)/teachers/programmes/[slug]/[tab]/Components/ImplementationGuideCallout/ImplementationGuideCallout.test.tsx]: Fixed in latest

src/app/(core)/teachers/programmes/[slug]/[tab]/Components/ImplementationGuideCallout/ImplementationGuideCallout.test.tsx
Comment (orangemug) [reply src/app/(core)/teachers/programmes/[slug]/[tab]/Components/ImplementationGuideCallout/ImplementationGuideCallout.test.tsx]: Removed

src/hooks/useCookieFlag/README.md
orangemug: Added in latest and updated `README.md` in general

src/hooks/useCookieFlag/useCookieFlag.ts
```diff
@@ -0,0 +1,112 @@
...
+  const [currentValueLocal, setCurrentValueLocal] = useState(() => {
+    return activeFlags.includes(key);
+  });
+
+  // Cleanup any cookies that are not in the flags list.
+  const cleanupOldCookieFlags = useCallback(async () => {
+    if (cookieStore && isClient) {
+      const currentCookies = await (cookieStore instanceof CookieStore
+        ? cookieStore.getAll()
+        : cookieStore.getAll(""));
```
CatMillerCM, line 47: how come we need to differentiate and check this here?

jest.setup.js
```diff
@@ -5,6 +5,8 @@ import "@testing-library/jest-dom";
 import "whatwg-fetch";
 import bugsnag from "@bugsnag/js";
 import { installMockIntersectionObserver } from "@oaknational/oak-components";
+import "./src/__tests__/__helpers__/mockBroadcastChannel";
+import "./src/__tests__/__helpers__/mockCookieStore";
```
CatMillerCM, line 9: do we have to import them here so they're globally available in every test?

CatMillerCM (approved):
just a couple of qus but looks-good-and-slightly-mind-boggling-to-me 😄

src/hooks/useCookieFlag/useCookieFlag.ts
orangemug: Next.js has a very slightly different usage I will post a PR to update add some comments

jest.setup.js
orangemug: I assume it's just easier to add general mocks globally that "should exist in the env". I _think_ jest is a pre-fork setup anyway so should be quicker also

PR #4454, opened 2026-08-21
feat(mcp): add the Oak Curriculum MCP landing page at /ai-plugin (replaces #4443)

mantagen:
## Description

Adds the Oak Curriculum MCP landing page at `/ai-plugin` — a static, teacher-facing page introducing Oak's MCP app, with a one-click prompt that opens Claude ready to install it.

- New route at `src/app/(core)/ai-plugin/page.tsx` (`force-static`, with page metadata)
- `McpView` composed from section components: hero, "See it in action", capabilities, assistant picker, feedback panel
- All copy in one constant (`mcpContent.ts`), so copy review doesn't mean reading through components
- Three SVGs under `public/images/mcp/`
- `/ai-plugin` added to `getDeploymentTestUrls`

The page is named `/ai-plugin` by product decision (Slack, 2026-09-04). That is explicitly **not its permanent home** — how this is communicated alongside Oak's other open curriculum services is a broader discussion still to be had.

Replaces #4443: the same page, rebuilt on a branch whose history never contained the third-party logo assets that one carried.

**`/mcp` is deliberately untouched by the rename** — `/mcp/carousel/*` is an external contract held by Anthropic's directory listing, guarded by `mcp-carousel-assets.test.ts` and a Playwright spec asserting no redirect. Measured 2026-09-07: all three carousel images serve 200 with zero redirects, while `/mcp` itself is a plain 404. Any future `/mcp` → `/ai-plugin` redirect must therefore match the exact path only, never a wildcard.

Follow-ups tracked separately: `MCP-684` (copy to Sanity, images to Cloudinary) and `MCP-685` (the rename's remaining surfaces, including the two `www.thenational.academy/mcp` links on the live privacy policy).

## Issue(s)

Relates to `MCP-94`. It does not close it — the AI-experiments nav dropdown is still outstanding there.

## How to test

1. Go to https://oak-web-application-website-git-feat-mcp-landing-page.vercel.thenational.academy/ai-plugin
2. The page should render in order: hero → "See it in action" → "What can you do?" → "Choose your AI assistant" → "Give feedback"
3. Click **Try in Claude** — opens `claude.ai` in a new tab with the install prompt pre-filled
4. Click **Share feedback** — opens a `mailto:` to support@thenational.academy
5. Check mobile and desktop widths

Claude is the only assistant offered at launch; the copy says ChatGPT and others are to follow.

## Screenshots

n/a — new page, see the preview URL above.

## Checklist

- [ ] Added or updated tests where appropriate
[...]

PR #4443, opened 2026-08-17
feat(mcp): add Oak Curriculum MCP landing page at /mcp

mantagen:
## Description

Music year: 1986

Implements the OWA Oak MCP landing page at `/mcp`, built against the final Figma designs and copy ([🌳 Oak MCP v1, frame `OWA OAK MCP landing page - thenational.academy/mcp >1280`](https://www.figma.com/design/t3tkWC64R8j21SnjSgJUSH/%F0%9F%8C%B3-Oak-MCP-v1?node-id=24394-42298)).

- New App Router page at `src/app/(core)/mcp/`, so it inherits `TopNav`, `<main>` and `LayoutSiteFooter` from the `(core)` layout. `dynamic = "force-static"`, wrapped in `withPageErrorHandling`
- Built from `@oaknational/oak-components` primitives and design tokens. The Figma file uses Oak's own token names (`heading-2`, `body-2`, `bg-decorative1-main`, `border-radius-xl`, spacing 1:1), so the mapping is direct
- All copy and hrefs live in `mcpContent.ts`, so copy review does not mean reading components
- Video placeholder occupies the designed slot — the final design still marks it `[Video to come]`
- `/mcp` added to `getDeploymentTestUrls` so pa11y (axe) and Percy cover it in CI

### ⚠️ Merge order dependency

This page now lives at `/mcp`, which is only free once the MCP server has moved to `mcp.thenational.academy`. At the time of writing `www.thenational.academy/mcp` still returns a 200 HTML page titled "Oak Curriculum MCP (HTTP)" from a non-OWA origin, and that exact URL is the server's OAuth protected-resource identifier (RFC 8707 audience) — so **do not merge until the DNS change has landed**, or the route will be shadowed. Preview deployments are unaffected, as they serve from a different host.

No redirect from `/teachers/mcp`: the page has never shipped, so nothing links to the old path.

### Page structure

Hero → "See it in action" → "What can you do?" → "Choose your AI assistant" → "Use it responsibly" → illustration → "How it works" (with "Questions or problems?" nested) → AI-output banner → "Give feedback". Section order is confirmed by the `375` frame.

### Not in the final design, so not built

- **ChatGPT.** The final hero and assistant section are Claude only, with the note "We're starting with Claude, and working to bring Oak to more AI assistants soon." The unused `chatgpt-logo.png` is deleted.
[...]

PR #4459, opened 2026-08-24
feat(nav): expand the AI experiments link into a dropdown

assadk88:
## Description

Music year: 1994

Expands the `AI experiments` link in the global nav into a dropdown, per the new Figma design: `Labs home` and `Aila` (both external, to Labs) plus an internal `Oak Curriculum MCP` link to `/mcp`. The MCP link is not gated, so this should merge after the `/mcp` landing page (`feat/mcp-landing-page`) is live.

## Issue(s)

Fixes #

[Figma](https://www.figma.com/design/t3tkWC64R8j21SnjSgJUSH/%F0%9F%8C%B3-Oak-MCP-v1?node-id=24394-39037&p=f&t=OI1X2xn0TaRRXZnn-0)

## How to test

1. Go to [OWA Preview URL from Vercel bot comment] as a teacher
2. On desktop widths, click `AI experiments` in the top nav
3. You should see a dropdown with `Labs home` and `Aila` (each with an external-link icon, opening Labs in a new tab) and `Oak Curriculum MCP` linking to `/mcp`
4. Keyboard: Tab to the `AI experiments` button, press Enter, Tab through the links, then Escape; focus should return to the button
5. On mobile widths, open the hamburger; `AI experiments` is now a submenu button; tap it to see the same three links, and the back button should return focus to it

>[!NOTE]
>
> This is to be merged in after the Oak MCP landing page work goes in

## Screenshots

How it used to look:

How it should now look:

## Checklist

- [ ] Design sign-off
- [ ] ~~Does this PR update a package with a breaking change~~ (no package changes)

### DIFF
--- a/src/__tests__/pages/__snapshots__/index.test.tsx.snap
+++ b/src/__tests__/pages/__snapshots__/index.test.tsx.snap
@@ -359,12 +359,12 @@ exports[`Teachers Page Page component renders 1`] = `
                       <div
                         class="sc-aXZVg kpXzs yellow-shadow"
                       />
-                      <a
-                        aria-label="Ai experiments (this will open in a new tab)"
+                      <button
+                        aria-expanded="false"
+                        aria-haspopup="true"
                         class="sc-dcJsrY sc-jXbUNg hodcpD kwUki internal-button"
-                        href="https://labs.thenational.academy"
-                        id="teachers-labs"
-                        target="_blank"
+                        data-testid="teachers-aiExperiments"
+                        id="teachers-aiExperiments"
                       >
                         <div
                           class="sc-aXZVg sc-gEvEer jPvJsJ kejpBe"
@@ -372,23 +372,10 @@ exports[`Teachers Page Page component renders 1`] = `
                           <span
                             class="sc-eqUAAy kJIzWq"
                           >
-                            Ai experiments
-                          </span>
-                          <span
-                            class="sc-aXZVg jJQFnr"
-                          >
-                            <img
-                              alt=""
-                              class="sc-iGgWBj lcrpYX"
-                              data-nimg="fill"
-                              decoding="async"
-                              loading="lazy"
-                              src="https://next_public_oak_assets_host/NEXT_PUBLIC_OAK_ASSETS_PATH/v1699953892/icons/hlxmejse3mcr4tqo6t8u.svg"
-                              style="position: absolute; height: 100%; width: 100%; left: 0px; top: 0px; right: 0px; bottom: 0px; color: transparent;"
-                            />
+                            AI experiments
                           </span>
                         </div>
-                      </a>
+                      </button>
                     </div>
                   </li>
                 </ul>
--- a/src/__tests__/pages/about-us/__snapshots__/get-involved.test.tsx.snap
+++ b/src/__tests__/pages/about-us/__snapshots__/get-involved.test.tsx.snap
@@ -97,19 +97,19 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   font-family: --var(google-font),Lexend,sans-serif;
 }
 
-.c38 {
+.c37 {
   position: relative;
   max-width: 11.25rem;
   display: none;
   font-family: --var(google-font),Lexend,sans-serif;
 }
 
-.c39 {
+.c38 {
   display: none;
   font-family: --var(google-font),Lexend,sans-serif;
 }
 
-.c40 {
+.c39 {
   position: relative;
   height: -webkit-fit-content;
   height: -moz-fit-content;
@@ -130,17 +130,17 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   letter-spacing: -0.005rem;
 }
 
-.c40:hover {
+.c39:hover {
   cursor: pointer;
 }
 
-.c44 {
+.c43 {
   top: 50%;
   right: 0.5rem;
   font-family: --var(google-font),Lexend,sans-serif;
 }
 
-.c45 {
+.c44 {
   position: relative;
   width: -webkit-max-content;
   width: -moz-max-content;
@@ -148,7 +148,7 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   font-family: --var(google-font),Lexend,sans-serif;
 }
 
-.c51 {
+.c50 {
   position: relative;
   width: 2rem;
   min-width: 2rem;
@@ -159,7 +159,7 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   font-family: --var(google-font),Lexend,sans-serif;
 }
 
-.c53 {
+.c52 {
   position: absolute;
   top: 0rem;
   width: 100%;
@@ -168,12 +168,12 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   font-family: --var(google-font),Lexend,sans-serif;
 }
 
-.c56 {
+.c55 {
   display: block;
   font-family: --var(google-font),Lexend,sans-serif;
 }
 
-.c59 {
+.c58 {
   width: -webkit-fit-content;
   width: -moz-fit-content;
   width: fit-content;
@@ -193,7 +193,7 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   letter-spacing: 0.0115rem;
 }
 
-.c61 {
+.c60 {
   position: relative;
   width: 1.5rem;
   min-width: 1.5rem;
@@ -203,12 +203,12 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   font-family: --var(google-font),Lexend,sans-serif;
 }
 
-.c62 {
+.c61 {
   display: none;
   font-family: --var(google-font),Lexend,sans-serif;
 }
 
-.c65 {
+.c64 {
   display: block;
   font-family: --var(google-font),Lexend,sans-serif;
 }
@@ -732,7 +732,7 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   gap: 0.25rem;
 }
 
-.c37 {
+.c36 {
   display: -webkit-box;
   display: -webkit-flex;
   display: -ms-flexbox;
@@ -744,7 +744,7 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   gap: 1.5rem;
 }
 
-.c41 {
+.c40 {
   display: -webkit-box;
   display: -webkit-flex;
   display: -ms-flexbox;
@@ -755,14 +755,14 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   align-items: center;
 }
 
-.c46 {
+.c45 {
   display: -webkit-box;
   display: -webkit-flex;
   display: -ms-flexbox;
   display: flex;
 }
 
-.c50 {
+.c49 {
   display: -webkit-box;
   display: -webkit-flex;
   display: -ms-flexbox;
@@ -781,7 +781,7 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   gap: 0rem;
 }
 
-.c52 {
+.c51 {
   display: -webkit-box;
   display: -webkit-flex;
   display: -ms-flexbox;
@@ -796,7 +796,7 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   justify-content: center;
 }
 
-.c60 {
+.c59 {
   display: -webkit-box;
   display: -webkit-flex;
   display: -ms-flexbox;
@@ -1189,7 +1189,7 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   text-align: left;
 }
 
-.c55 {
+.c54 {
   font-family: --var(google-font),Lexend,sans-serif;
   font-weight: 600;
   font-size: 1rem;
@@ -1367,7 +1367,7 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   cursor: default;
 }
 
-.c48 {
+.c47 {
   background: none;
   color: inherit;
   border: none;
@@ -1381,12 +1381,12 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   color: #222222;
 }
 
-.c48:disabled {
+.c47:disabled {
   pointer-events: none;
   cursor: default;
 }
 
-.c57 {
+.c56 {
   background: none;
   color: inherit;
   border: none;
@@ -1399,12 +1399,12 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   font-family: --var(google-font),Lexend,sans-serif;
 }
 
-.c57:disabled {
+.c56:disabled {
   pointer-events: none;
   cursor: default;
 }
 
-.c63 {
+.c62 {
   background: none;
   color: inherit;
   border: none;
@@ -1434,12 +1434,12 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   border-radius: 0.25rem;
 }
 
-.c63:disabled {
+.c62:disabled {
   pointer-events: none;
   cursor: default;
 }
 
-.c66 {
+.c65 {
   background: none;
   color: inherit;
   border: none;
@@ -1463,7 +1463,7 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   border-style: none;
 }
 
-.c66:disabled {
+.c65:disabled {
   pointer-events: none;
   cursor: default;
 }
@@ -1529,15 +1529,15 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   object-fit: contain;
 }
 
-.c36 {
-  -webkit-filter: invert(10%) sepia(1%) saturate(236%) hue-rotate(314deg) brightness(95%) contrast(91%);
-  filter: invert(10%) sepia(1%) saturate(236%) hue-rotate(314deg) brightness(95%) contrast(91%);
+.c53 {
+  -webkit-filter: invert(98%) sepia(98%) saturate(0%) hue-rotate(328deg) brightness(102%) contrast(102%);
+  filter: invert(98%) sepia(98%) saturate(0%) hue-rotate(328deg) brightness(102%) contrast(102%);
   object-fit: contain;
 }
 
-.c54 {
-  -webkit-filter: invert(98%) sepia(98%) saturate(0%) hue-rotate(328deg) brightness(102%) contrast(102%);
-  filter: invert(98%) sepia(98%) saturate(0%) hue-rotate(328deg) brightness(102%) contrast(102%);
+.c66 {
+  -webkit-filter: invert(10%) sepia(1%) saturate(236%) hue-rotate(314deg) brightness(95%) contrast(91%);
+  filter: invert(10%) sepia(1%) saturate(236%) hue-rotate(314deg) brightness(95%) contrast(91%);
   object-fit: contain;
 }
 
@@ -1685,13 +1685,13 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   display: none;
 }
 
-.c64 {
+.c63 {
   position: relative;
   width: 100%;
   height: 100%;
 }
 
-.c64:hover {
+.c63:hover {
   -webkit-text-decoration: none;
   text-decoration: none;
   color: #ffffff;
@@ -1699,27 +1699,27 @@ exports[`pages/about/get-involved.tsx renders 1`] = `
   border-color: #575757;
 }
 
-.c64:hover [data-state="selected"] {
+.c63:hover [data-state="selected"] {
   display: none;
 }
 
-.c64:active {
+.c63:active {
   background: #222222;
   border-color: #222222;
   color: #ffffff;
 }
 
-.c64:active [data-state="selected"] {
+.c63:active [data-state="selected"] {
   display: none;
 }
 
-.c64:disabled {
+.c63:disabled {
   background: #808080;
   border-color: #808080;
   color: #ffffff;
 }
 
-.c64:disabled [data-state="selected"] {
+.c63:disabled [data-state="selected"] {
   display: none;
 }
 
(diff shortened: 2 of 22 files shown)


mantagen (commented):
> ### 🤖 Agent review, posted on Matt Gregory's behalf
>
> This was written by a Claude Code agent working with Matt, and posted under his
> account because our bot identity has no write access to this repo (its app
> installation covers exactly one repo, `oak-open-curriculum-ecosystem`). **Treat it
> as an agent's review, not as Matt's own sign-off** — he has not read every line of
> it. Push back on anything that looks wrong; where I could not measure something I
> have said so explicitly rather than guessed.

## Verdict

**Good PR.** Two specialist reviewers (code quality, accessibility) plus my own checks
found **no defect that renders wrongly today**, and no WCAG 2.2 AA regression introduced.
The slug rename is internally consistent across query, fixture and components; the dead
`MainMenuLink` was removed cleanly with all three imports it alone justified; and the new
`TopNavDropdown` test is the strongest test in the change.

The blocker is outside the diff, and everything else is either a small pre-merge tidy or
a latent trap for the next editor.

---

## 🔴 Blocker: this ships a nav link to a live 404

The dropdown links `/mcp`. Measured just now:

```
GET https://www.thenational.academy/mcp          → 404
GET https://www.thenational.academy/mcp/healthz  → 404
GET https://mcp.thenational.academy/mcp/healthz  → 200   (control: the MCP server itself is healthy)
```

Your description already calls this dependency correctly — *"this should merge after the
`/mcp` landing page is live"* — so this is confirmation, not a new finding. Two things
worth knowing:

- **#4454** ("add the Oak Curriculum MCP landing page at /mcp") is the PR that creates
  `src/app/(core)/mcp/page.tsx`. It is still **open**, last commit 21 August.
- `/mcp` isn't merely un-launched, it is **actively dead**. The Cloudflare rule that used
  to route `www/…/mcp` to the standalone MCP app was removed on 21 August
  (`oaknational/cloud-config#561`), which was the right call — but the replacement OWA
  page has been sitting in #4454 ever since. So there has been no page at `/mcp` for
  eleven days, and this PR would advertise it in the global nav.

**Also**: **#4443** is still open even though #4454's own title says it replaces it. Worth
closing to keep the queue honest.

Ordering is otherwise correct and you have it the right way round: #4454 owns the route,
this PR owns the `urls.ts` entry, and merging in that order is safe.

---

## Before merge

| | Item |
|---|---|
| 1 | **PR is `CONFLICTING`** with `main` — needs a rebase regardless of everything else. |
| 2 | **SonarCloud quality gate is failing** on 1 new issue, and nobody has read it. |
| 3 | **37 Percy diffs need a real approval, not a wave-through** — see the scope item below. |
| 4 | **Update the description** to say the change also restyles the *About us* and *Guidance* dropdowns (item S1). |
| 5 | **One test line to swap** (item A1). |
| 6 | **One description claim is not true** (item A2). |

---

## Scope

### S1 · The `width` change restyles two dropdowns that aren't in scope, and has zero test footprint

`TopNavDropdown.tsx:372`, `width={"spacing-160"}` → `width={"max-content"}`.

`TeachersLinksSection` is shared by **Guidance (4 links), About us (5–6), and AI
experiments (3)** — so this restyles 12–13 link buttons across three dropdowns on every
page that renders the nav. That is almost certainly your 37 Percy diffs.

I read this as an *unavoidable consequence* rather than gratuitous scope — "Oak Curriculum
MCP" is the longest label the section has ever carried, and the old fixed width would have
wrapped or clipped it. You also correctly left `renderNavButtons` (line 216) on
`spacing-160`, so phase/keystage buttons are untouched. Good instinct.

Two things follow:

- **The Percy diffs are load-bearing.** We censused all 1,933 changed snapshot lines: the
  semantic content is one edit repeated eight times (the `<a>`→`<button>` swap and
  `"Ai experiments"`→`"AI experiments"`), and **everything else is styled-components class
  renumbering** caused by the removed `` shifting the hash sequence. No `width`
  declaration differs in any snapshot, and the dropdown link sections don't appear in the
  page snapshots at all. So Percy is the *only* evidence this change looks right, in either
  the new dropdown or the two collateral ones.
- **Worth one line of thought**: `$minWidth={"spacing-160"}` alongside `width={"max-content"}`
  would hold the existing two dropdowns exactly as they are and widen only where the label
  demands it — cutting the visual surface to the section you actually changed. Your call.
- Minor: this swaps a design **token** for a raw CSS keyword on a shared component. Both
  typecheck against the same prop, so it's legal, but it's a design-system question worth a
  second's thought.

---

## Accessibility

Full WCAG 2.2 AA / ARIA 1.2–1.3 / APG pass. **Nothing AA-blocking is introduced.** One
line to change, one description claim to fix, two pre-existing items for separate tickets.

### A1 · Please swap one test assertion (in this PR)

`SubNav.test.tsx` adds:

```ts
expect(aiExperimentsButton).toHaveAttribute("aria-haspopup", "true");
```

Per ARIA 1.2 and 1.3 (identical normative text), `aria-haspopup="true"` **must** be treated
as `"menu"`, and:

> "Authors MUST ensure that the role of the element that serves as the container for the
> popup content is `menu`, `listbox`, `tree`, `grid`, or `dialog`, and that the value of
> `aria-haspopup` matches the role of the popup container."

The container here is an `OakGrid as="ul"` (`TopNavDropdown.tsx:339-348`) — role `list`. And
the traversal is Tab (`focusManager.handleTabKeyDown`, line 375), with no Arrow handlers
inside the section, whereas APG's menu pattern states *"Tab and Shift + Tab do not move
focus among the items in the menu."* APG's own **Disclosure Navigation Menu** example — the
canonical pattern for exactly this UI — says:

> "This implementation of site navigation does not use the menu role because it does not
> provide the complex functionality that assistive technologies expect in a widget that has
> the menu role."

and uses `aria-expanded` alone; `aria-haspopup` appears **zero times** on that page.

**This is pre-existing, not yours** — `SubNav.tsx` isn't in your diff at all. The attribute
already sits on Primary, Secondary, Guidance and About us via the shared `getButtonProps`.
Your change just enrols a fifth item.

**The ask is only this**: the new assertion converts a pre-existing spec violation into an
asserted contract, so whoever fixes it later has to delete a test too. Please use the
attribute the Disclosure pattern actually requires — the snapshot confirms it's already on
the rendered element:

```ts
expect(aiExperimentsButton).toHaveAttribute("aria-expanded", "false");
```

### A2 · The Escape claim in the description isn't true for this nav — verified

Step 4 says *"press Escape; focus should return to the button."* I checked, because the
method does exist and it would be easy to assume it's wired:

- `DropdownFocusManager.ts:301-314` defines `handleEscapeKey` — closes the menu, focuses the
  ancestor node. Correct implementation.
- Swept all 23 `.ts`/`.tsx` files under `components/AppComponents/TopNav/` (control probe:
  the search correctly finds the definition and `isDropdownMenuItem`'s two real consumers).
  `handleEscapeKey` has exactly **two** occurrences: its definition, and **one call site —
  `TopNavMinimal.tsx:45-50`**, a different component.
- It is **not wired in `SubNav.tsx`, `TopNav.tsx`, or `TopNavDropdown.tsx`**. Those wire only
  `handleTabKeyDown` and `handleArrowKeyDown`.
- `DropdownFocusManager.test.ts` contains **zero** Escape coverage.

So on the main teachers nav, Escape does nothing. This is **pre-existing** (Guidance and
About us are identical) and **does not block** — but please either drop that step from the
description or, if you fancy it, wire `handleEscapeKey` in `SubNav`, which would satisfy
APG's *"closes it and sets focus on the button that controls that dropdown"* for all five
triggers at once.

**The hamburger half of your claim is true, and provably so**: `HamburgerMainMenu.tsx:27-43`
computes `returnFocusId = prevSubmenu.value + "button"` → `"AI experimentsbutton"`, and
`MainMenuButton` sets `id={title + "button"}` (line 202) → exact match. Nice. It just isn't
covered by a test either — the new hamburger test clicks in, asserts the three links, and
stops.

### A3 · `max-content` — assessed, not a failure

- **1.4.10 Reflow / 1.4.4 Resize Text: no failure.** The desktop dropdown is
  `$display={["none","none","flex"]}` (`SubNav.tsx:136`) and the hamburger is
  `["block","block","none"]` — mutually exclusive. At 1280px/400% zoom (the 320 CSS px
  condition) the desktop nav is `display: none` and the hamburger path, which doesn't use
  `TeachersLinksSection`, is what a user gets.
- **2.5.8 Target Size: no failure identified**, with one caveat. `width` is horizontal only
  and button height is untouched. The change *shrinks* short labels — "Aila" and Guidance's
  "Help" go from a fixed box to intrinsic width. Four characters plus padding very likely
  clears 24px, but "very likely" isn't measured, and `$rg="spacing-8"` gives only 8px
  vertical gap so the Spacing exception wouldn't rescue it. If you're already opening the
  dropdowns for Percy, a `getBoundingClientRect()` on "Aila" and "Help" would settle it.
- **Non-normative but worth a look**: the grid is `1fr 1fr 1fr` with `$cg="spacing-40"`. A
  `max-content` child in a `1fr` column *overflows* the column rather than wrapping when its
  intrinsic width exceeds the fraction. "Oak Curriculum MCP" + trailing icon is the longest
  label in the set, so it's the one to eyeball against the adjacent column.

### A4 · Confirmed fine — nothing was lost in the hamburger

The deleted `MainMenuLink` carried `iconName="external"`, `target="_blank"` and the
new-tab aria-label. All three now sit on the control that actually navigates
(`HamburgerSubMenu.tsx:127-145`), and your new test asserts exactly that — icon, name
suffix, and `target="_blank"` for Labs home and Aila, `_self` and no suffix for MCP.
The replacement button correctly carries `chevron-right` instead of an external icon,
because it no longer navigates. Right call, and the test rename is legitimate: the old
test's subject genuinely ceased to exist.

### A5 · Settled — `MaybeVisuallyHidden` is not a focus trap

Worth recording since it looked like a risk: `TopNav.tsx:41-68` returns `null` for
non-selected sections once `afterInitialRender` is true, with the comment *"Fully remove the
visually hidden elements from the page when the client initialises to prevent issues with
focus and screen readers."* So no invisible-but-focusable links after hydration. **Non-issue.**
(Pre-hydration they are in the DOM offscreen, which is a deliberate no-JS/SEO trade.)

### A6 · Accessible-name inconsistency — not a violation, one-word fix if you want it

Three call sites, all in the nav:

| Site | String |
|---|---|
| `TopNavDropdown.tsx:369` | `(opens in a new tab)` |
| `HamburgerSubMenu.tsx:136` | `(this will open in a new tab)` |
| `SubNav.tsx:68` | `(this will open in a new tab)` |

The prevailing convention is **"(this will open in a new tab)"** (2 of 3 surviving sites,
and the `MainMenuLink` you deleted used it too). `TopNavDropdown` is the outlier. Not a
3.2.4 failure — the two variants sit behind mutually exclusive breakpoints, so no single
user meets both — and 2.5.3 Label in Name is satisfied either way. Purely tidiness, and
you're already in both files.

---

## Code

### C1 · The dropdown only opens because `slug === object-key`, and nothing tests it

This is the finding I'd act on if you only take one.

`topNav.query.ts:141` renames `slug: "labs"` → `slug: "aiExperiments"`. That looks
cosmetic; it is now **the mechanism**, in three independent places:

- `SubNav.tsx:81` — `onClick(btn.slug as keyof T)`
- `TopNavDropdown.tsx:65,88` — `Object.entries(teachersData)`, gated on `selectedMenu === menu` where `menu` is the **object key**
- `SubNav.tsx:85` / `TopNavDropdown.tsx:347` — `aria-controls` and the panel `id`, both `topnav-teachers-${slug}`

On `main` the mismatch (`slug: "labs"` vs key `aiExperiments`) was harmless because the item
was a plain `<a>`. Now it isn't.

**Failure mode**: someone adds a fourth dropdown as `{ key: "support", slug: "help-and-support" }`.
Types pass. Nothing throws. The button renders, is focusable, has `aria-haspopup` — and
**clicking it opens nothing**, because no `Object.entries` key matches and every section stays
hidden. `aria-controls` also points at a non-existent id.

**Your new test cannot catch this**, because it passes `selectedMenu="aiExperiments"` as a
prop and bypasses the click wiring. I checked `TopNav.test.tsx` specifically for a
click-and-assert-visible loop over `subnavLabels`: there isn't one. Both uses are
keyboard-only — a tab-order assertion (line 244) and Arrow roving (line 323) — and the one
click-to-open test targets "Secondary" by name. So the open path for AI experiments is
genuinely uncovered.

Cheapest fix pins it for every present and future entry:

```ts
// topNav.query.test.ts
Object.entries(res.teachers!).forEach(([key, item]) => {
  expect(getItemSlug(item as Record<string, unknown>)).toBe(key);
});
```

`getItemSlug` is already exported from the schema and already handles the browse case. Plus
one assertion in `SubNav.test.tsx` that clicking calls `onClick` with `"aiExperiments"`.

### C2 · `analyticsPageName: "Landing Page"` collides with `/lp/:lpSlug`

`"Landing Page"` occurs exactly once on `main` (`urls.ts:856`), bound to
`"landing-page": { pathPattern: "/lp/:lpSlug" }`. So from merge day, every event carrying
`pageName: "Landing Page"` mixes `/mcp` traffic with CMS marketing-page traffic. Recoverable
by also filtering on path, but historical rows can't be relabelled.

This matters more than it looks: the MCP work has a usage OKR attached, and `/mcp` is its
front door — so "did the nav dropdown drive traffic" is the one number you'd want clean.

**I can see why you did it**: `createOakPageConfig` requires `analyticsPageName: PageNameValueType`
for `configType: "internal"`, and that union is generated from the Avo tracking plan, so you
couldn't invent a value in-repo. Options, best first:

1. Add an MCP page name to the Avo tracking plan (correct; out-of-repo).
2. `ExternalPageName` *is* locally editable — so if `/mcp` ends up served by a rewrite rather
   than an OWA route, `configType: "external"` + `"[external] Oak Curriculum MCP"` resolves it
   entirely in-repo. **Not applicable here** — I checked, #4454 adds a real
   `src/app/(core)/mcp/page.tsx` (`force-static`, no auth gate), so `internal` is the honest
   config and your ungated nav link is coherent with it.
3. Failing 1, pick an existing value not already bound to a different route family, with a
   comment saying why.

### C3 · `isNavDropDownButtonItem` has no call sites — the edit was speculative

`topNav.schema.ts:230-239` gained its third slug (`"aiExperiments"`). I swept all 83
nav-named files in `src/` for consumers, with a control probe (`isDropdownMenuItem` correctly
returned its definition plus its two real consumers, `SubNav.tsx` and `TopNavDropdown.tsx`).

`isNavDropDownButtonItem` appears in **exactly one file: its own definition.** Zero call
sites. The two components use the *structural* `isDropdownMenuItem` (`"children" in section`)
instead.

*Scope caveat: 83 nav-named files, not an exhaustive sweep of all 2,667 `src` blobs — GitHub's
code-search API is returning 404 for this repo, so one `rg "isNavDropDownButtonItem" src/` in
a local clone would confirm.*

So this is dead code, and your edit to it changes nothing. **Either delete it, or leave it and
don't feel obliged to maintain it.** The deeper issue: it answers a *structural* question by
string identity, and the `section is NavDropDownButton` predicate means the compiler can never
disagree. The root cause is that the `NavButton` union has no discriminant field — a `kind`
discriminant would let both guards go. Follow-up, not this PR.

### C4 · Nested ternary: the fallback silently lies

`HamburgerSubMenu.tsx:108-113`:

```ts
submenuOpen.value === "About us" ? navData.aboutUs
  : submenuOpen.value === "AI experiments" ? navData.aiExperiments
  : navData.guidance
```

The union is `"About us" | "Guidance" | "AI experiments"` — fully known at compile time — but
`guidance` is an implicit fallback. It just grew from 2 to 3 **in this PR**, which is the
evidence that it grows.

**Failure mode**: add a `"Support"` entry. This file compiles unchanged. The user taps
"Support", the heading reads **"Support"** (from `submenuOpen.value`) and the body renders the
**Guidance** links. No type error, no runtime error, no failing test — and a heading that
actively lies about its content.

Five lines fixes it now (`const _exhaustive: never = value`). But the better fix deletes the
mapping: the root cause is that `OakState.value` is a **display title** while `navData` is
keyed by **slug**, so a hand-maintained title→slug map has to exist at all. Key the state by
slug and it vanishes:

```ts
type OakState = { menu: "OakMenu"; value: "aboutUs" | "guidance" | "aiExperiments" };
const links = navData[submenuOpen.value];
// <SubmenuContainer title={links.title} …>
```

Zero branches, exhaustive by construction, and the heading comes from the same object as the
links so it can't contradict them. It would also incidentally fix a latent bug: `"About us" + "button"`
and `"AI experiments" + "button"` produce ids **containing a space**, which are invalid HTML
ids and unreachable via `querySelector("#…")` — `getElementById` tolerating them is the only
reason back-focus works today. If you do re-key, the return-focus ids in
`HamburgerMainMenu.tsx:36-38,202` must move to slugs in the same change. Follow-up sized.

### C5 · `href` and `external` are computed in the query, then thrown away

`TopNavDropdown.tsx:361-363` and `HamburgerSubMenu.tsx:140-142` both ignore `link.href` and
re-derive from the slug: `resolveOakHref({ page: link.slug } as …)`.

So `NavLink.href` is **dead data** for every dropdown child. Pre-existing (Guidance and About
us already do it), but this PR triples the children flowing through it and adds the `aila` and
`mcp` `OAK_PAGES` entries solely to satisfy the indirection.

- Your new `topNav.query.test.ts` asserts those exact `href` values — **a value that provably
  never reaches a user.** Edit `href` for a `utm_source` and the test stays green while the DOM
  doesn't change.
- Add a child whose slug isn't an `OAK_PAGES` key and the `as` cast silences the compiler; at
  runtime `resolveOakHref` catches, reports `OakError("urls/failed-to-resolve")`, and **returns
  `"/"`** — a nav item silently linking to the homepage.

Minimal fix, which makes the compiler enforce it and deletes both casts:

```ts
export type NavLink = {
  title: string;
  slug: OakLinkPropsRequiringPageOnly["page"];   // was: string
  href: string;
  external?: boolean;
};
```

Narrow `NavLink` **alone** — `SubjectsMenu.slug` / `Ks4OptionsMenu.slug` are subject slugs, not
page keys, and must keep `string`.

Also: `external` is a third duplicated fact — `configType: "external"` in `OAK_PAGES` already
knows. Data can contradict the registry.

### C6 · The green `lint, check types` is not why this works

Worth knowing, because it affects how much the passing CI tells you. `SubNav.tsx:42-47` does
`Object.values(data)` where `data` is the rest of a **generic** `SubNavProps<T>`. With no index
signature, TS falls through to the `values(o: {}): any[]` overload — verified with a reduced
repro against the workspace `tsc`: the generic case yields `any[]`, the concrete
`TeachersSubNavData` case yields `(NavDropDownButton | TeachersBrowse)[]`. A probe reading
`btn.thisFieldDoesNotExistAnywhere` compiles clean under `--strict`.

So inside `SubNav`, `btn.href` / `btn.title` / `btn.slug` / `btn.external` are **unchecked**. Had
the link/dropdown split been missed, `getLinkProps(btn.slug, btn.href, …)` would have compiled
and rendered `href={undefined}`.

To be clear: **every consumer was in fact updated correctly.** The reason nothing broke is the
*runtime* structural guard `isDropdownMenuItem` routing the now-childed `aiExperiments` to the
button branch — not the type system. `Object.values(data as SubNavData)` at line 42 restores
the union.

Minor related note: with `aiExperiments` now having children, **every** teachers top-level item
is a dropdown, so `SubNav`'s link branch (lines 154-160) is reachable only for pupils `help`.

### C7 · Pre-existing, separate ticket: the hamburger back button is named after the page

`HamburgerSubMenu.tsx:79-90` — the back control's accessible name is the submenu title
("AI experiments"), with `ariaLabel` left `undefined`; the only back affordance is the
`chevron-left` glyph. A VoiceOver user hears "AI experiments, button" for the control that
goes *back*, having just activated a button of that name to get there. SC 2.4.6 asks that
labels describe purpose. Also an `h3` nested inside a button, so the visual heading isn't a
real heading to AT. Unchanged by this PR — `SubmenuContainer` already serves About us,
Guidance, Phases and KeystageOptions the same way.

---

## What's good, explicitly

- **Dead-code removal is complete**: `MainMenuLink` gone along with `OakIconName`, `next/link`
  and `resolveOakHref` — the three imports it alone justified. No orphans.
- **`aria-controls` / panel-`id` pairing is consistent** on both sides, and correctly
  conditional on the target existing.
- **Fixture and query agree exactly** on slugs, titles, hrefs and flags — so the component
  tests aren't testing a fiction.
- **`renderNavButtons` deliberately left alone**, containing the width blast radius.
- **Return-focus wiring survived** a change that could easily have broken it.
- **The `TopNavDropdown` test is genuinely good** — heading, three labels, three hrefs, per-item
  `target`. That's describing required behaviour, not ratifying the implementation.
- **No security, auth, PII or injection surface.** No hand-written `any`, no new `!`, no new
  `process.env`. No lint cap raised, no ignore added, no dependency cycle.

---

## Suggested split

**In this PR**: rebase; read the Sonar issue; get Percy properly approved; mention the
Guidance/About us restyle in the description; swap the `aria-haspopup` assertion for
`aria-expanded` (A1); drop or fix the Escape step (A2); add the slug↔key invariant test (C1).

**Follow-up tickets**: `analyticsPageName` for `/mcp` via Avo (C2); remove `aria-haspopup`
from all five desktop triggers plus `MainMenuButton`, and name the hamburger back button
(A1 + C7 — one a11y ticket); re-key `OakState` by slug and move the return-focus ids (C4);
narrow `NavLink.slug` and delete the two casts (C5); add a `NavButton` discriminant and
delete both type guards, or delete the dead one (C3).

**And the actual gate**: get #4454 merged first, or `/mcp` 404s for every teacher who finds
the new dropdown.

---

<sub>Reviewed with two specialist agents (code quality; accessibility against WCAG 2.2 AA,
ARIA 1.2/1.3 and the APG) plus direct verification. Not verified: the SonarCloud issue
(couldn't read the dashboard), the Percy diffs (not rendered — no visual verdict issued),
`OakLeftAlignedButton`'s handling of a raw CSS keyword vs a spacing token, and TDD commit
sequencing (squashed diff). Production measurements taken 2026-09-01 ~10:42–11:05 UTC.</sub>

messmike (approved):

messmike:
Hi @mantagen apologies for approving prematurely! The nav item should now be 'AI plugin' with the url update to /ai-plugin. 
Screenshot 2026-09-23 at 8 42 39 am

It might be good to have a QA have a quick review also.

luke-arnold-oak:
One thing I found: the new mcp entry uses analyticsPageName: "Landing Page", so /ai-plugin page views will share a page name with the /lp/* CMS landing pages in PostHog. Is there a dedicated page name for it in Avo yet? The repo's last pull was 4 September and doesn't have one.

messmike:
thanks for the name update @mantagen - is it possible to have it say 'AI plugin' without the 'p' capitalised?

PR #4464, opened 2026-08-26
Feat/lesq 2195/keystage filter desktop [LESQ-2195]

njblore:
## Description

Music year: {owa_music_year}

Refactor filter browse store:
- create a hook dir and moves useBrowseFilters, and add `useKeyStagePresence` for to reduce duplication of shared filter calculations in filter components
- add a typed filter state and move utilities for parsing, applying, serialising, and resolving URL search parameters from `filtering` to the filter store utils folder.
- updates the store to maintain yearsForKeystage whenever filters change, removing this logic from consumer components
- refactors programme filter components -> move from `CurricFilters...` to `BrowseFilters..` living alongside the programme page
- remove duplicated legacy curriculum filter components and tests

Add KS filters
- create a `BrowseFilterKeystages` component
- add a store fn to set keystage filter state and use in component

## How to test

1. Go to [OWA Preview URL from Vercel bot comment]teachers/programmes/science-secondary-aqa/units
2. Click on `Key stage 3`
3. The page heading should update, the year filters should reduce down to ks3 applicable filters, and KS4 filters should be removed
4. Click on `Keystage 4`
5. The page heading should update, year filters should reduce to ks4 applicable filters and the ks4 filters should appear, ks3 filter should disappear
6. Other filter behaviour should work the same as it did previously

## Screenshots

Screenshot 2026-08-28 at 10 37 26


njblore:
@RBrownsett I didn't realise, but this ticket automatically does the [mobile filters](https://app.notion.com/p/oaknationalacademy/Add-key-stage-filter-to-programme-page-Mobile-3ab26cc4e1b18051950cc24bbd74af1f?source=copy_link) as well

src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ProgrammeFilters.tsx
```diff
@@ -74,42 +73,44 @@ export function ProgrammeFilters({
   ks4Options,
   ks4OptionFilterDimensions,
 }: Readonly<ProgrammeFiltersProps>) {
-  const { filters, onChangeFilters } = useBrowseFilters();
   return (
     <>
-      {getDisplayedFilters(data, filters, slugs, ks4Options).map(
-        ({ key, component: FilterComponent, shouldDisplayFilter }) => {
+      {useDisplayedFilters(data, slugs, ks4Options).map(
```
Comment (couplm) [inline src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ProgrammeFilters.tsx:78]: I'm not 100% certain on this, but I think that if you memoize the each of the filters, then only the ones that need to change will, which could help us a bit with our INP score :)

Comment (njblore) [reply src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ProgrammeFilters.tsx:78]: had a look and since `useKeystagePresence` already memoises available filters/year I don't think it's worth adding much more here since it's a pretty small list to begin with

RBrownsett:
@njblore tested both desktop and mobile, looks good. The only small question/issue is the keystage filter shows for Rule of law (the year filters are hidden) wonder if its best to keep the same behaviour and not show it. What do you think?  https://oak-web-application-website-git-feat-lesq-2195keystage-f-50eeb6.vercel.thenational.academy/teachers/programmes/rule-of-law-primary/units

njblore:
@RBrownsett yeah it's a bit odd, because there are multiple key stages in the data the filters will show, years is different because we override them all as `all-years` and we hide when there is only 1 filter option. We could add a subject specific override for Rule of Law for now if it's a blocker but longer term I'd rather use an override and exception to manage hiding filters when needed

RBrownsett:
> Preview



> @RBrownsett yeah it's a bit odd, because there are multiple key stages in the data the filters will show, years is different because we override them all as `all-years` and we hide when there is only 1 filter option. We could add a subject specific override for Rule of Law for now if it's a blocker but longer term I'd rather use an override and exception to manage hiding filters when needed

@njblore yeah makes sense, not a blocker works fine. I'll raise a ticket to use the over ride and exceptions to help manage the view of filters

src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/BrowseFilters/BrowseFiltersChildSubjects.tsx
```diff
@@ -0,0 +1,66 @@
...
+export type BrowseFiltersChildSubjectsProps = {
+  data: CurriculumUnitsFormattedData<Unit>;
+};
+
+export function BrowseFiltersChildSubjects({
+  data,
+}: Readonly<BrowseFiltersChildSubjectsProps>) {
+  const { filters, setChildSubjectFilter, yearsForKeystage } =
+    useBrowseFilters();
+  const id = useId();
```
Comment (Taha-Hassan-Git) [inline src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/BrowseFilters/BrowseFiltersChildSubjects.tsx:25]: I get that this is just a renamed component, but any idea what this line is for?

src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ProgrammeFilters.tsx
```diff
@@ -2,37 +2,44 @@ import {
...
   ks4Options: Ks4Option[],
 ) => {
+  const { filters } = useBrowseFilters();
+  const { childSubjectsAt, subjectCategoriesAt, tiersAt } =
+    useKeyStagePresence(data);
+
   return [
     {
+      key: "keystages",
+      shouldDisplayFilter: data.keystages.length > 1,
```
Comment (Taha-Hassan-Git) [inline src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ProgrammeFilters.tsx:29]: Not sure how, but we don't seem to have any tests for hiding the filters. I think this is a pretty likely area for regressions and bugs so definitely worth some coverage!

src/components/CurriculumComponents/CurricVisualiserFilters/index.tsx
```diff
@@ -1,5 +1,6 @@
-export { CurricFiltersChildSubjects } from "./CurricFiltersChildSubjects";
-export { CurricFiltersSubjectCategories } from "./CurricFiltersSubjectCategories";
 export { CurricFiltersThreads } from "./CurricFiltersThreads";
-export { CurricFiltersTiers } from "./CurricFiltersTiers";
 export { CurricFiltersYears } from "./CurricFiltersYears";
+export { BrowseFiltersYears } from "../../../app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/BrowseFilters/BrowseFiltersYears";
```
Taha-Hassan-Git: Alias imports!

src/utils/curriculum/filtering.ts
```diff
@@ -377,6 +238,48 @@ export function childSubjectForFilter(
     .find((childSubject) => childSubject.subject_slug === slug);
 }
 
+export function shouldDisplayFilter(
```
Taha-Hassan-Git, line 242: Hmm, looks like we now have this logic in two places which isn't ideal. Is there a reason for that, and any way to avoid it?

Taha-Hassan-Git (commented):
Quite a big PR to warm up my code-reading skills!

I've noticed there's a bug with units which have been filtered out still showing as a match for thread highlighting, despite not being on the page. This feels familiar. Not sure if it's a bug we've previously fixed and has regressed again, or if it's something we've been meaning to fix and haven't got around to. @RBrownsett ?

Screenshot 2026-09-01 at 10 35 52

A few other things below.

RBrownsett:
> Quite a big PR to warm up my code-reading skills!
> 
> I've noticed there's a bug with units which have been filtered out still showing as a match for thread highlighting, despite not being on the page. This feels familiar. Not sure if it's a bug we've previously fixed and has regressed again, or if it's something we've been meaning to fix and haven't got around to. @RBrownsett ?
> 
> Screenshot 2026-09-01 at 10 35 52
> A few other things below.

@Taha-Hassan-Git yeah I thought it was fixed. I tried to recreate the filters you had and I got one highlighted unit which looked correct? I had computing primary, KS1, Year 2 with the thread Data and information 

image

src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/BrowseFilters/BrowseFiltersChildSubjects.tsx
Comment (njblore) [reply src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/BrowseFilters/BrowseFiltersChildSubjects.tsx:25]: Hmm yeah not too sure here, it is usually used to create a unique id for proper aria labelling but not sure if it's being used that way internally on the radio group? it looks like it's just being used to generate a unique id to name the group, potentially to avoid clashes on page with other radio groups?

src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ProgrammeFilters.tsx
Comment (njblore) [reply src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/ProgrammeFilters.tsx:29]: oh good catch! will add some in

src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/BrowseFilters/BrowseFiltersChildSubjects.tsx
Comment (Taha-Hassan-Git) [reply src/app/(core)/teachers/programmes/[slug]/[tab]/Components/Filters/BrowseFilters/BrowseFiltersChildSubjects.tsx:25]: Seems sensible. Maybe a pattern I'll pinch for the future.

Taha-Hassan-Git:
@RBrownsett I'm dumb, it was just obscured by that filter drawer. Please have me fired!

src/utils/curriculum/filtering.ts
njblore: refactored to use shouldDisplayFilters in the programme page directly so the core logic is shared

RBrownsett (approved):
Nice work, testing notes on the ticket

Taha-Hassan-Git (approved):
Nice!

PR #4469, opened 2026-08-28
Fix/lesq 2243/slow examboard toggle [LESQ-2243]

njblore:
## Description

Music year: 1959

Updated a few areas that seem to have been contributing to slowness on the programme page, particularly when toggling between ks4 options (which require a full page load as they're baked into the programme slug):
- add prefetching to the ks4 options buttons, because these are radio buttons and not `Link`s there was no prefetching happening on these previously
- update next config to add a `staleTimes` option which increases the amount of time prefetched links are valid on dynamic routes (default is `0`)
- add caching to `getFileSizes` on the programme page, this request was running fresh on every render and seemed to be doing quite a lot of work
- add caching to `getImplementationGuides` sanity request
- remove waterfalling requests that have no dependency on the programme page, use Promise.all to run concurrently
- move `draftMode` out of the core layout, this was opting the entire sub route into dynamic rendering! it's still needed on the programme page for now
- add memoization to the filtering logic for the units on the programme page

*Tagging adopt squaddies to check over the implementation guides feature as I've added caching to the `getFileSizes` and `implementationGuides` sanity request, as well as restructure the fetches to remove waterfall delays

***Note***
Using APIs like `cookies`, `searchParams` and `draftMode` in the page opts the whole route into `dynamic` rendering, ideally we'd avoid using `cookies` here and find a solution for `draftMode` and `searchParams`. Partial prerendering could be a way forward but it's experimental in Next15, so we'd need to update to v16. There will be follow on tickets to explore this further.

## How to test

1. Go to https://oak-web-application-website-git-fix-lesq-2243slow-exambo-c2ef0a.vercel.thenational.academy/teachers/programmes/science-secondary-edexcel/units?tiers=foundation
2. Click on KS4 option toggles
3. It should feel a bit faster than production initially, and subsequent toggles should feel much faster, ie. going back and forth between the options
4. The page should work as before otherwise


njblore:
@copilot can you generate a test plan for a QA to use to review this PR

RBrownsett (approved):
Approved, notable improvement in performance when switching exam boards. Tested KS4 and subject pages. Regression tested subjects without exam boards and primary with no issues.

orangemug (approved):
LGTM

Taha-Hassan-Git (approved):
FAB!

PR #4473, opened 2026-09-01
Feat/lesq 2245/heading component

couplm:
## Description

Music year: 1953

- Added `TeachWithOakHeader`
- Amended AboutSharedHeader to allow options for keeping the image on collapse
- Fix overflow text overlap issue

## How to test

1. Go to Storybook
2. Click on TeachWithOakHeader

## Screenshots

How it should now look:
image

This is a placeholder image, proper dimensions and configuration can be set with a new image.

## Checklist

- [ ] Design sign-off


src/components/GenericPagesComponents/OaksImpactHeader/index.tsx
couplm, line 1: I didn't know what to rename it to 😳 so I left it for now but something like `MediaHeader` or `PageHeaderWithMedia` was suggested by copilot.

src/components/GenericPagesComponents/OaksImpactHeader/index.tsx
```diff
@@ -67,65 +86,89 @@ const LeftTextLayout = styled(OakFlex)`
...
+                  transcript={video.transcript ?? undefined}
+                  showTranscript={true}
+                />
+              )}
+              {image && mediaDescription && (
+                <OakImage
+                  src={image}
+                  alt={mediaDescription}
+                  $aspectRatio={"16/9"}
+                  $objectFit={"cover"}
```
couplm: couldn't see a good convention for image size when dynamically changing so put it as this for now

messmike:
Thanks @couplm - looking good! The only thing I can spot is the overlap between the back button and the header when resizing the browser window: 
Screenshot 2026-09-01 at 15 56 57

Is it possible to have the back button appear above the heading, and not overlap?

src/components/GenericPagesComponents/OaksImpactHeader/index.tsx
```diff
@@ -67,65 +86,99 @@ const LeftTextLayout = styled(OakFlex)`
 export type OaksImpactHeaderProps = {
```
couplm, line 67: I also considered making optional types, something like `OakImpactHeaderWithImage | OakImpactHeaderWithVideo` so you could only pick one, instead of as below where you could technically have both. I thought more flexibility was better but open to opinions on this

src/components/GenericPagesComponents/OaksImpactHeader/index.tsx
njblore: I wonder if this should become an OakComponent now that it will be shared? One to check in with Mike about I reckon

couplm: @messmike, what do you think? Should this be an Oak Component now?

src/components/GenericPagesComponents/OaksImpactHeader/index.tsx
njblore: Yeah I think something like that would be a bit easier to reason about personally! `props = {...sharedProps} & (ImageProps | VideoProps)`

src/components/GenericPagesComponents/OaksImpactHeader/index.tsx
messmike: Yeah, I think it should be an Oak Component, shared. I think it's worth raising in the oak-components Slack channel for visibility as well.

src/components/GenericPagesComponents/OaksImpactHeader/index.tsx
```diff
@@ -67,65 +86,99 @@ const LeftTextLayout = styled(OakFlex)`
...
-                  title={video.title}
-                  isLegacy={false}
-                  location="marketing"
-                  omitBorder={true}
+              {body}
+            </OakP>
+          </LeftTextLayout>
+          <OakFlex $flexDirection={"column"} $gap={"spacing-16"}>
+            <MediaLayout>
+              {video && mediaDescription && (
```
njblore: is there a variant where we'd have a video and an image? if not I'd rather be explicit that it's one or the other in the render method by using a ternary, eg. `hasVideo ? <Video /> : hasImage ? <Image /> : null`. (if it can have both, ignore me!)

src/components/TeacherComponents/TeachWithOakHeader/TeachWithOakHeader.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <HeaderLayout>
+        <AboutSharedHeader
+          title={"The thinking behind Oak lessons"}
+          content={
+            "See how our lessons are designed to support learning - and make the most of them in your classroom."
+          }
+          showImageOverflow={true}
+        >
+          <StyledAboutSharedHeaderImage
+            imageUrl={
```
couplm: This is a placeholder image and the width set in the styling is to make it wider for Storybook

njblore: can we use a Cloudinary image and get the url with `getCloudinaryImageUrl` instead of hardcoding a full path here ([example](https://console.cloudinary.com/app/c-56adc306ce4b3e1a8e77e2e0b7eb5a/assets/media_library/folders/c9ddcb3b9389332ddc986e43ca4ea1f13f/asset/a70465dad215780565c5c76097761fdf/manage?view_mode=mosaic&context=manage) you can get the image id by copying the image URL and taking everything after `v1234/OWA/...`)

couplm: Thank you :) I don't have access to check but just used a different placeholder

couplm:
@messmike, writing here for visibility: I've updated the work to be a new component that uses the shared `AboutSharedHeader`. This is instead of it being a new component in Oak Components as the other one we previously identified was an edge case component and this one is preferred. We hadn't seen it originally but it suits our use case better.

The overflow issue should also be resolved now with this new component 🤞

messmike:
Thanks @couplm - agree that all good now with the back button! There seems to be some layout issues on desktop and tablet breakpoints (also noticed mobile doesn't seem to stack) - sorry if it's not ready to review that aspect - feel free to ignore if so!:

Designs: 
Screenshot 2026-09-03 at 09 26 45
Screenshot 2026-09-03 at 09 27 08

Storybook preview:
Screenshot 2026-09-03 at 09 30 16
Screenshot 2026-09-03 at 09 30 27
Screenshot 2026-09-03 at 09 30 31

src/components/TeacherComponents/TeachWithOakHeader/TeachWithOakHeader.tsx
```diff
@@ -0,0 +1,65 @@
...
+import styled from "styled-components";
+
+import {
+  AboutSharedHeader,
+  AboutSharedHeaderImage,
+} from "@/components/GenericPagesComponents/AboutSharedHeader";
+import { NewGutterMaxWidth } from "@/components/GenericPagesComponents/NewGutterMaxWidth";
+import { getCloudinaryImageUrl } from "@/utils/getCloudinaryImageUrl";
+
+const HeaderLayout = styled.div`
```
njblore: any reason not to use an `OakFlex` for this?

src/components/GenericPagesComponents/AboutSharedHeader/index.tsx
```diff
@@ -65,20 +66,24 @@ export function BackgroundHeaderLoop() {
   );
 }
 
+type AboutSharedHeaderImageProps = {
+  imageUrl: string;
+  imageAlt?: string;
+} & Omit<React.ComponentProps<typeof OakImage>, "alt" | "src">;
```
njblore: do we have an `OakImageProps` export we can use here?

src/components/TeacherComponents/TeachWithOakHeader/TeachWithOakHeader.tsx
```diff
@@ -0,0 +1,65 @@
...
+  display: flex;
+  flex-direction: row;
+`;
+
+const StyledAboutSharedHeaderImage = styled(AboutSharedHeaderImage)`
+  width: 632px;
```
njblore: could we get alignment on the image styling across pages rather than having to override here? we generally don't want to be sizing things by pixel sizing where at all avoidable!

couplm: Yes I agree, I had a lot of problems before with trying to specify it to auto adjust or be 100% as there wasn't a sized container but I'll try again with the aspect ratio set, maybe that's fixed it

src/components/TeacherComponents/TeachWithOakHeader/TeachWithOakHeader.tsx
couplm: No I just forgot 😭

couplm:
@messmike thanks Mike, yep it does look weird you're right. I amended the spacing to be 16px consistently across the breakpoints which I think has helped a bit. It looks OK on the default view but the autodocs view looks weird. I think it's because the component is meant to allow the image to disappear instead of stack. Should the image stack earlier, such as on tablet viewports, or maybe reduce in width?

messmike:
> @messmike thanks Mike, yep it does look weird you're right. I amended the spacing to be 16px consistently across the breakpoints which I think has helped a bit. It looks OK on the default view but the autodocs view looks weird. I think it's because the component is meant to allow the image to disappear instead of stack. Should the image stack earlier, such as on tablet viewports, or maybe reduce in width?

@couplm I think closer to how the containers behave across all breakpoints with the `OaksImpactHeader` ideally (not sure if it is possible to copy from that?


https://github.com/user-attachments/assets/4e2106ff-74be-41a8-aa44-6d8c408706b8

RBrownsett:
@couplm just looking at this, is the “Teach with Oak Image” just placeholder alt text in the component? Just thinking that if we stick with this image then I'd expect it to render with alt="" in the final implementation?

messmike:
> @couplm just looking at this, is the “Teach with Oak Image” just placeholder alt text in the component? Just thinking that if we stick with this image then I'd expect it to render with alt="" in the final implementation?

yeah this is a placeholder for now @RBrownsett I was discussing replacing this with an illustration with Kat, but will need to check with her when she is back what we want to do with that.

couplm:
@messmike, please can you take another look? I think I have resolved everything. I was having a lot of problems not wanting to amend the shared component too much to our use case given it's used so much on the about-us pages. What I did feels a little hacky code-wise but seems to be an OK compromise and hopefully looks OK?

messmike:
> @messmike, please can you take another look? I think I have resolved everything. I was having a lot of problems not wanting to amend the shared component too much to our use case given it's used so much on the about-us pages. What I did feels a little hacky code-wise but seems to be an OK compromise and hopefully looks OK?

@couplm I think it's gone the other way with the header and subheader now taking up more of the space at the desktop and tablet breakpoints when compared to the designs and the `OaksImpactHeader` component: 

https://github.com/user-attachments/assets/ae373d62-68fe-4da9-b425-c4130d95207f


Mobile is looking good though!

Also, a small thing, but the colour token for the background of the header should use `bg-decorative2-main`.

couplm:
@messmike I'm so sorry I was so focused on getting the image to move I didn't even notice that. I've given it another update 🤞 
image

messmike:
> Preview

No apologies needed! looking good, thanks @couplm

messmike (approved):

src/components/GenericPagesComponents/AboutSharedHeader/index.tsx
```diff
@@ -15,13 +15,30 @@ import { PortableTextReactComponents } from "@portabletext/react";
...
+    display: ${({ $showImageOverflow }) =>
+      $showImageOverflow ? "block" : "none"};
+  }
+`;
+
+const TextLayout = styled(OakFlex)<{ $showImageOverflow: boolean }>`
+  flex: ${({ $showImageOverflow }) =>
+    $showImageOverflow ? "1 1 0" : "0 1 auto"};
+  min-width: 0;
+`;
```
njblore: could this not be done inline as `<OakFlex $minWidth="spacing-0" $flexBasis={showImageOverflow ? 0 : 'auto' } ... >`? same with `HeaderLayout` that could just be `<OakFlex $flexDirection={[showImageOverflow? "column" : "row", "row"]} >` inline. I think that would be more consistent with how we use components in OWA, we usually only use `styled` as a last resort!

njblore (approved):
🖌️

njblore:
Just noticed there's a bit of overlap on the heading background! (the bottom of the `g` on the top row)
Screenshot 2026-09-04 at 10 22 45

src/components/GenericPagesComponents/AboutSharedHeader/index.tsx
couplm: I've removed styled for these, thanks for the advice, and I fixed the overlap issue below by increasing the line height, which hopefully is fine :)

src/components/GenericPagesComponents/AboutSharedHeader/index.tsx
```diff
@@ -100,28 +108,47 @@ export type AboutSharedHeaderProps = {
...
+          $flexDirection={"column"}
+          $gap={"spacing-24"}
+        >
           <OakHeading tag="h1" $font={["heading-4", "heading-2", "heading-2"]}>
             <OakSpan
               $background={titleHighlight ?? "bg-decorative1-main"}
               $ph={"spacing-4"}
+              style={{
+                lineHeight: 1.25,
+                boxDecorationBreak: "clone",
```
orangemug, line 151: Note for self https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/box-decoration-break looks like this isn't widely support check how it looks in safari

orangemug: @couplm you need to add the following for safari support

```
"-webkit-box-decoration-break": "clone",
```

Without `-webkit-box-decoration-break`

Screenshot 2026-09-08 at 10 58 32

With `-webkit-box-decoration-break`
 
Screenshot 2026-09-08 at 10 58 24

couplm: Thanks Jamie! I've updated :)

orangemug (approved):

PR #4474, opened 2026-09-02
Remove download link after clicking

Taha-Hassan-Git:
## Description

Music year: 1960

This PR removes download links after they've been clicked in the lesson and programme downloads flow. Previously elements were being attached to the body element outside of Reacts ecosystem and left there, in Next App Router, React no longer manipulates elements inside the #next div, but the body tag directly. This led to issues with React and the Dom getting out of sync.

- Removes download links after clicked

## Issue(s)

Fixes #LESQ-2182

## How to test

Test for regressions in lesson and programme downloads

1. Go to [a lesson download page](https://oak-web-application-website-git-fix-lesq-2182try-catch-i-c93f46.vercel.thenational.academy/teachers/programmes/computing-primary-ks1/units/digital-painting/lessons/painting-using-computers/downloads?preselected=all)
2. Download the lesson

Test on safari also, as I had to touch code that fixed an issue with the download button on safari.

## Checklist

- [ ] Design sign-off


src/components/SharedComponents/helpers/downloadAndShareHelpers/createAndClickHiddenDownloadLink.tsx
```diff
@@ -1,15 +1,18 @@
...
-
+const reportError = errorReporter("createAndClickHiddenDownloadLink");
 export const createLink = () => {
   const previousLink = getDownloadLink();
-  if (previousLink) {
-    previousLink.remove();
+  try {
+    previousLink?.remove();
+  } catch (error) {
+    reportError(error);
```
njblore: thinking about it I'm not sure if we actually need to catch/report this, there's nothing for us to act on and no user impact, wdyt?

njblore (approved):
looks good! just a q on the reporting

src/components/SharedComponents/helpers/downloadAndShareHelpers/createAndClickHiddenDownloadLink.tsx
Taha-Hassan-Git: This is strange actually because this method [shouldn't be throwing anyway](https://developer.mozilla.org/en-US/docs/Web/API/Element/remove), and we've already checked if previousLink is null.

Looking at [the bugsnag error reports](https://app.bugsnag.com/oak-national-academy/oak-web-application/errors/6a43ee11d7a912dd2473b80c?filters[event.since]=30d&filters[error.status]=open&filters[search]=app%3Acore-layout&filters[user.id]=01a00cb9-c916-7b5f-8c3e-308e57213b81&pivot_tab=event) I no longer think this line is the source of them. They're reporting on a error after calling `.removeChild()` (which does throw). I can't find anywhere in our codebase where we call that directly, so my guess is that it's an internal React or Next thing. Very hard to identify where it's happening with the minimised class names in prod.

Taha-Hassan-Git: Oooohhh, I asked Claude and it thinks that now we're using App router appending dom nodes to the `<body>` outside of the React ecosystem is messing things up. It is these lines, but not for the reasons we thought. So if we append, click and then remove the link right after I think it should be fine. I think that's the actual fix.

njblore: oooh! ok interesting,  frustrating we can't figure out where it's coming from but that sounds like a reasonable fix!

src/components/SharedComponents/helpers/downloadAndShareHelpers/createAndClickHiddenDownloadLink.tsx
```diff
@@ -22,10 +17,13 @@ export const hideAndClickDownloadLink = (url: string, a: HTMLAnchorElement) => {
...
   a.addEventListener("click", () => {
     // Allows verification that the link has been clicked, used in the teacher lesson download journey
-    a.setAttribute("clicked", "true");
+    downloadLinkClicked = true;
   });
   document.body.appendChild(a);
   a.click();
+  // We need to remove the link immediately after clicking
+  // <body> is React-owned under the app router, so leaving a foreign node in it  can cause issues
+  a.remove();
```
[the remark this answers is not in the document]
Taha-Hassan-Git: .remove() does not throw when the element isn't there, it just no-ops.

RBrownsett (approved):
Approved, tested lesson and unit downloads in Chrome and Safari

PR #4491, opened 2026-09-07
refactor(mcp): serve the carousel images at /ai-plugin/carousel

mantagen:
## Description

Follows the `/mcp` → `/ai-plugin` landing-page rename (#4454). The MCP submission carousel images move on disk to `public/ai-plugin/carousel/`, and a **rewrite** keeps the published `/mcp/carousel/…` URLs serving the same bytes, at the same URL, with no redirect.

- `public/mcp/carousel/` → `public/ai-plugin/carousel/` (a `git mv`; bytes unchanged, digests still match the pinned values)
- `next.config.ts`: `/mcp/carousel/:file` → `/ai-plugin/carousel/:file`, in `rewrites()`, with an explicit removal condition
- The Playwright spec now covers **both** paths; the unit test follows the bytes to their new directory

**A rewrite, never a redirect.** Anthropic's MCP directory listing stores those URLs and refetches the images itself, indefinitely — the submission portal takes no uploads, so the URL *is* the delivery mechanism. The spec asserts no redirect on Oak's behalf, so a redirect is the exact failure this guards against. Nothing in this repo links these files, so a break would be invisible to every other check.

The Anthropic-held path literals in the spec are **unchanged, byte for byte** — deliberately not re-pointed. A second self-contained table covers the new canonical paths, digests repeated rather than shared, so "both URLs serve the same bytes" is measured rather than true by construction.

## Issue(s)

Fixes MCP-688

References MCP-686 — the **parent**, which this PR contributes to but does not complete. Its other two children are out of this PR's scope by boundary: MCP-689 is the Anthropic portal edit (out-of-repo, so no PR can close it) and MCP-690 removes this rewrite once MCP-689 is done. The removal condition is written into `next.config.ts`.

## How to test

Against the preview URL from the Vercel bot comment below:

1. `curl -sI <preview>/mcp/carousel/carousel_image_1.png` → `200`, `content-type: image/png`, and **no `Location` header**
2. `curl -sI <preview>/ai-plugin/carousel/carousel_image_1.png` → the same
3. Both bodies should hash to `06cbdbf1704e6960afb3ad6b43ddaa42c2c256689e60fe5061cca8c108ffa8a5`
4. Or run the spec directly: `BASE_URL=<preview> pnpm exec playwright test src/tests/e2e/mcp/carousel-assets.spec.ts` → 6 passed

Verified locally against a served production-mode build before raising this: `routes-manifest.json` carries the `afterFiles` rewrite, and all six URLs returned `200` / `image/png` / zero redirects with digests matching the three pinned values. Controls, so none of that is a false green: unknown paths under `/mcp/carousel/` return `text/html`, and with the server stopped the spec fails 6/6.

## Screenshots

n/a — no rendered output changes.

## Checklist

- [ ] Design sign-off

---

Agent-authored on behalf of `@mantagen` by Beacon tracks Embers (claude-code, claude-opus-5) — raised under Matthew's credential because no bot identity has write access to this repository. Questions route to Matthew.


luke-arnold-oak (approved):
Looks good to me. Playwright report confirms all six paths serve 200 / image/png with matching digests on the preview. Happy to approve.

PR #4511, opened 2026-09-10
fix: update withErrorHandling

njblore:
## Description

Music year: 2005

- next usually catches use of request time APIs and opts the route into dynamic rendering, but the `withPageErrorHandling` wrapper we added swallows the error that Next uses internally to turn the route dynamic
- add `unstable_rethrow` to the handler which will catch and rethrow any nextjs internal errors so they don't get swallowed
- add tests for `withPageErrorHandling` and update page tests to include `unstable_rethrow` in next/navigation mocks

## How to test

1. Go to https://oak-web-application-website-git-fix-static-rendering-of-5174c9.vercel.thenational.academy/teachers/teach-with-oak
2. You should be directed to a 404 page unless you are opted into the feature flag `teachers-teach-with-oak`


orangemug (approved):
Oh that broke my brain for a second, but after some reading makes sense to me.

Interesting how we have to use `unstable_*` for things like this

Taha-Hassan-Git (approved):
Great detective work! There's always more to learn when it comes to Next. Hope it stops soon. 😬

PR #4520, opened 2026-09-14
Chore/dependency updates/jest jsdom

couplm:
## Description

Music year: 1991

- Updated jsdom to v30
- Updated jest and dependencies to v30

## Checklist

- [ ] Design sign-off


src/__tests__/pages/teachers/lessons/[lessonSlug]/media.test.tsx
```diff
@@ -20,6 +20,30 @@ jest.mock("@/utils/handleTranscript.ts", () => ({
   populateMediaClipsWithTranscripts: jest.fn(),
 }));
 
+jest.mock(
```
Comment (Taha-Hassan-Git) [inline src/__tests__/pages/teachers/lessons/[lessonSlug]/media.test.tsx]: I was curious about these mocks and did some investigating. The need for this seems to be coming from a global jest mock in tests. Previously these hooks would just stay in the loading state rather than immediately return an empty string and error out. 

Rather than globally mocking fetch, I think it would be a better idea to keep it opt in. Then we wouldn't need these extra mocks in this and other files.

Claude has also identified a couple of snapshots that no longer embed the images they're supposed to which seems to also be a consequence of this global mock.

src/pages-helpers/curriculum/docx/__snapshots__/docx.test.ts.snap
```diff
@@ -206,14 +206,14 @@ exports[`docx insertImages test 1`] = `
       "word/media/hash_f1c6d68f4906606ef3ae58fac887d210ae8b33ce7275c21ee8e177090278e249png",
     ],
     "type": "CREATE",
-    "value": "4a5204e50a2367de974264a2056566e7af61af9d192b8699883d11491a37b599",
+    "value": "",
```
Taha-Hassan-Git: Here it is! This used to assert an image was embedded and now asserts that it is not embedded instead.

src/components/TeacherViews/LessonDownloads.view.test.tsx
```diff
@@ -13,7 +14,16 @@ import lessonDownloadsFixture from "@/node-lib/curriculum-api-2023/fixtures/less
 import renderWithProviders from "@/__tests__/__helpers__/renderWithProviders";
 import { resolveOakHref } from "@/common-lib/urls";
 
-const render = renderWithProviders();
+const renderWithProvidersFn = renderWithProviders();
+
+// The HubSpot sync effect resolves asynchronously, so flush it inside act()
+const render = async (ui: React.ReactElement) => {
```
Taha-Hassan-Git, line 21: Might not need this either after opt in fetch mocks

Taha-Hassan-Git (commented):
Great work on this, I just have a suggestion on fetch mocking, let me know what you think.

Taha-Hassan-Git:
I got Claude to help me look into this. I can push a commit with opt-in fetch mocking if you agree it's the right approach, but would ofc need your review on that code before we could commit.

couplm:
@Taha-Hassan-Git thank you! We were just trying to fix the issues and get this done so if you and Claude have a better way of doing that, that is great :)

Taha-Hassan-Git:
Take a look at my commits, if you're happy then feel free to merge with the dependency updates branch. I'm happy with everything at this point. 👍

PR #4530, opened 2026-09-18
Feat/lesq 2248/two analytics

njblore:
## Description

Music year: 1996

- List of changes

## How to test

2. Click on \_\_\_\_\_\_


src/app/(core)/teachers/teach-with-oak/components/TeachWithOakHeader/extractLessonAccessedPropsFromHref.ts
```diff
@@ -0,0 +1,84 @@
...
+  try {
+    pathname = new URL(returnTo, baseUrl).pathname;
+  } catch {
+    pathname = returnTo;
+  }
+
+  const segments = pathname.split("/").filter(Boolean);
+
+  if (
+    segments[0] !== "teachers" ||
```
Taha-Hassan-Git, line 37: Wondering about how to maintain this if/when url structure changes. At the moment I don't think any tests would blow up, this would just silently fail. Could we use reolveHref here to return a real url for a lesson?

Taha-Hassan-Git (commented):
Just one concern. Let me know what you think!

RBrownsett:
@njblore teach with oak event fires from both the teacher tip and promo area - just checking lessonAccessed is not firing when going back from the teach with oak page, should it be?

